PR #47, opened 2021-04-27
Added support for wrong control hashes.

DamianStraszak:


src/terminal.rs
```diff
@@ -222,13 +222,47 @@ impl<H: HashT, NI: NodeIdT> Terminal<H, NI> {
...
+            .unit_store
+            .get(u_hash)
+            .expect("Unit to be added to dag must be in store");
         self.post_insert.iter().for_each(|f| f(u.clone()));
         if let Some(children) = self.children_hash.remove(u_hash) {
             for v_hash in children {
                 self.new_parent_in_dag(&v_hash);
             }
         }
+        let u = self
```
mike1729, line 235: We've just fetched `u` from `unit_store` above, so we may clone it if we need another copy.

src/terminal.rs
```diff
@@ -243,11 +277,19 @@ impl<H: HashT, NI: NodeIdT> Terminal<H, NI> {
...
         for p_hash in u_parents.into_iter().flatten() {
-            let p = self.unit_store.get(&p_hash).unwrap();
-            if p.status == UnitStatus::InDag {
-                n_parents_in_dag += NodeCount(1);
-            } else {
-                self.add_hash_trigger(&p_hash, u_hash);
+            let maybe_p = self.unit_store.get(&p_hash);
+            // p might not be even in store because u might be a unit with wrong control hash
+            match maybe_p {
+                Some(p) => {
```
mike1729, line 283: You may simplify this:
`Some(p) if p.status == UnitStatus::InDag =>n_parents_in_dag += NodeCount(1),`
`_ => self.add_hash_trigger(&p_hash, u_hash),`

src/terminal.rs
DamianStraszak: fixed, as discussed

src/terminal.rs
DamianStraszak: Right, rewritten!

PR #51, opened 2021-05-06
Refactor hashes to avoid passing a function around

timorl:
Not yet tested with aleph-node.


timorl:
Needs complete rewrite after the environment refactor.

PR #54, opened 2021-05-11
Fixed tests in consensus after the Member refactor.

DamianStraszak:


src/testing.rs
```diff
@@ -42,186 +46,151 @@ pub mod mock {
...
     )]
-    pub struct Hash(pub u32);
+    pub struct Hash64(pub u64);
 
-    impl From<u32> for Hash {
-        fn from(h: u32) -> Self {
-            Hash(h)
+    impl From<u64> for Hash64 {
+        fn from(h: u64) -> Self {
+            Hash64(h as u64)
```
timorl, line 62: The `as u64` here seems redundant.

timorl (commented):
Do I understand correctly that this PR fixes tests for the old files, but `member` is completely untested? Will creating tests for `member` be a separate task? Will the tests in `consensus` be changed to go through member then?

timorl:
My questions have been answered during a meeting: yes, yes, no. Modulo that one other comment the change looks good to me.

src/testing.rs
DamianStraszak: Yes, removed.

PR #55, opened 2021-05-13
Refactor Signatures and Keyboxes

michalseweryn:
Please wait for the PR to aleph-node

### DIFF
--- a/src/bft.rs
+++ b/src/bft.rs
@@ -1,10 +1,10 @@
-use crate::{units::SignedUnit, Data, Hasher, NodeIndex};
+use crate::{units::UncheckedSignedUnit, Data, Hasher, NodeIndex};
 use codec::{Decode, Encode};
 
 #[derive(Debug, Encode, Decode)]
 pub(crate) struct ForkProof<H: Hasher, D: Data, Signature: Clone + Encode + Decode> {
-    pub(crate) u1: SignedUnit<H, D, Signature>,
-    pub(crate) u2: SignedUnit<H, D, Signature>,
+    pub(crate) u1: UncheckedSignedUnit<H, D, Signature>,
+    pub(crate) u2: UncheckedSignedUnit<H, D, Signature>,
 }
 
 #[derive(Debug, Encode, Decode)]
@@ -12,5 +12,5 @@ pub(crate) struct Alert<H: Hasher, D: Data, Signature: Clone + Encode + Decode> 
     pub(crate) sender: NodeIndex,
     pub(crate) forker: NodeIndex,
     pub(crate) proof: ForkProof<H, D, Signature>,
-    pub(crate) legit_units: Vec<SignedUnit<H, D, Signature>>,
+    pub(crate) legit_units: Vec<UncheckedSignedUnit<H, D, Signature>>,
 }
--- a/src/bft.rs
+++ b/src/bft.rs
@@ -1,16 +1,16 @@
 use crate::{units::UncheckedSignedUnit, Data, Hasher, NodeIndex};
 use codec::{Decode, Encode};
 
-#[derive(Debug, Encode, Decode)]
-pub(crate) struct ForkProof<H: Hasher, D: Data, Signature: Clone + Encode + Decode> {
-    pub(crate) u1: UncheckedSignedUnit<H, D, Signature>,
-    pub(crate) u2: UncheckedSignedUnit<H, D, Signature>,
+#[derive(Clone, Debug, Decode, Encode)]
+pub(crate) struct ForkProof<H: Hasher, D: Data, S> {
+    pub(crate) u1: UncheckedSignedUnit<H, D, S>,
+    pub(crate) u2: UncheckedSignedUnit<H, D, S>,
 }
 
-#[derive(Debug, Encode, Decode)]
-pub(crate) struct Alert<H: Hasher, D: Data, Signature: Clone + Encode + Decode> {
+#[derive(Clone, Debug, Decode, Encode)]
+pub(crate) struct Alert<H: Hasher, D: Data, S> {
     pub(crate) sender: NodeIndex,
     pub(crate) forker: NodeIndex,
-    pub(crate) proof: ForkProof<H, D, Signature>,
-    pub(crate) legit_units: Vec<UncheckedSignedUnit<H, D, Signature>>,
+    pub(crate) proof: ForkProof<H, D, S>,
+    pub(crate) legit_units: Vec<UncheckedSignedUnit<H, D, S>>,
 }
--- a/src/member.rs
+++ b/src/member.rs
@@ -253,9 +253,7 @@ where
             .unit_by_hash(&hash)
             .cloned()
             .expect("Our units are in store.");
-        let message = ConsensusMessage::<H, D, KB::Signature>::NewUnit(UncheckedSignedUnit::from(
-            signed_unit,
-        ));
+        let message = ConsensusMessage::<H, D, KB::Signature>::NewUnit(signed_unit.into());
         let command = NetworkCommand::SendToAll(message.encode());
         debug!(target: "rush-member", "Sending a unit {:?} over network after delay {:?}.", hash, interval);
         self.send_network_command(command);
@@ -315,7 +313,7 @@ where
     fn validate_unit_parents(&self, su: &SignedUnit<'a, H, D, KB>) -> bool {
         // NOTE: at this point we cannot validate correctness of the control hash, in principle it could be
         // just a random hash, but we still would not be able to deduce that by looking at the unit only.
-        let pre_unit = &su.signed().inner;
+        let pre_unit = &su.as_signable().inner;
         if pre_unit.n_members() != self.config.n_members {
             debug!(target: "rush-member", "Unit with wrong length of parents map.");
             return false;
@@ -343,26 +341,22 @@ where
         // TODO: make sure we check all that is necessary for unit correctness
         // TODO: consider moving validation logic for units and alerts to another file, note however
         // that access to the authority list is required for validation.
-        if su.signed().session_id != self.config.session_id {
+        if su.as_signable().session_id != self.config.session_id {
             // NOTE: this implies malicious behavior as the unit's session_id
             // is incompatible with session_id of the message it arrived in.
-            debug!(target: "rush-member", "A unit with incorrect session_id! {:?}", su.unchecked);
+            debug!(target: "rush-member", "A unit with incorrect session_id! {:?}", su.as_unchecked());
             return false;
         }
-        if !su.verify(su.signed().creator()) {
-            debug!(target: "rush-member", "A unit with incorrect signature! {:?}", su.unchecked);
+        if su.as_signable().round() > self.store.limit_per_node() {
+            debug!(target: "rush-member", "A unit with too high round {}! {:?}", su.as_signable().round(), su.as_unchecked());
             return false;
         }
-        if su.signed().round() > self.store.limit_per_node() {
-            debug!(target: "rush-member", "A unit with too high round {}! {:?}", su.signed().round(), su.unchecked);
-            return false;
-        }
-        if su.signed().creator().0 >= self.config.n_members.0 {
-            debug!(target: "rush-member", "A unit with too high creator index {}! {:?}", su.signed().creator(), su.unchecked);
+        if su.as_signable().creator().0 >= self.config.n_members.0 {
+            debug!(target: "rush-member", "A unit with too high creator index {}! {:?}", su.as_signable().creator(), su.as_unchecked());
             return false;
         }
         if !self.validate_unit_parents(su) {
-            debug!(target: "rush-member", "A unit did not pass parents validation. {:?}", su.unchecked);
+            debug!(target: "rush-member", "A unit did not pass parents validation. {:?}", su.as_unchecked());
             return false;
         }
         true
@@ -370,7 +364,7 @@ where
 
     fn add_unit_to_store_unless_fork(&mut self, su: SignedUnit<'a, H, D, KB>) {
         if let Some(sv) = self.store.is_new_fork(&su) {
-            let creator = su.signed().creator();
+            let creator = su.as_signable().creator();
             if !self.store.is_forker(creator) {
                 // We need to mark the forker if it is not known yet.
                 let proof = ForkProof {
@@ -383,20 +377,20 @@ where
             // There is no point in keeping this unit in any kind of buffer.
             return;
         }
-        let u_round = su.signed().round();
+        let u_round = su.as_signable().round();
         let round_in_progress = self.store.get_round_in_progress();
         if u_round <= round_in_progress + ROUNDS_MARGIN {
             self.store.add_unit(su, false);
         } else {
-            debug!(target: "rush-member", "Unit {:?} ignored because of too high round {} when round in progress is {}.", su.unchecked, u_round, round_in_progress);
+            debug!(target: "rush-member", "Unit {:?} ignored because of too high round {} when round in progress is {}.", su.as_unchecked(), u_round, round_in_progress);
         }
     }
 
     fn move_units_to_consensus(&mut self) {
         let mut units = Vec::new();
         for su in self.store.yield_buffer_units() {
-            let hash = su.signed().hash();
-            let unit = Unit::new_from_preunit(su.signed().inner.clone(), hash);
+            let hash = su.as_signable().hash();
+            let unit = Unit::new_from_preunit(su.as_signable().inner.clone(), hash);
             units.push(unit);
         }
         if !units.is_empty() {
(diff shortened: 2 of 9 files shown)


src/signed.rs
```diff
@@ -0,0 +1,68 @@
+use crate::{Index, KeyBox};
+use codec::{Decode, Encode};
+
+pub trait Signable {
```
mike1729, line 4: How about adding `AsRef<[u8]>` instead of this trait?

src/member.rs
```diff
@@ -628,21 +651,26 @@ where
...
         use ConsensusMessage::*;
         match message {
-            NewUnit(signed_unit) => {
-                debug!(target: "rush-member", "New unit received {:?}.", signed_unit);
-                self.on_unit_received(signed_unit, false);
+            NewUnit(unchecked) => {
+                debug!(target: "rush-member", "New unit received {:?}.", unchecked);
+                if let Some(su) = Signed::from(unchecked, self.keybox) {
+                    self.on_unit_received(su, false);
+                }
```
mike1729, line 663: We should note a wrong signature.

src/member.rs
```diff
@@ -628,21 +651,26 @@ where
...
             RequestCoord(coord) => {
                 self.on_request_coord(peer_id, coord);
             }
-            ResponseCoord(signed_unit) => {
-                debug!(target: "rush-member", "Fetch response received {:?}.", signed_unit);
-                self.on_unit_received(signed_unit, false);
+            ResponseCoord(unchecked) => {
+                debug!(target: "rush-member", "Fetch response received {:?}.", unchecked);
+
+                if let Some(su) = Signed::from(unchecked, self.keybox) {
```
mike1729, line 671: Same here.

src/signed.rs
```diff
@@ -0,0 +1,68 @@
...
+            unchecked: self.unchecked.clone(),
+            key_box: self.key_box,
+        }
+    }
+}
+
+impl<'a, T: Signable + Index, KB: KeyBox> Signed<'a, T, KB> {
+    pub fn sign(key_box: &'a KB, signable: T) -> Self {
+        let signature = key_box.sign(&signable.bytes_to_sign());
+        let signed = signable;
```
mike1729, line 32: Why this temp var is created?

src/signed.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    pub fn from(unchecked: UncheckedSigned<T, KB::Signature>, key_box: &'a KB) -> Option<Self> {
+        let signed = Signed { unchecked, key_box };
+        if !signed.verify() {
+            return None;
+        }
+        Some(signed)
+    }
+
+    pub(crate) fn verify(&self) -> bool {
```
mike1729, line 50: It seems that this method is called only after `Signed::from` was called. Maybe we could make all fields in `Signed` private and leave just `from` as a constructor?

src/signed.rs
```diff
@@ -0,0 +1,68 @@
...
+        Signed {
+            unchecked: UncheckedSigned {
+                signable: signed,
+                signature,
+            },
+            key_box,
+        }
+    }
+
+    pub fn from(unchecked: UncheckedSigned<T, KB::Signature>, key_box: &'a KB) -> Option<Self> {
```
mike1729, line 42: It should be called `try_from`.

src/bft.rs
```diff
@@ -1,16 +1,16 @@
-use crate::{units::SignedUnit, Data, Hasher, NodeIndex};
+use crate::{units::UncheckedSignedUnit, Data, Hasher, NodeIndex};
 use codec::{Decode, Encode};
 
 #[derive(Debug, Encode, Decode)]
 pub(crate) struct ForkProof<H: Hasher, D: Data, Signature: Clone + Encode + Decode> {
-    pub(crate) u1: SignedUnit<H, D, Signature>,
-    pub(crate) u2: SignedUnit<H, D, Signature>,
+    pub(crate) u1: UncheckedSignedUnit<H, D, Signature>,
+    pub(crate) u2: UncheckedSignedUnit<H, D, Signature>,
```
timorl, line 7: Semantically this is wrong – a `ForkProof` should refer to two properly signed units, not the unchecked variants.

src/member.rs
```diff
@@ -332,77 +326,80 @@ where
...
+        if su.signed().session_id != self.config.session_id {
             // NOTE: this implies malicious behavior as the unit's session_id
             // is incompatible with session_id of the message it arrived in.
-            debug!(target: "rush-member", "A unit with incorrect session_id! {:?}", su);
+            debug!(target: "rush-member", "A unit with incorrect session_id! {:?}", su.unchecked);
             return false;
         }
-        if !su.verify_signature(&self.keybox) {
-            debug!(target: "rush-member", "A unit with incorrect signature! {:?}", su);
+        if !su.verify() {
```
timorl, line 347: Shouldn't the unit already by verified by the virtue of its type?

src/member.rs
```diff
@@ -653,6 +681,10 @@ where
                 // TODO: these responses are quite heavy, we should at some point add
                 // checks to make sure we are not processing responses to request we did not make.
                 // TODO: we need to check if the response (and alert) does not exceed some max message size in network.
+                let parents = parents
+                    .into_iter()
+                    .filter_map(|unchecked| Signed::from(unchecked, self.keybox))
```
timorl, line 686: In the context of Michał's comments above we should probably also note the wrong signatures here somehow, although it's much more annoying.

src/signed.rs
timorl: `AsRef` is supposed to be for cheap conversions and here we want to perform some encoding. Perhaps `Into<[u8]>` could work, but the semantics are a bit different – perhaps we only want a part of the whole object (in our case maybe the hash?) signed, not the whole representation for whatever reason?

src/signed.rs
```diff
@@ -0,0 +1,68 @@
...
+        Signed {
+            unchecked: self.unchecked.clone(),
+            key_box: self.key_box,
+        }
+    }
+}
+
+impl<'a, T: Signable + Index, KB: KeyBox> Signed<'a, T, KB> {
+    pub fn sign(key_box: &'a KB, signable: T) -> Self {
+        let signature = key_box.sign(&signable.bytes_to_sign());
```
timorl, line 31: We should check here that `signable.index()` agrees with our index – I'm not sure whether the `KeyBox` is `Index`, but it should be.

src/signed.rs
timorl, line 50: I don't like the semantics here. I would rather have a method `check` on `UncheckedSigned`, which would return `Option<Signed>`, and call that in `from` above. The only ways of creating a `Signed` would create an already verified signature, so this method would not be necessary – that was the whole point behind making it into a separate type.

src/signed.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    pub(crate) fn verify(&self) -> bool {
+        let signed = &self.unchecked.signable;
+        let index = signed.index().unwrap();
+        self.key_box
+            .verify(&signed.bytes_to_sign(), &self.unchecked.signature, index)
+    }
+
+    pub(crate) fn signed(&self) -> &T {
```
timorl, line 57: Well, this name was confusing me throughout the whole review. Maybe `contents` or `(signed_?)data` or something? I'm not sure what would be a good name, but I'm pretty sure `signed` suggests that the returned thing actually contains the signature.

src/signed.rs
```diff
@@ -0,0 +1,68 @@
+use crate::{Index, KeyBox};
+use codec::{Decode, Encode};
+
+pub trait Signable {
+    fn bytes_to_sign(&self) -> Vec<u8>;
+}
+
+#[derive(Clone, Debug, Decode, Encode)]
+pub struct UncheckedSigned<T: Signable + Index, S> {
```
timorl, line 9: I would parametrize here by `KB` rather than just by `S`. Even if we don't technically use `KB`, in principle an `UncheckedSigned` is associated with a signing scheme, not just the signature type. And if you follow my suggestion about the `check` method below, it would actually really depend on `KB`.

Hm, after thinking this through before clicking "Finish" – is the problem here that if we parametrize by `KB` we cannot derive all the traits above?

timorl (changes requested):

src/signed.rs
mike1729: Hmm...so maybe `sign<S, F: Fn(&[u8]) -> S>(&self, signing: F) -> S` ?

timorl: Making `bytes_to_sign` return `&[u8]` seems simpler (can we do that?). We could also autoimplement for anything that is `AsRef<[u8]>` if we really want to.

mike1729: It would be way simpler, but we cannot return references :(

timorl: ...how about we return `impl AsRef<[u8]>` then?

mike1729: But then we need to create a temporary object that will implement this trait and the idea (at least mine) was to change this trait in a way that avoids it.

timorl: I thought `&[u8]` automatically implements `AsRef<[u8]>`, doesn't it? Then we could return `as_ref` directly for types that implement it.

mike1729: It does but still borrow checker won't let you return type `&T`.

src/bft.rs
michalseweryn: Yes, but... We want this type to implement Encode + Decode, which is unimplementable for a struct storing a reference to a keybox...

src/member.rs
michalseweryn: Fair point

src/signed.rs
michalseweryn: It does implement `Index` buti I don't think it should...

src/signed.rs
michalseweryn: Agree

src/signed.rs
michalseweryn: We can always implement the traits manually.

src/signed.rs
michalseweryn: Is there still a scenario, where we do not want to sign_to_bytes is essentially different from encode()? maybe we should just replace Signable with Encode?

timorl: Then we are always copying, aren't we? I think in this case it would be nicer to keep it separate, to keep the semantics clear, but I won't insist.

Whatever we end up with I would prefer avoiding the `sign<...` option, it also requires a `verify<...` function, and both of these have to be implemented by the author of the type to be signed and would always have to contain almost identical logic, with the exception of how it acquires the `&[u8]`. This would cause lots of code duplication or worse, actual differences in the logic in the implementations.

src/bft.rs
timorl: Right, so it would require a separate `UncheckedForkProof` type... Probably not worth it, what do you think?

src/signed.rs
timorl: It only has one private key, the one corresponding to `Index`, so I think the meaning is reasonable. Here it adds the possibility for some validation – if for whatever reason we tried signing something created by someone else, we would get an explicit error instead of massive confusion in the consensus algorithm when someone sends such an object.

If you don't want that check here then maybe implement this signing procedure only for `T` without `Index`? Then the semantics would be clear again – just sign the bytes, there cannot be a mismatch with an index if the index is not officially there.

src/signed.rs
timorl: If we want to keep the possibility of avoiding copying then we can either do [some magic](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&code=trait%20Signable%3C%27b%2C%20B%3A%20AsRef%3C%5Bu8%5D%3E%3E%20%7B%0A%20%20%20%20%20fn%20bytes_to_sign(%26%27b%20self)%20-%3E%20B%3B%0A%7D%0A%0Astruct%20Wrapper%3C%27a%2C%20T%3A%20AsRef%3C%5Bu8%5D%3E%3E%20%7B%0A%20%20%20%20data%3A%20%26%27a%20T%2C%0A%7D%0A%0Aimpl%3C%27a%2C%20T%3A%20AsRef%3C%5Bu8%5D%3E%3E%20AsRef%3C%5Bu8%5D%3E%20for%20Wrapper%3C%27a%2C%20T%3E%20%7B%0A%20%20%20%20fn%20as_ref(%26self)%20-%3E%20%26%5Bu8%5D%20%7B%0A%20%20%20%20%20%20%20%20self.data.as_ref()%0A%20%20%20%20%7D%0A%7D%0A%0Aimpl%3C%27b%2C%20T%3A%20AsRef%3C%5Bu8%5D%3E%3E%20Signable%3C%27b%2C%20Wrapper%3C%27b%2C%20T%3E%3E%20for%20T%20%7B%0A%20%20%20%20fn%20bytes_to_sign(%26%27b%20self)%20-%3E%20Wrapper%3C%27b%2C%20T%3E%20%7B%0A%20%20%20%20%20%20%20%20Wrapper%7Bdata%3A%20self%7D%0A%20%20%20%20%7D%0A%7D%0A%0Astruct%20Hello%3B%0A%0Aimpl%3C%27b%3E%20Signable%3C%27b%2C%20Vec%3Cu8%3E%3E%20for%20Hello%20%7B%0A%20%20%20%20fn%20bytes_to_sign(%26%27b%20self)%20-%3E%20Vec%3Cu8%3E%20%7B%0A%20%20%20%20%20%20%20%20%22Hello%22.as_bytes().to_vec()%0A%20%20%20%20%7D%0A%7D%0A%0Astruct%20DummyContainer%3C%27a%2C%20T%3A%20Signable%3C%27a%2C%20Vec%3Cu8%3E%3E%3E%20%7B%0A%20%20%20%20hi%3A%20%26%27a%20T%2C%0A%7D%0A%0Afn%20main()%20%7B%0A%20%20%20%20let%20hi%20%3D%20DummyContainer%7Bhi%3A%20%26Hello%7B%7D%7D%3B%0A%20%20%20%20println!(%22%7B%3A%3F%7D%22%2Chi.hi.bytes_to_sign())%3B%0A%20%20%20%20println!(%22%7B%3A%3F%7D%22%2Chi.hi.bytes_to_sign())%3B%0A%7D), or force the implementer to figure out how to return a reference (e.g. by keeping the bytes to sign as part of the struct).

michalseweryn: Note that the `Signable` trait is not in the public interface of rush, and if I am not wrong, the only implementors of this trait which we plan for now are (full) units and alerts. Even if we only look at the `FullUnit` type, it seems unavoidable to create a temporary object as we need to call the `encode` method on the data of the unit. Unless we have a reason to think that this will be the bottleneck of rush, I think we should go with the simplest solution, which is making the `sign_to_bytes` method return `Vec<u8>`.

src/signed.rs
michalseweryn: The idea was that semantically `signable` is no longer just an instance of the `Signable`, but some something which is signed. But now that I look at it i agree that it does not seem reasonable given how simple the function is.

src/signed.rs
michalseweryn: Fixed.

src/bft.rs
michalseweryn: Yes, I think this is reasonable to only have an "unchecked" variant

src/member.rs
michalseweryn: Turns out that `Result<Vec<T>, E>` [implements](https://doc.rust-lang.org/std/result/enum.Result.html#impl-FromIterator%3CResult%3CA%2C%20E%3E%3E) `FromIterator<Result<T, E>>`, which gives a quite nice solution to catch an error if there is any.

src/signed.rs
michalseweryn: The type `T` is no longer required to implement the `Index` trait, so we don't have to ensure that the indices match

src/signed.rs
mike1729: Unless we sign a hash of unit, which would save us calling `encode` and creating a temporary object (we should change what we sign regardless of what will happen to `Signable` trait). Then, the easiest solution would be to use `AsRef`, probably without even creating a new trait, just requiring it as a bound in `Signed`. I believe that it fits there pretty well.

src/signed.rs
michalseweryn: `as_signable` seems to follow the conventions

src/signed.rs
michalseweryn: I tried it, and if we want to parameterize `UncheckedSigned` by the key box, we need to add a ton of boilerplate, which I don't think we should do. I am going it as is

src/signed.rs
michalseweryn: The method no longer exists (I followed a [suggestion](https://github.com/Cardinal-Cryptography/rush/pull/55#discussion_r632381593) of timorl)

src/signed.rs
michalseweryn: We decided to give `Signable` a single method  `fn hash(&self) -> Self::Hash`and to sign an object means exactly to sign its hash.

src/member.rs
michalseweryn: Fixed

src/member.rs
michalseweryn: Fixed

PR #60, opened 2021-05-19
Refactor units

michalseweryn:
Refactor of the module `units`. The following changes are introduced:

* the implementation of `UnitStore` is moved to a new separate file `src/units/store.rs`,
* the structs in the module are reordered in a more natural order.

* the coordinates of the unit (`UnitCoord`) now store the round number as `u16` instead of `u64`.
* every struct which had fields `round: Round` and `creator: NodeIndex` has them now replaced with `coord: UnitCoord`
* the type of the set of parents in `ControlHash` is changed from `Vec<bool>` with `bit_vec::BitVec<u32>` which consists of roughly N bits.
* in order to standardize the interface, all fields of the `*Unit` structs are made private and are accesed via getters. For convenience, we also have getters to access fields of fields, so that instead of `full_unit.pre_unit.coord.round` we now can write `full_unit.round()`.
* the hash of a `FullUnit` is calculated once on creation and stored in a field of `FullUnit`. The hash is not included in the encoding of `FullUnit`
* removed "helper" functions, which do not make reading the code any simpler, and are used for example only once.

### DIFF
--- a/src/member.rs
+++ b/src/member.rs
@@ -447,15 +447,11 @@ where
         let (u_round, u_control_hash, parent_ids) = match self.store.unit_by_hash(&u_hash) {
             Some(su) => {
                 let full_unit = su.as_signable();
-                let parents = &full_unit.control_hash().parents_mask;
-                let parents = (0..parents.len()).into_iter().map(|i| parents[i]);
+                let parents: Vec<_> = full_unit.control_hash().parents().collect();
                 (
                     full_unit.round(),
-                    full_unit.control_hash().hash,
-                    parents
-                        .enumerate()
-                        .filter_map(|(i, b)| if b { Some(i.into()) } else { None })
-                        .collect::<Vec<NodeIndex>>(),
+                    full_unit.control_hash().combined_hash,
+                    parents,
                 )
             }
             None => {
--- a/src/member.rs
+++ b/src/member.rs
@@ -166,11 +166,7 @@ where
     fn on_create(&mut self, u: PreUnit<H>) {
         debug!(target: "rush-member", "On create notification.");
         let data = self.data_io.get_data();
-        let full_unit = FullUnit {
-            inner: u,
-            data,
-            session_id: self.config.session_id,
-        };
+        let full_unit = FullUnit::new(u, data, self.config.session_id);
         let hash = full_unit.hash();
         // TODO: beware: sign_unit blocks and is quite slow!
         let signed_unit = Signed::sign(self.keybox, full_unit);
@@ -312,7 +308,7 @@ where
     fn validate_unit_parents(&self, su: &SignedUnit<'a, H, D, KB>) -> bool {
         // NOTE: at this point we cannot validate correctness of the control hash, in principle it could be
         // just a random hash, but we still would not be able to deduce that by looking at the unit only.
-        let pre_unit = &su.as_signable().inner;
+        let pre_unit = su.as_signable().as_pre_unit();
         if pre_unit.n_members() != self.config.n_members {
             debug!(target: "rush-member", "Unit with wrong length of parents map.");
             return false;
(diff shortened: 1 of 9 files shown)


src/member.rs
```diff
@@ -453,15 +447,14 @@ where
         let (u_round, u_control_hash, parent_ids) = match self.store.unit_by_hash(&u_hash) {
             Some(su) => {
                 let full_unit = su.as_signable();
+                let parents = &full_unit.control_hash().parents_mask;
+                let parents = (0..parents.len()).into_iter().map(|i| parents[i]);
```
mike1729, line 451: Why not just `let parents = full_unit.control_hash().parents_mask.iter();`?

src/units.rs
```diff
@@ -1,116 +1,109 @@
...
     }
 }
 
-impl<H: Hasher> From<&Unit<H>> for UnitCoord {
-    fn from(unit: &Unit<H>) -> Self {
-        UnitCoord {
-            creator: unit.creator(),
-            round: unit.round() as u64,
-        }
+impl<H: Hasher> Encode for ControlHash<H> {
```
mike1729, line 72: Add a test checking if composition of encode and decode is an identity.

src/units.rs
```diff
@@ -1,116 +1,109 @@
...
-
-#[derive(Clone, Debug, PartialEq, Encode, Decode)]
+/// The simplest type representing a unit, consisting
+#[derive(Clone, Debug, Default, PartialEq, Encode, Decode)]
 pub(crate) struct PreUnit<H: Hasher> {
-    pub(crate) creator: NodeIndex,
-    pub(crate) round: UnitRound,
-    pub(crate) control_hash: ControlHash<H>,
+    coord: UnitCoord,
+    // round is stored as `u16` not `Round = usize` for efficiency. Also, `usize` does not implement
```
mike1729, line 98: Maybe the conversion should happen in encoding? Not sure because most probably compiler is smart enough to optimize this anyway.

src/units.rs
```diff
@@ -122,231 +115,128 @@ impl<H: Hasher> PreUnit<H> {
...
-#[derive(Clone, Debug, Default, PartialEq, Encode, Decode)]
-pub(crate) struct Unit<H: Hasher> {
-    pub(crate) creator: NodeIndex,
-    round: UnitRound,
-    pub(crate) hash: H::Hash,
-    pub(crate) control_hash: ControlHash<H>,
+///
+#[derive(Debug, Default, Clone, Encode, Decode)]
+pub(crate) struct FullUnit<H: Hasher, D: Data> {
+    pre_unit: PreUnit<H>,
```
mike1729, line 133: Why not store Unit and skip encoding hash there?

mike1729 (changes requested):

src/member.rs
michalseweryn: Given what we do later, it becomes even simpler:

`let parents: Vec<_> = full_unit.control_hash().parents().collect()`

src/units.rs
michalseweryn: Done

src/units.rs
michalseweryn: I would rather avoid implementing `Encode + Decode` manually. There is an attribute `#[codec(encoded_as = "OtherType")]`, but it requres implementing `EncodeAsRef` trait, and that would require introducing wrappers (implementing `EncodeAsRef<'_, usize>` for `u16` is just wrong).

src/units.rs
michalseweryn: Then we would have to initialize the hash of that Unit after constructing the FullUnit, which seems like a very bad practice.
Also, for me the semantics of codec(skip) is that the field is reduntant and can be inferred from other fields. And in Unit, this does not hold for the field hash. My solution is the lmost logical I can see.

src/units.rs
michalseweryn: I misunderstood you intially, and added such a test for FullUnit. Now there is a test for ControlHash as well.

PR #61, opened 2021-05-20
Byzantine tests

DamianStraszak:
Tests for Member, including tests with a "byzantine Member".
What these tests do is essentially checking if Member works correctly, i.e., that if a set of Members connected via network (mock in this case) achieves their intended goal, i.e., consistently order some pieces of "Data".
There are test scenarios with honest Members only, and ones where there are byzantine Members, who make unit forks.
Currently the slowest test runs quite long (i.e., depending on randomness between 20s - 30s):
1) This is because there are hardcoded delays of the order of "a couple of secs", for replaying requests for parents (when there is wrong control hash detected etc.). One TODO would be to add these delays to Member configuration.
2) Another reason is that requests at the moment are rather inefficient, in the sense that we request from a random node, and only once per couple of seconds. This is to be improved.
3) In case you wonder: why is there randomness in these tests? Why not make it deterministic? There is no way to make it truly deterministic, because of concurrency which causes a few threads to race.

### DIFF
--- a/src/testing/byzantine.rs
+++ b/src/testing/byzantine.rs
@@ -1,0 +1,297 @@
+use codec::{Decode, Encode};
+
+use log::{debug, error};
+
+use futures::{channel::oneshot, FutureExt, StreamExt};
+
+use std::collections::HashMap;
+
+use crate::{
+    nodes::NodeMap,
+    signed::Signed,
+    testing::mock::{
+        configure_network, init_log, node_ix_to_peer, spawn_honest_member, AlertHook, Data,
+        Hasher64, KeyBox, Network, PeerId, Signature, Spawner,
+    },
+    units::{ControlHash, FullUnit, PreUnit, SignedUnit, UnitCoord},
+    Hasher, Network as NetworkT, NetworkCommand, NetworkEvent, NodeCount, NodeIndex, SessionId,
+    SpawnHandle,
+};
+
+type Hash64 = <Hasher64 as Hasher>::Hash;
+
+use crate::member::{ConsensusMessage, ConsensusMessage::*};
+
+struct MaliciousMember<'a> {
+    node_ix: NodeIndex,
+    n_members: NodeCount,
+    threshold: NodeCount,
+    session_id: SessionId,
+    forking_round: usize,
+    round_in_progress: usize,
+    keybox: &'a KeyBox,
+    network: Network,
+    unit_store: HashMap<UnitCoord, SignedUnit<'a, Hasher64, Data, KeyBox>>,
+}
+
+impl<'a> MaliciousMember<'a> {
+    fn new(
+        keybox: &'a KeyBox,
+        network: Network,
+        node_ix: NodeIndex,
+        n_members: NodeCount,
+        session_id: SessionId,
+        forking_round: usize,
+    ) -> Self {
+        let threshold = (n_members * 2) / 3 + NodeCount(1);
+        MaliciousMember {
+            node_ix,
+            n_members,
+            threshold,
+            session_id,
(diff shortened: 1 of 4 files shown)


src/testing/byzantine.rs
```diff
@@ -0,0 +1,409 @@
...
+    }
+
+    fn pick_parents(&self, round: usize) -> Option<NodeMap<Option<Hash64>>> {
+        // Outputs a parent map if there are enough of them to create a new unit.
+        let mut parents = NodeMap::new_with_len(self.n_members);
+        if round == 0 {
+            return Some(parents);
+        }
+        let mut count = 0;
+        let mut our_unit_available = false;
```
mike1729, line 79: Maybe check it straight away? I'm not sure how often pick_parents misses, though.

src/testing/byzantine.rs
```diff
@@ -0,0 +1,409 @@
...
+fn gen_config(ix: NodeIndex, n_members: NodeCount) -> Config {
+    Config {
+        node_id: ix,
+        session_id: 0,
+        n_members,
+        create_lag: Duration::from_millis(100),
+    }
+}
+
+type TestMember<'a> = member::Member<'a, Hasher64, Data, DataIO, KeyBox, Network>;
```
mike1729, line 233: HonestMember?

mike1729 (changes requested):
I propose the following code reorganization:
- move `init_log`, `gen_config`, `configure_network`, and `spawn_honest_member` to `mock.rs`,
- create a new file `crash.rs` and move `honest_members_agree_on_batches` and related test there,
- leave the rest in `byzantine.rs`.

src/testing/mock.rs
```diff
@@ -175,4 +181,353 @@ impl Spawner {
...
+        loop {
+            let rand_peer = self
+                .peer_list
+                .choose(&mut rng)
+                .expect("Peer list must be non-empty.")
+                .clone();
+            if rand_peer != self.my_peer_id {
+                return rand_peer;
+            } else {
+                continue;
```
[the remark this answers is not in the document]
DamianStraszak: Somehow it didn't, fixed :)

src/testing/mock.rs
```diff
@@ -175,4 +181,353 @@ impl Spawner {
...
+        peer: PeerId,
+        reliable: bool,
+    ) -> Result<(), ()> {
+        let addr_msg = AddressedMessage {
+            message: msg,
+            sender: self.my_peer_id.clone(),
+            receiver: peer,
+            reliable,
+        };
+        match self.tx.unbounded_send(addr_msg) {
```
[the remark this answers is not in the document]
DamianStraszak: Done.

src/testing/mock.rs
```diff
@@ -175,4 +181,353 @@ impl Spawner {
...
+    }
+}
+
+#[derive(Hash, Debug, Default, Clone, PartialEq, Eq, Encode, Decode)]
+pub(crate) struct Signature {}
+
+pub(crate) struct DataIO {
+    ix: NodeIndex,
+    //Arc + Mutex because get_data gets immutable self
+    round_counter: Arc<Mutex<usize>>,
```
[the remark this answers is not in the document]
DamianStraszak: Changed to Cell.

src/testing/byzantine.rs
DamianStraszak: Added this check -- it avoids the loop now, so it's better for sure.

src/testing/byzantine.rs
DamianStraszak: Makes sense, changed.

PR #64, opened 2021-05-21
Implement BoolNodeMap and multisigning tools

michalseweryn:
`BoolNodeMap` is a struct with the same functionality as `NodeMap<bool>` but with efficient encoding.

`MultiKeychain` is a new trait extending `KeyBox` with multisigning functionality. Also, as an equivalent of `Signed<'a, T, KB>` for multisignatures, two struct are introduced: `PartiallyMultisigned<'a, T, MK>` and
`MultiSigned<'a, T, MK>`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -19,7 +19,8 @@ mod creator;
 mod extender;
 mod member;
 pub mod nodes;
-pub mod signed;
+mod signed;
+pub use signed::*;
 mod terminal;
 mod testing;
 mod units;
@@ -28,12 +29,6 @@ pub trait DataIO<Data> {
     type Error: Debug;
     fn get_data(&self) -> Data;
     fn send_ordered_batch(&mut self, data: OrderedBatch<Data>) -> Result<(), Self::Error>;
-}
-
-pub trait KeyBox: Index {
-    type Signature: Debug + Clone + Encode + Decode;
-    fn sign(&self, msg: &[u8]) -> Self::Signature;
-    fn verify(&self, msg: &[u8], sgn: &Self::Signature, index: NodeIndex) -> bool;
 }
 
 #[async_trait::async_trait]
--- a/src/nodes.rs
+++ b/src/nodes.rs
@@ -1,4 +1,4 @@
-use codec::{Decode, Encode, Error as CodecError, Input, Output};
+use codec::{Decode, Encode, Error as CodecError, Error, Input, Output};
 use derive_more::{Add, AddAssign, Display, From, Into, Sub, SubAssign, Sum};
 use std::{
     iter::FromIterator,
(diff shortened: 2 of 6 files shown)


src/lib.rs
```diff
@@ -36,6 +36,23 @@ pub trait KeyBox: Index {
...
+        index: NodeIndex,
+        signature: &Self::Signature,
+    );
+    fn is_complete(partial: &Self::PartialMultisignature) -> bool;
+    fn verify_partial(
+        &self,
+        msg: &[u8],
+        partial: &Self::PartialMultisignature,
+        indices: &BoolNodeMap,
+    ) -> bool;
```
DamianStraszak, line 53: Only a comment, not a change request: it might turn out that the interface with verify_partial is not ideal. In all schemes I'm aware of, one verifies each individual signature of a multisignature separately. So something like
```
verify_share(&self, msg: &[u8], s: Self::Signature, index:usize)
```
might be better. Anyway, this is easy to change later.

src/nodes.rs
```diff
@@ -135,17 +135,87 @@ impl<'a, T> IntoIterator for &'a NodeMap<T> {
...
         }
     }
+
+    #[test]
+    fn decoding_bool_node_map_works() {
+        let bool_node_map = BoolNodeMap([true, false, true, true, true].iter().cloned().collect());
+        let encoded: Vec<_> = bool_node_map.encode();
+        let decoded = BoolNodeMap::decode(&mut encoded.as_slice()).expect("decode should work");
+        assert_eq!(decoded, bool_node_map);
+    }
```
DamianStraszak, line 220: Can we have one additional test that checks that the encoding of a BoolNodeMap of length N is close to N/8 rather than N? Just to be sure :P

src/nodes.rs
```diff
@@ -135,17 +135,87 @@ impl<'a, T> IntoIterator for &'a NodeMap<T> {
...
+
+    pub fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = bool> + '_ {
+        self.0.iter()
+    }
+
+    pub fn true_indices(&self) -> impl Iterator<Item = NodeIndex> + '_ {
```
mike1729, line 162: how about `node_indices` or just `nodes`?

src/lib.rs
mike1729: Looking at the implementation of `PartiallyMultisigned::add_signature` line 146, I believe that the author meant `verify` that is inherited by `MultiKeychain` from `KeyBox` to be used as `verify_share` that you propose. And only check `is_complete` on the final multisignature.

src/lib.rs
```diff
@@ -36,6 +36,23 @@ pub trait KeyBox: Index {
...
 
+pub trait MultiKeychain: KeyBox {
+    type PartialMultisignature: Debug + Clone + Encode + Decode;
+    fn get_signers(partial: &Self::PartialMultisignature) -> BoolNodeMap;
+    fn add_signature(
+        partial: &mut Self::PartialMultisignature,
+        index: NodeIndex,
+        signature: &Self::Signature,
+    );
+    fn is_complete(partial: &Self::PartialMultisignature) -> bool;
```
mike1729, line 47: It seems that `get_signers`, `add_signature`, and `is_complete` could form a trait `PartialMultisignature` as those methods don't depend on `MultiKeychain`. Is the current version more convenient for some reason?

src/signed.rs
```diff
@@ -98,3 +119,59 @@ impl<'a, T: Signable, KB: KeyBox> From<Signed<'a, T, KB>> for UncheckedSigned<T,
...
+    pub indices: BoolNodeMap,
+    pub keychain: &'a MK,
+}
+
+pub struct IncompleteMultisignatureError<'a, T: Signable, MK: MultiKeychain> {
+    pub partial: PartiallyMultisigned<'a, T, MK>,
+}
+
+impl<'a, T: Signable, MK: MultiKeychain> PartiallyMultisigned<'a, T, MK> {
+    pub fn add_signature(&mut self, signature: &MK::Signature, index: NodeIndex) {
```
mike1729, line 140: There is not much code here but in a sense a lot happens here and a small mistake can have serious implications. Could you add a test? I understand that it requires a mock of `MultiKeychain` but the mock will be helpful in other tests as well.

src/signed.rs
```diff
@@ -37,9 +43,24 @@ impl<T: Signable, S> UncheckedSigned<T, S> {
             key_box,
         })
     }
+
+    pub fn check_partial<'a, MK: MultiKeychain<PartialMultisignature = S>>(
+        self,
+        keychain: &'a MK,
+        indices: &BoolNodeMap,
+    ) -> Result<PartiallyMultisigned<'a, T, MK>, SignatureError<T, S>> {
+        if !keychain.verify_partial(self.signable.hash().as_ref(), &self.signature, indices) {
```
mike1729, line 52: Shouldn't it be just verify in fact?

mike1729 (changes requested):

src/lib.rs
michalseweryn: Yes, that was my intention. the MultiKeychain builds on top of KeyBox and adds signature aggrgation functionality.

src/nodes.rs
michalseweryn: I wanted to make it compatible with the interface of NodeMap<T>. I also considered renaming BoolNodeMap to NodeSet, but set does not suggest existence of the total count of the nodes. Also, the len function would be very confusing. And technically, the false'y nodes are still in the BoolNodeMap, but with false value ...
Maybe the best solution would be to indeed rename the struct to NodeSet...

src/lib.rs
michalseweryn: I wanted to avoid having an extra trait, but probably the extra trait would actually improve the readability

michalseweryn: On the side note, I can't see a scenario, where we use a PartialSignature without the set of node indices who have signed it, so it seems natural that the list of the signers is a part of the PartialSignature. Probably it would be more consistent to require that the Signature contains the index of the signer? This would simplify the interface in all places where we pass both the index and the signature. The only problem is a minor redundancy: when we sign a FullUnit, it already contains the index, but if we really don't like this redundancy, but maybe it would be better to remove it and only store it in the signature?

src/nodes.rs
mike1729: Maybe NodeSet with capacity instead of length?

src/lib.rs
mike1729: 1. In threshold signatures we don't need to track the list of signers
2. I like the idea of putting index into signature and changing full unit so that it doesn't carry the index.

src/nodes.rs
michalseweryn: Now that I think about it, there is no need for a capacity, we can simply define the capacity as i/8 + 1where i is the largest contained index (unless we really need to store the total number of nodes in the NodeSet, which seems a code smell)

src/lib.rs
michalseweryn: Ad 1. true, but I think we already gave up the threshold signatures.
Ad 2. Awesome, I will do it that way then.

mike1729: In what sense we gave up the threshold signatures? You mean to fit in this trait? We'll definitely use them for randomness.

src/nodes.rs
mike1729: In a scenario where NodeSet is used it's the only data_structure that carries the info on total number of parents. Maybe it doesn't fit to this abstraction but is needed in the application. For me, personally, NodeSet with capacity is a valid construct. I would even propose renaming `new_with_len` to `new_with_capacity` as it seems more natural when I think about it now.

michalseweryn: `new_with_capacity` sounds good. But I think, we usually have the access to the key(box|chain), which seems a more appropriate place to store the node count?

src/lib.rs
michalseweryn: Oh, I didn't get that part. @timorl kept sayng that the threshold signatures require too complicated set up, and we want to avoid using them. If we want to use them, then probably we don't want the `get_signers` method.

src/nodes.rs
mike1729: We use `n_members` from ControlHash in Terminal where we don't have access to the KeyBox. If you have an idea how to cleanly replace it with sth else, then the node count could disapear from NodeSet.

DamianStraszak: I think capacity is a good name, and `new_with_capacity` is the way to go. Having capacity != n_members might backfire at some point, as the code is written with this assumption in mind (at least the part for representing unit parents).

michalseweryn: I will name it `with_capacity` as in `Vec::with_capacity`

DamianStraszak: For the sake of Terminal I think it would not be a big deal to add a field n_members to Terminal. But generally the concern with not having fixed capacity are strange issues with encoding/hashing/validating units. We save a lot of effort (thinking about what the validation logic should be, what are the edge cases etc.) by specifying that parents map has fixed capacity.

src/nodes.rs
michalseweryn: Done :)

src/signed.rs
michalseweryn: Nope. The `UncheckedSigned` is used by both Signed and (Parially)Multisigned

src/nodes.rs
michalseweryn: Agreed. The BoolNodeMap has now capacity.

src/signed.rs
michalseweryn: Added a test and fixed a mistake :)

mike1729 (approved):

DamianStraszak (approved):

PR #66, opened 2021-05-21
Refactor Network

timorl:


src/member.rs
```diff
@@ -704,45 +748,43 @@ where
...
             .await
         });
         self.tx_consensus = Some(tx_consensus);
+        let (alert_sink, mut incoming_alerts) = mpsc::unbounded();
+        let (outgoing_alerts, alert_stream) = mpsc::unbounded();
+        let (unit_sink, mut incoming_units) = mpsc::unbounded();
+        let (outgoing_units, unit_stream) = mpsc::unbounded();
+        let (network_exit, exit_stream) = oneshot::channel();
+        spawn_handle.spawn("member/network", async move {
+            NetworkHub::new(network, unit_stream, unit_sink, alert_stream, alert_sink)
```
mike1729, line 782: Wouldn't it be clearer to construct NetworkHub in the constructor of Member, store it as Option, and take it here? It would also allow to store `self.outgoing_[units,alerts]` directly and get rid of unwraps in corresponding send methods.

mike1729 (changes requested):

src/network.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+pub enum Recipient {
+    Everyone,
+    Node(NodeIndex),
+}
+
+pub(crate) struct NetworkHub<N: Network, H: Hasher, D: Data, S: Encode + Decode> {
+    network: N,
+    outgoing_units: Receiver<(UnitMessage<H, D, S>, Recipient)>,
```
michalseweryn, line 32: Is it just me, or is calling a receiver an output (as in `outgoing`) counterintuitive?

src/member.rs
timorl: The unwraps won't happen after the full refactor is complete – a property of the general design I'm trying to work towards is that no one creates their own channels. `Member` would create the `outgoing_[units,alerts]` channels, but then pass them on to a component that would actually take care of unit-related messages (the parts of `Member` that currently do that would be moved to that component). In the end `Member` should just start all the other components and itself stop existing as a struct (the way `consensus` now just exists as `run`).

In general having `Option` fields in structs that mean "this is here always, but we cannot initialize it during construction" is always bad design, so I definitely did not want to leave it like that forever. I assumed it would be as temporary as `self.tx_consensus` which already follows this anti-pattern.

src/network.rs
timorl: Yup, very good point. These are outgoing units from the perspective of Rush as a whole, thus the name here, but I agree it's not good. How about `_to_send` and `_received` (with `units` and `alerts` as prefixes ofc)?

src/member.rs
mike1729: Ok, so I wait for the final version :)

timorl: Oh, I don't want to complete the full refactor in this PR, that would be a bit much. <_<"

mike1729: sure sure. It took me some time to fix network hub analog in aleph-node to not contain Options, so I know that it's best done on final version. Waiting for the next PRs then :)

src/member.rs
```diff
@@ -31,19 +32,24 @@ const INITIAL_MULTICAST_DELAY: time::Duration = time::Duration::from_secs(3);
...
     /// Response to a request by coord.
     ResponseCoord(UncheckedSignedUnit<H, D, S>),
     /// Request for the full list of parents of a unit.
-    RequestParents(H::Hash),
+    RequestParents(NodeIndex, H::Hash),
     /// Response to a request for a full list of parents.
     ResponseParents(H::Hash, Vec<UncheckedSignedUnit<H, D, S>>),
+}
+
+/// A message concerning alerts, either about a new alert or a request for it.
```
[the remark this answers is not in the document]
timorl: Yes, this type will be moved into the component responsible for alerts in next stages of the refactor, similarly `UnitMessage` will be moved into the component responsible for the consensus.

src/member.rs
```diff
@@ -117,37 +123,40 @@ pub struct Config {
...
     tx_consensus: Option<Sender<NotificationIn<H>>>,
     data_io: DP,
     keybox: &'a KB,
-    network: N,
     store: UnitStore<'a, H, D, KB>,
     requests: BinaryHeap<ScheduledTask<H>>,
     threshold: NodeCount,
+    n_members: NodeCount,
+    outgoing_units: Option<Sender<(UnitMessage<H, D, KB::Signature>, Recipient)>>,
+    outgoing_alerts: Option<Sender<(AlertMessage<H, D, KB::Signature>, Recipient)>>,
```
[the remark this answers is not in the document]
timorl: Alerts do not, that's why they have to be added. I think we sometimes might get some data that implies an alert happens, and then the alert system might want to send a request for someone's commitment? Then such a request would go to a specific node.

If you are relatively sure that the above example is wrong and all messages within the alert system will go to `Everyone` I can remove this – re-adding it shouldn't be hard if the need arises.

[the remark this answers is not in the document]
DamianStraszak: If this are some internal messages for the alert systems that will go to the "Alerts" component then the recipients should of course be there. I interpreted it as the stream of alerts for the future "Consensus" component, which does not send any addressed messages.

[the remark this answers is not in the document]
timorl: Ah, yes, that part is not here yet, and there obviously we will have proper alerts with no recipents.

src/member.rs
```diff
@@ -202,11 +211,54 @@ where
...
+        rand::thread_rng()
+            .gen_range(0..self.n_members.into())
+            .into()
+    }
+
+    fn index(&self) -> NodeIndex {
+        self.keybox.index()
+    }
+
+    fn send_units(&mut self, message: UnitMessage<H, D, KB::Signature>, peer_id: NodeIndex) {
```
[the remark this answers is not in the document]
timorl: Done.

src/network.rs
timorl: Might be better now.

src/testing/mock.rs
```diff
@@ -372,21 +310,20 @@ impl Future for NetworkHub {
...
                 debug!("Simulated network fail.");
                 continue;
             }
             let mut peers = self.peers.lock();
-            if let Some(peer) = peers.get_mut(&addr_msg.receiver.clone()) {
+            if let Some(peer) = peers.get_mut(&node_id) {
                 let hooks = self.hook_list.lock();
                 for hook in hooks.iter() {
-                    hook.update_state(addr_msg.clone());
+                    hook.update_state(data.clone());
```
DamianStraszak, line 323: It would be good to allow the hook to see not only the data, but the sender and receiver as well. This then allows to make sophisticated tests without the need to modify this Router mock.

DamianStraszak (changes requested):

src/testing/mock.rs
timorl: Done, this also somewhat improved the semantics of incoming messages, although this isn't used in any way.

src/network.rs
michalseweryn: Perfect!

PR #67, opened 2021-05-23
Honest node binary example

mike1729:


examples/honest.rs
```diff
@@ -0,0 +1,323 @@
...
+        if finalized.len() == n_finalized {
+            break;
+        }
+    }
+    close_member.send(()).expect("should send");
+    close_network.send(()).expect("should send");
+}
+
+#[derive(PartialEq, Eq, Clone, Debug)]
+struct Hasher64;
```
timorl, line 105: I doubt this is cryptographically secure.

examples/honest.rs
```diff
@@ -0,0 +1,323 @@
...
+                next_data: Arc::new(Mutex::new(0)),
+                finalized_tx,
+            },
+            finalized_rx,
+        )
+    }
+}
+
+#[derive(Clone, Debug, Encode, Decode)]
+struct Signature;
```
timorl, line 150: This is _definitely_ not cryptographically secure. :D

It's fine to have these, but I think it's a good idea to point out in comments what these types _actually_ should satisfy.

examples/honest.rs
```diff
@@ -0,0 +1,323 @@
...
+
+#[derive(Clone)]
+struct KeyBox {
+    index: NodeIndex,
+}
+
+impl rush::KeyBox for KeyBox {
```
timorl, line 157: We will also need `MultiKeychain` implemented.

examples/honest.rs
```diff
@@ -0,0 +1,323 @@
...
+}
+
+impl rush::Index for KeyBox {
+    fn index(&self) -> NodeIndex {
+        self.index
+    }
+}
+
+#[derive(Clone)]
+struct Spawner;
```
timorl, line 174: Shouldn't this somehow take care of making sure the threads are joined?

timorl (changes requested):
While having mock inputs seems fine, I think there should be comments explaining what is actually required from real types to get a secure algorithm.

examples/honest.rs
mike1729: I wasn't suppose to be, adding comment.

examples/honest.rs
mike1729: I wasn't suppose to be, adding comment.

examples/honest.rs
```diff
@@ -0,0 +1,323 @@
...
+    let (network, mut manager) = Network::new().await.unwrap();
+    let (close_network, exit) = oneshot::channel();
+    tokio::spawn(async move { manager.run(exit).await });
+
+    let (data_io, mut to_finalize) = DataIO::new();
+
+    let mut interval = tokio::time::interval(Duration::from_millis(500));
+    interval.tick().await;
+
+    while network.connected_peers() != n_members - 1 {
```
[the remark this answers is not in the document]
mike1729: I needed it in some prev version and forgot to remove. Thanks for spotting it!

examples/honest.rs
mike1729: This is implicitly done by `#[tokio::main]`, that's why we need to send close signal.

examples/honest.rs
mike1729: It will be added when the API changes.

README.md
```diff
@@ -3,3 +3,17 @@
 # rush
 
 rust implementation of Aleph protocol
+
+# examples
+
+There is a dummy implementation of an honest committee member that is not
+cryptographically secure and servers only as a working example of what has to be
+implemented and not how it should be implemented. For detailed information on
+what is required from a client using this library see [Damian's description](??).
```
timorl, line 12: While I very strongly approve of giving credit I think this should probably be referred to under a different name. :>

mike1729: Placeholder until Damian's PR is not merged :)

PR #70, opened 2021-05-25
Reliable Multicast Implementation

michalseweryn:
A simple implementation of RMC Box.

The struct `ReliableMulticast<'a, H: Signable + Hash, MK: MultiKeychain>` multicasts hashes of the type `H` and provides the public methods:

* `pub fn new(...) -> Self`,
* `pub fn start_rmc(&mut self, hash: H) -> Result<(), HashAlreadyExistsError>`,
* `pub async fn next_multisigned_hash(&mut self) -> Option<Multisigned<'a, H, MK>>`.
In order for the box to progress, the function `next_multisigned_hash` needs to be polled.

### DIFF
--- a/src/nodes.rs
+++ b/src/nodes.rs
@@ -7,7 +7,7 @@ use std::{
 };
 
 /// The index of a node
-#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, From)]
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash, From)]
 pub struct NodeIndex(pub usize);
 
 impl Encode for NodeIndex {
@@ -96,7 +96,7 @@ impl<T> IndexMut<NodeIndex> for NodeMap<T> {
     }
 }
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, Default, PartialEq)]
 pub struct BoolNodeMap(bit_vec::BitVec<u32>);
 
 impl BoolNodeMap {
--- a/src/rmc.rs
+++ b/src/rmc.rs
@@ -1,0 +1,263 @@
+use crate::{
+    nodes::{NodeCount, NodeIndex},
+    signed::{PartiallyMultisigned, Signable, Signed, UncheckedSigned},
+    Indexed, MultiKeychain, Multisigned, PartialMultisignature, Signature,
+};
+use async_trait::async_trait;
+use core::fmt::Debug;
+use futures::{
+    channel::mpsc::{unbounded, UnboundedReceiver, UnboundedSender},
+    FutureExt, StreamExt,
+};
+use log::debug;
+use std::{
+    cmp::Ordering,
+    collections::{BinaryHeap, HashMap},
+    hash::Hash,
+};
+use tokio::time;
+
+/// A message consists of either a signed (indexed) hash, or a multisigned hash.
+#[derive(Clone)]
+pub enum Message<H: Signable, S: Signature, M: PartialMultisignature> {
+    SignedHash(UncheckedSigned<Indexed<H>, S>),
+    MultisignedHash(UncheckedSigned<H, M>),
+}
+
+impl<H: Signable, S: Signature, M: PartialMultisignature> Message<H, S, M> {
+    pub fn hash(&self) -> &H {
+        match self {
+            Message::SignedHash(unchecked) => unchecked.as_signable().as_signable(),
+            Message::MultisignedHash(unchecked) => unchecked.as_signable(),
+        }
+    }
+}
+
+pub enum Task<H: Signable, MK: MultiKeychain> {
+    BroadcastMessage(Message<H, MK::Signature, MK::PartialMultisignature>),
+}
+
+#[async_trait]
+pub trait TaskScheduler<T> {
+    fn add_task(&mut self, task: T);
+    async fn next_task(&mut self) -> Option<T>;
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+struct ScheduledTask<T> {
+    scheduled_time: time::Instant,
+    task: T,
+    delay: time::Duration,
+}
+
+impl<T> ScheduledTask<T> {
+    fn new(scheduled_time: time::Instant, task: T, delay: time::Duration) -> Self {
+        ScheduledTask {
+            scheduled_time,
+            task,
+            delay,
+        }
+    }
+}
+
+impl<T: Ord> Ord for ScheduledTask<T> {
+    fn cmp(&self, other: &Self) -> Ordering {
+        other
+            .scheduled_time
+            .cmp(&self.scheduled_time)
+            .then_with(|| self.task.cmp(&other.task))
+            .then_with(|| self.delay.cmp(&other.delay))
+    }
+}
+
+impl<T: Ord> PartialOrd for ScheduledTask<T> {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+pub struct DoublingDelayScheduler<T> {
+    initial_delay: time::Duration,
+    scheduled_tasks: BinaryHeap<ScheduledTask<T>>,
+    on_new_task_tx: UnboundedSender<T>,
+    on_new_task_rx: UnboundedReceiver<T>,
+}
+
+impl<T: Ord> DoublingDelayScheduler<T> {
+    pub fn new(initial_delay: time::Duration) -> Self {
+        let (on_new_task_tx, on_new_task_rx) = unbounded();
+        DoublingDelayScheduler {
+            initial_delay,
+            scheduled_tasks: BinaryHeap::new(),
+            on_new_task_tx,
+            on_new_task_rx,
+        }
+    }
+}
+
+#[async_trait]
+impl<T: Ord + Send + Clone> TaskScheduler<T> for DoublingDelayScheduler<T> {
+    fn add_task(&mut self, task: T) {
+        self.on_new_task_tx
+            .unbounded_send(task)
+            .expect("We own the the rx, so this can't fail");
+    }
+
+    async fn next_task(&mut self) -> Option<T> {
+        let mut delay: futures::future::Fuse<_> = match self.scheduled_tasks.peek() {
+            Some(task) => tokio::time::delay_until(task.scheduled_time).fuse(),
+            None => futures::future::Fuse::terminated(),
+        };
+        // wait until either the scheduled time of the peeked task or a next call of add_task
+        futures::select! {
+            _ = delay => {},
+            task = self.on_new_task_rx.next() => {
+                if let Some(task) = task {
+                    let curr_time = time::Instant::now();
+                    let scheduled_task = ScheduledTask::new(curr_time, task, self.initial_delay);
+                    self.scheduled_tasks.push(scheduled_task);
+                } else {
+                    debug!(target: "rmc", "The tasks ended");
+                    return None;
+                }
+            }
+        }
+        let ScheduledTask {
+            scheduled_time,
+            task,
+            delay,
+        } = self.scheduled_tasks.pop().expect("Task must be ready");
+        let scheduled_task = ScheduledTask::new(scheduled_time + delay, task.clone(), 2 * delay);
+        self.scheduled_tasks.push(scheduled_task);
+        Some(task)
+    }
+}
+
+pub struct HashAlreadyExistsError {}
+
+pub struct ReliableMulticast<'a, H: Signable + Hash, MK: MultiKeychain> {
+    hash_states: HashMap<H, PartiallyMultisigned<'a, H, MK>>,
+    network_rx: UnboundedReceiver<Message<H, MK::Signature, MK::PartialMultisignature>>,
+    network_tx: UnboundedSender<(
+        NodeIndex,
+        Message<H, MK::Signature, MK::PartialMultisignature>,
+    )>,
+    keychain: &'a MK,
+    node_count: NodeCount,
+    scheduler: Box<dyn TaskScheduler<Task<H, MK>>>,
+    multisigned_hashes_tx: UnboundedSender<Multisigned<'a, H, MK>>,
+    multisigned_hashes_rx: UnboundedReceiver<Multisigned<'a, H, MK>>,
+}
+
+impl<'a, H: Signable + Hash + Eq + Clone + Debug, MK: MultiKeychain> ReliableMulticast<'a, H, MK> {
+    pub fn new(
+        network_rx: UnboundedReceiver<Message<H, MK::Signature, MK::PartialMultisignature>>,
+        network_tx: UnboundedSender<(
+            NodeIndex,
+            Message<H, MK::Signature, MK::PartialMultisignature>,
+        )>,
+        keychain: &'a MK,
+        node_count: NodeCount,
+        scheduler: impl TaskScheduler<Task<H, MK>> + 'static,
+    ) -> Self {
+        let (multisigned_hashes_tx, multisigned_hashes_rx) = unbounded();
+        ReliableMulticast {
+            hash_states: HashMap::new(),
+            network_rx,
+            network_tx,
+            keychain,
+            node_count,
+            scheduler: Box::new(scheduler),
+            multisigned_hashes_tx,
+            multisigned_hashes_rx,
+        }
+    }
+
+    pub fn start_rmc(&mut self, hash: H) {
+        let indexed_hash = Indexed::new(hash, self.keychain.index());
+        let signed_hash = Signed::sign(indexed_hash, self.keychain);
+        let message = Message::SignedHash(signed_hash.into_unchecked());
+        self.handle_message(message);
+    }
+
+    fn handle_message(&mut self, message: Message<H, MK::Signature, MK::PartialMultisignature>) {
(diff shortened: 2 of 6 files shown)


src/rmc.rs
```diff
@@ -0,0 +1,426 @@
...
+        keychain: &'a MK,
+        multisigned_hashes_tx: &UnboundedSender<Multisigned<'a, H, MK>>,
+    ) -> Option<Self> {
+        match maybe_state {
+            Some(HashState::SignatureAggregation {
+                signed,
+                mut partial,
+            }) => {
+                partial.add_signature(signed_indexed_hash);
+                match partial._try_into_complete(keychain) {
```
mike1729, line 89: This flow with returning sth on Err seems odd. Imho more natural would be to check `if keychain.is_complete(partial)` and then proceed with the Ok branch below.

src/rmc.rs
```diff
@@ -0,0 +1,426 @@
...
+    ) -> Option<Self> {
+        match maybe_state {
+            Some(HashState::Completed { .. }) => maybe_state,
+            Some(HashState::MultisignatureAggregation {
+                multisigned,
+                mut confirmed,
+                ..
+            }) => {
+                confirmed.set(signer_index);
+                if 3 * confirmed.true_indices().count() > 2 * node_count.0 {
```
mike1729, line 127: I propose to make the threshold here either configurable or define a method in lib.rs and use it here and in some other places where we calculate threshold.

src/rmc.rs
```diff
@@ -0,0 +1,426 @@
...
+                    Some(HashState::Completed {
+                        _multisigned: multisigned,
```
mike1729, line 129: Why this underscore?

src/rmc.rs
```diff
@@ -0,0 +1,426 @@
...
+        node_count: NodeCount,
+        keychain: &'a MK,
+        multisigned_hashes_tx: &UnboundedSender<Multisigned<'a, H, MK>>,
+    ) -> Option<Self> {
+        match message.message {
+            Message::SignedHash(unchecked) => {
+                let signed_indexed_hash = match unchecked.check(keychain) {
+                    Ok(signed_indexed_hash) => signed_indexed_hash,
+                    Err(_) => {
+                        debug!("Received a hash with a bad signature");
```
mike1729, line 165: Add a target to `debug!` so it may be filtered.

src/rmc.rs
```diff
@@ -0,0 +1,426 @@
...
+                        return maybe_state;
+                    }
+                };
+                let indexed_hash = signed_indexed_hash.as_signable();
+                let index = indexed_hash.index();
+                if index != message.sender {
+                    debug!("Index of a message does not match the sender");
+                    // This is an unusual situation, but we can still process the hash because
+                    // it is signed by the node with the index from the message.
```
mike1729, line 174: Is then `message.sender` necessary at all if we process anyway?

src/rmc.rs
```diff
@@ -0,0 +1,426 @@
...
+                Self::handle_signed_hash(
+                    maybe_state,
+                    signed_indexed_hash,
+                    node_count,
+                    keychain,
+                    multisigned_hashes_tx,
+                )
+            }
+            Message::MultisignedHash(unchecked) => {
+                let signed_indexed_unchecked = match unchecked.check(keychain) {
```
mike1729, line 185: What would be an attack scenario with sending a valid multisignature that has invalid signature? The multisignature is implicitly signed by all partial signers so I believe that the additional signature is unnecessary and we could define `MultisignedHash(Indexed<UncheckedSigned<H, M>, S>)`.

src/rmc.rs
```diff
@@ -0,0 +1,426 @@
...
+                    }
+                };
+                let indexed_unchecked = signed_indexed_unchecked.as_signable();
+                let index = indexed_unchecked.index();
+                if index != message.sender {
+                    debug!("Index of a message does not match the sender");
+                    // This is an unusual situation, but we can still process the hash because
+                    // it is signed by the node with the index from the message.
+                }
+                let unchecked_multisigned = indexed_unchecked.as_signable().clone();
```
mike1729, line 199: By using `MK::is_complete`, we could get rid of this clone.

mike1729 (changes requested):

timorl (changes requested):
Nice, although there was some miscommunication regarding the shape of the protocol after the last meeting – it seems I really should have updated the doc. Fortunately the changes are only simplifying the code and shouldn't be too bothersome.

Other than that some nitpicks and some minimal design improvement suggestions.

src/rmc.rs
```diff
@@ -0,0 +1,426 @@
...
+/// A message consists of either a signed (indexed) hash, or a multisigned hash together with a
+/// signature of the multisignature by the sender.
+#[derive(Clone)]
+pub enum Message<H: Signable, S: Signature, M: PartialMultisignature> {
+    SignedHash(UncheckedSigned<Indexed<H>, S>),
+    MultisignedHash(UncheckedSigned<Indexed<UncheckedSigned<H, M>>, S>),
+}
+
+impl<H: Signable, S: Signature, M: PartialMultisignature> Message<H, S, M> {
+    fn hash(&self) -> &H {
```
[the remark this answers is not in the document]
michalseweryn: Fixed

src/rmc.rs
```diff
@@ -0,0 +1,426 @@
...
+    pub recipient: NodeIndex,
+}
+
+/// Stores the current phase of a hash, and all relevant data useful in that phase:
+///
+/// * in the signature aggregation phase we store a signed hash and an aggregated partial
+/// multisignature under the hash
+/// * in the multisignature aggregation phase, we store the complete multisignature of the hash
+/// and the set of nodes about who we know that they have a multisignature
+/// * in the completed phase we only store the multisigned hash.
```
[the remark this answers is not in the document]
michalseweryn: Fixed.

src/rmc.rs
```diff
@@ -0,0 +1,426 @@
...
+        confirmed: BoolNodeMap,
+    },
+    Completed {
+        _multisigned: Multisigned<'a, H, MK>,
+    },
+}
+
+impl<'a, H: Signable + Clone, MK: MultiKeychain> HashState<'a, H, MK> {
+    fn new(signed: Signed<'a, Indexed<H>, MK>, keychain: &'a MK) -> Self {
+        let partial = signed.clone()._into_partially_multisigned(keychain);
```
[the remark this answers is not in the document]
michalseweryn: Yes.

src/rmc.rs
michalseweryn: I would rather avoid calling `keychain.is_complete(partial)` and then unwrapping the result. This can be made explicit in type by making `PartiallyMultisigned<'a, T, MK>` and `enum` with two variants: `Incomplete{ signable: T, signature: MK::PartialMultisignature}` and `Complete(Multisigned<'a, T, MK>)`. Then instead of calling a conversion function, we can simply pattern match the enum. Then in fact the `HashState` can be simply replaced with `PartiallyMultisigned<'a, T, MK>`.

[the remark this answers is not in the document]
michalseweryn: Fixed

src/rmc.rs
michalseweryn: Outdated.

src/rmc.rs
michalseweryn: At that moment we didn't use this field yet. I have added `pub fn get_multisigned(&self, hash: &H) -> Option<Multisigned<'a, H, MK>>`, so this will be used now.

src/rmc.rs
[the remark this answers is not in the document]
michalseweryn: Got rid of the `IncomingMessage` type

src/rmc.rs
michalseweryn: Replaced it with even simpler `MultisignedHash(UncheckedSigned<H, M>)`

src/rmc.rs
```diff
@@ -55,164 +42,58 @@ pub struct OutgoingMessage<H: Signable, MK: MultiKeychain> {
...
-                    Ok(signed_indexed_unchecked) => signed_indexed_unchecked,
-                    Err(_) => {
-                        debug!("Received a multisigned hash with a bad signature");
-                        return maybe_state;
+                },
+                Message::SignedHash(unchecked) => {
+                    let signed_hash = match unchecked.check(keychain) {
+                        Ok(signed_hash) => signed_hash,
+                        Err(_) => {
+                            debug!(target: "rmc", "Received a hash with a bad multisignature");
```
timorl, line 73: This and the other debug message seem switched.

src/rmc.rs
```diff
@@ -55,164 +42,58 @@ pub struct OutgoingMessage<H: Signable, MK: MultiKeychain> {
...
+                            return maybe_state;
+                        }
+                    };
+                    let partial = match maybe_state {
+                        None => signed_hash.into_partially_multisigned(keychain),
+                        Some(HashState::SignatureAggregation { mut partial }) => {
+                            partial.add_signature(signed_hash);
+                            partial
+                        }
+                        Some(HashState::Completed { .. }) => unreachable!(),
```
timorl, line 83: Ouch, now I see why you complained. How about abstracting this, together with the state change code below, into a function again, and explicitly handling this case there (by returning the unchanged state from the function ofc)?

src/rmc.rs
```diff
@@ -390,15 +263,19 @@ impl<'a, H: Signable + Hash + Eq + Clone + Debug, MK: MultiKeychain> ReliableMul
...
+
+    pub fn get_multisigned(&self, hash: &H) -> Option<Multisigned<'a, H, MK>> {
+        match self.hash_states.get(hash)? {
+            HashState::SignatureAggregation { .. } => None,
+            HashState::Completed { complete } => Some(complete.clone()),
         }
     }
 
-    pub async fn _next_multisigned_hash(&mut self) -> Option<Multisigned<'a, H, MK>> {
+    pub async fn next_multisigned_hash(&mut self) -> Option<Multisigned<'a, H, MK>> {
```
timorl, line 277: Uh, I just noticed nothing here is thread-safe. You have to be polling this to make RMC work, but at the same time you want to sometimes call at least `start_rmc` and perhaps `get_multisigned`. Both these require touching the map, which is not protected by a mutex or anything.

If you removed `get_multisigned` (as mentioned before I'm not completely sure it's necessary), then maybe the other case could be resolved by fiddling with the incoming messages channel (and using that in `start_rmc`), otherwise we need a mutex I believe. We might need it anyway, because without it `next_multisigned_hash` has to be `mut`?

timorl (changes requested):
I beleive there are some thread-safety problems, but I might be wrong.

src/rmc.rs
```diff
@@ -0,0 +1,302 @@
...
+                    };
+                    let partial = match maybe_state {
+                        None => signed_hash.into_partially_multisigned(keychain),
+                        Some(HashState::SignatureAggregation { mut partial }) => {
+                            partial.add_signature(signed_hash);
+                            partial
+                        }
+                        Some(HashState::Completed { .. }) => unreachable!(),
+                    };
+                    match partial.try_into_complete(keychain) {
```
mike1729, line 85: I like the idea of making `PartiallyMultisigned<'a, T, MK>` an enum and pattern matcing it here instead of calling the conversion function. Will you change it?

src/rmc.rs
```diff
@@ -0,0 +1,302 @@
...
+        }
+    }
+}
+
+pub enum Task<H: Signable, MK: MultiKeychain> {
+    BroadcastMessage(Message<H, MK::Signature, MK::PartialMultisignature>),
+}
+
+#[async_trait]
+pub trait TaskScheduler<T> {
```
timorl, line 105: After Damian's changes this should probably be based on a `DelaySchedule`, but I guess this can be left for the refactoring that unifies the usage of both throughout rush...

DamianStraszak: We talked about this a bit. I think it's good for now and we can try to unify it later. I'm thinking of TaskScheduler that supports DelaySchedules :), but again -- would leave it for later.

src/rmc.rs
michalseweryn: I refactored the code completely, and replaced the HashState type with PartiallyMultiSigned. The message handling logic is now simple enough that I decided to move it to a single method of ReliableMulticast. This allows to avoid passing a ton of arguments to a function and writing a spaghetti code

src/rmc.rs
michalseweryn: The problem is that also `multisigned_hashes_rx` would need to be protected by a mutex, and in the method `next_multisigned_hash` we would have to lock on it. I don't see yet how we can solve it...

src/rmc.rs
michalseweryn: Done

src/rmc.rs
michalseweryn: Correct

src/rmc.rs
michalseweryn: Fixed

timorl (changes requested):
I'm not quite sure why multisignatures can now be partially validated without being complete, and I suspect this might be incompatible with threshold signatures.

Other than that some minor comments.

src/rmc.rs
michalseweryn: But do we intend to use an RMC instance in more than one thread?

src/signed.rs
```diff
@@ -192,54 +247,86 @@ impl<T: Signable> Index for Indexed<T> {
...
+
+#[derive(Debug)]
+pub(crate) struct IncompleteMultisignatureError<'a, T: Signable, MK: MultiKeychain> {
+    pub(crate) partial: PartiallyMultisigned<'a, T, MK>,
+}
+
 #[derive(Debug)]
-pub struct IncompleteMultisignatureError<'a, T: Signable, MK: MultiKeychain> {
-    pub partial: PartiallyMultisigned<'a, T, MK>,
+pub enum PartiallyMultisigned<'a, T: Signable, MK: MultiKeychain> {
```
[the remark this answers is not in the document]
michalseweryn: Disagree. Typically "partial" does not exclude possibility of completeness (like in partial function or partial order)

src/rmc.rs
timorl: Good point, perhaps not.

src/signed.rs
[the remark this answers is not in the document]
timorl: Feels a bit strange to me nonetheless (I think because in your examples you usually use only the properties implied by "partialness", while here you explicitly have "complete" as one of the options), but I'm not gonna insist.

src/nodes.rs
```diff
@@ -7,7 +7,7 @@ use std::{
 };
 
 /// The index of a node
-#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, From)]
+#[derive(Copy, Clone, Debug, Default, Eq, PartialEq, Hash, From)]
```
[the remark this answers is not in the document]
michalseweryn: Removed

src/rmc.rs
```diff
@@ -0,0 +1,292 @@
...
+        multisigned_hashes_tx: &UnboundedSender<Multisigned<'a, H, MK>>,
+    ) -> Option<PartiallyMultisigned<'a, H, MK>> {
+        if let Some(PartiallyMultisigned::Complete { .. }) = maybe_state {
+            return maybe_state;
+        }
+        match message {
+            Message::MultisignedHash(unchecked) => match unchecked.check_multi(keychain) {
+                Ok(multisigned) => Some(PartiallyMultisigned::Complete { multisigned }),
+                Err(_) => {
+                    debug!(target: "rmc", "Received a hash with a bad ,multisignature");
```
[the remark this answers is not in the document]
michalseweryn: Removed

src/signed.rs
```diff
@@ -53,26 +55,40 @@ pub trait Signable {
 /// to upgrade this `struct` to `[Signed<'a, T, KB>]` which ensures that the signature matches the
 /// signed object, and the method `[UncheckedSigned::check_partial]` can be used to upgrade to
 /// `[PartiallyMultisigned<'a, T, MK>]`.
-#[derive(Clone, Debug, Decode, Encode)]
+#[derive(Clone, Debug, Default, Decode, Encode)]
```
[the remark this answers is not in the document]
michalseweryn: Removed

src/signed.rs
```diff
@@ -89,20 +105,44 @@ impl<T: Signable + Index, S: Clone> UncheckedSigned<T, S> {
...
-    pub fn check_partial<MK: MultiKeychain<PartialMultisignature = S>>(
+    /// Verifies whether the signature matches the key with the index as in the signed data.
+    pub(crate) fn check_multi<MK: MultiKeychain<PartialMultisignature = S>>(
         self,
         keychain: &MK,
-    ) -> Result<PartiallyMultisigned<T, MK>, SignatureError<T, S>> {
-        if !keychain.verify_partial(self.signable.hash().as_ref(), &self.signature) {
+    ) -> Result<Multisigned<T, MK>, SignatureError<T, S>> {
+        if !(keychain.verify_partial(self.signable.hash().as_ref(), &self.signature)
+            && keychain.is_complete(&self.signature))
```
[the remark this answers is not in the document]
michalseweryn: Merged `verify_partial` into `is_complete`

mike1729 (approved):

timorl (approved):

PR #71, opened 2021-05-26
cov scripts and small changes in signed

mike1729:
- scripts for code coverage generation
- more tests in signed
- some fixes in signed

### DIFF
--- a/src/signed.rs
+++ b/src/signed.rs
@@ -1,7 +1,7 @@
 use crate::{nodes::NodeIndex, Index};
 use codec::{Decode, Encode};
 use log::debug;
-use std::fmt::Debug;
+use std::{fmt::Debug, marker::PhantomData};
 
 /// The type used as a signature. The Signature typically does not contain the index of the node who
 /// signed the data.
@@ -76,7 +76,7 @@ impl<T: Signable + Index, S: Clone> UncheckedSigned<T, S> {
         }
         Ok(Signed {
             unchecked: self,
-            key_box,
+            marker: PhantomData,
         })
     }
 }
@@ -91,7 +91,7 @@ impl<T: Signable, S: Clone> UncheckedSigned<T, S> {
         }
         Ok(PartiallyMultisigned {
             unchecked: self,
-            keychain,
+            marker: PhantomData,
         })
     }
 }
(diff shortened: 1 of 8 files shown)


src/signed.rs
```diff
@@ -206,15 +201,15 @@ pub struct IncompleteMultisignatureError<'a, T: Signable, MK: MultiKeychain> {
 }
 
 impl<'a, T: Signable, MK: MultiKeychain> PartiallyMultisigned<'a, T, MK> {
-    pub fn sign(signable: T, keychain: &'a MK) -> Self {
+    pub fn sign(signable: Indexed<T>, keychain: &'a MK) -> Self {
```
timorl, line 204: This shouldn't require an `Indexed` argument, anything can be signed and the wrapper is not even used for anything.

timorl: You can, you should just pass the underlying type here, not the wrapped type you had to pass to create the `Signed` object.

I actually had a discussion about this with Michał and we ended up creating a low priority task to make explicit usage of `Indexed` completely unnecessary.

michalseweryn: `add_signature` takes `&mut PartiallyMultisigned<..., T, ...>` and `Signed<..., Indexed<T>, ...>`. Your version is a bit redundant, and you probably should anyway assert that `signable.index` is equal to `keychain.index()`

mike1729: Meh...you're absolutely right. I totally misunderstood some compiler error and followed it wrongly. I've reverted the change.

src/signed/tests.rs
```diff
@@ -117,28 +137,68 @@ impl MultiKeychain for TestMultiKeychain {
...
+fn test_incomplete_multisignature() {
+    let msg = indexed_test_message(0);
+    let index: NodeIndex = 0.into();
+    let node_count: NodeCount = 2.into();
+    let keychain = TestMultiKeychain { node_count, index };
+
+    let partial = PartiallyMultisigned::sign(msg, &keychain);
+    assert!(
+        partial._try_into_complete(&keychain).is_err(),
+        "5 signatures should form a complete signature"
```
michalseweryn, line 182: Probably a message "1 signature does not form a complete signature" is more descriptive?

mike1729: Definitely.

PR #73, opened 2021-05-27
add install script and option to run specific tests

mike1729:


install_cov_tools.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+cargo install rustfilt
+rustup component add llvm-tools-preview
+cargo install cargo-binutils
```
DamianStraszak, line 5: can also add `cargo install ripgrep` for `rg` to work, unless everyone is supposed to have it :)

PR #74, opened 2021-05-28
Add missing test to member to cover fetching missing coordinates

mike1729:


src/member.rs
```diff
@@ -839,3 +839,113 @@ where
         let _ = network_exit.send(());
     }
 }
+
+#[cfg(test)]
```
timorl, line 843: General comment/question, not directly related to the review: do we want to keep tests in the same files as the code? I see that this is nice in some ways (information is nicely concentrated in one place), but also any changes to tests mean the non-test parts of code also have to be recompiled (at least I think so, is that true?). This is not bothersome right now, but will probably be when the codebase and compilation times grow.

src/signed.rs
```diff
@@ -55,7 +55,7 @@ pub trait Signable {
 /// `[PartiallyMultisigned<'a, T, MK>]`.
 #[derive(Clone, Debug, Decode, Encode)]
 pub struct UncheckedSigned<T: Signable, S> {
-    signable: T,
+    pub(crate) signable: T,
```
timorl, line 58: I think Misza's changes add a getter for this, which is nicer than making the field (crate)public.

src/testing/mock.rs
```diff
@@ -201,7 +201,7 @@ impl Spawner {
     }
 }
 
-type NetworkData = NetworkDataT<Hasher64, Data, Signature>;
+pub(crate) type NetworkData = crate::NetworkData<Hasher64, Data, Signature>;
```
timorl, line 204: Oh, nice, `testing::NetworkData`, <3.

src/units.rs
```diff
@@ -9,7 +9,7 @@ use std::{cell::RefCell, collections::HashMap, hash::Hash as StdHash};
 
 #[derive(Debug, Clone, Copy, Eq, PartialEq, Encode, Decode, StdHash)]
 pub(crate) struct UnitCoord {
-    round: u16,
+    pub(crate) round: u16,
```
timorl, line 12: I would strongly prefer accessors to making fields public, especially only for tests. Same for the other instances in this file.

timorl (changes requested):
Lovely, the only issue I have is making some fields public – I'd prefer methods for that.

src/member.rs
mike1729: This file is already quite big and this test is more of an integration test so I'll move it.

src/units.rs
mike1729: Good point, I'll change that.

src/signed.rs
```diff
@@ -59,6 +59,13 @@ pub struct UncheckedSigned<T: Signable, S> {
     signature: S,
 }
 
+impl<T: Signable, S: Signature> UncheckedSigned<T, S> {
+    #[cfg(test)]
```
timorl, line 63: I think you can slap this on even one line higher.

timorl (approved):

src/signed.rs
mike1729: Nope, fmt doesn't allow that.

DamianStraszak (approved):

PR #78, opened 2021-06-01
Rush documentation

DamianStraszak:


rush.md
```diff
@@ -0,0 +1,374 @@
...
+
+### 1.2 High level idea of Rush requirements.
+
+Let us index the nodes taking part in the protocol as `i = 0, 1, ..., N-1` and call them "the committee", below we list some high-level requirements to be able to run Rush among this nodes:
+
+1. The nodes are connected via a network (that is not assumed to be 100% reliable) allowing them to send arbitrary messages to each other,
+2. Each node knows the identities of all other nodes (via their public keys) and holds a private key allowing it to sign messages.
+3. Each node `i` holds a "DataIO" object (to be explained in detail below) that allows it 1) to receive fresh pieces of data (we refer to it as the input stream `in_i`), and 2) to check that a piece of data received from another node is "available". The availability is best understood when thinking about the blockchain example and data being block hashes. Then the availability question for a blockhash is essentially whether we locally hold a block with such a hash.
+4. At most `N/3` of the nodes can be malicious (act with the intent to break the protocol guarantees).
```
michalseweryn, line 17: Shouldn't it be: "There are *less* that `N/3` malicious nodes?"

rush.md
```diff
@@ -0,0 +1,374 @@
...
+
+### 1.3 High level idea of Rush guarantees.
+
+Rush guarantees (as long as at most `N/3` nodes act maliciously) that each node `i` running the protocol produces a stream of data `out_i` such that:
+
+1. The output streams of any two nodes `out_i` and `out_j` are consistent, in the sense that at any given time one is a prefix of another. Moreover, eventually (possibly with some delays) all the `out` streams output the same data items (i.e., none of them gets "stuck").
```
michalseweryn, line 23: Does not the latter imply the former?

rush.md
```diff
@@ -0,0 +1,374 @@
...
+	// The order is allowed to depend on r and on Head(r-1, D)
+	for l = 1, 2, ..., s:
+		if Decide(U_l, D) == Some(b):
+			if b==true:
+				return U_l
+			else:
+				continue the loop
+		else:
+			return None
+```
```
michalseweryn, line 143: What if all `Decide(U_l, D)` return false

rush.md
```diff
@@ -0,0 +1,374 @@
...
+def Decide(U, D):
+	if there exists any unit V in D such that DecideVia(U, V, D) == Some(b):
+		// In this case it is guaranteed that all bs will be the same
+		return Some(b)
+	else:
+		return None
+```
+
+To explain `DecideVia` we first need to define "virtual voting" (we refer to the [paper](https://arxiv.org/abs/1908.05156) for intuitions and clarifications) -- intuitively, the units that are "beyond U" (i.e. have higher round number) "vote" for whether we should decide `true` or `false` for `U`. It is good to think that this vote decides, whether "`U` is well known among nodes".
```
michalseweryn, line 158: Maybe you could emphasize that at this point the creator of a unit does not have any way to impact the vote and it is completely determined by the shape of the dag?

rush.md
DamianStraszak: Yes, I hate the thresholds. Fixed throughout the document.

rush.md
DamianStraszak: Yeah, kind of, although it is hard to formulate safety and liveness in layman terms without some redundancy :) Reformulated.

rush.md
```diff
@@ -0,0 +1,374 @@
...
+4. At most `N/3` of the nodes can be malicious (act with the intent to break the protocol guarantees).
+
+### 1.3 High level idea of Rush guarantees.
+
+Rush guarantees (as long as at most `N/3` nodes act maliciously) that each node `i` running the protocol produces a stream of data `out_i` such that:
+
+1. The output streams of any two nodes `out_i` and `out_j` are consistent, in the sense that at any given time one is a prefix of another. Moreover, eventually (possibly with some delays) all the `out` streams output the same data items (i.e., none of them gets "stuck").
+2. Each item in the output stream is "available" to at least half the honest (= not malicious) nodes in the committee.
+3. Roughly speaking, most of the data items "proposed" by honest nodes eventually land in the output stream.
```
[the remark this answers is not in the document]
DamianStraszak: Added this note.

rush.md
```diff
@@ -0,0 +1,374 @@
...
+
+
+### 2.1. Dag -- the main idea of Aleph.
+
+Recall that in the model we are interested in, the nodes keep receiving new data items from their input streams `in_i` and would like to produce a single output stream `out` (i.e., the unique stream consistent with their local output streams `out_i`). The main idea of Aleph is for the nodes to produce a **DAG** (Directed Acyclic Graph) of vertices that we call **Units** where each such unit is created by one of the nodes `i` and carries one piece of data from the input stream `in_i`. Whenever a unit is created by a node `i` it is then dispersed among all other nodes with the intention that each node at any given time has a sufficiently up-to-date version of the Dag. Aleph then specifies a combinatorial algorithm that given a Dag `D` allows to compute (statically, without any need for communication) a certain prefix of the output stream `out`, with the property that as the Dag `D` is updated with new vertices to form `D'` then the new prefix computed out of `D'` is either the same or **extends** the one obtained from `D`. In other words, by growing the Dag, we cannot "roll back" the output sequence.
```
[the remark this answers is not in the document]
DamianStraszak: Generally, protocols are instructions for **honest nodes** -- we do not say what malicious nodes can/will do. Especially in such an informal sketch.

rush.md
```diff
@@ -0,0 +1,374 @@
...
+4. The `control_hash` of `U` or rather of parents of `U` is a fingerprint of the list of actual units being parents of `U`. While Aleph protocol instructs honest nodes to create exactly one unit per each round, this behavior cannot be enforced among malicious nodes. This is the reason why the `parent_map` array by itself does not necessarily uniquely specify the parents of `U` and we thus attach a `control_hash = Hash(U_1, U_2, ..., U_k)` where `U_1, ..., U_k` are the parents of `U`.
+5. The `signature` is a signature under the unit (as a collection of all the above fields) generated by `U.creator`. The signature is a proof that a given node indeed created such a unit, all units must be sent along with signatures.
+
+By `parents(U)` we denote the set of all units that `U` has edges to (these are always units of round one less than `U`'s round). By `Hash(U)` we mean the hash of a serialization of the fields of `U` not including the `signature`. Since `Hash(U)` is slightly smaller than `U` and still uniquely specifies `U` we sometimes use hashes in the protocol instead of units.
```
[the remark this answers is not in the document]
DamianStraszak: Done.

rush.md
```diff
@@ -0,0 +1,374 @@
...
+		if can_create(r):
+			Create a unit U of round r:
+				Pick as its parents at least floor(2*N/3)+1 units (from different creators) of round (r-1), including node k previous unit.
+				Poll the input stream for a fresh data item `d` and place it in `U` in the `data` field
+			Send U to all other nodes
+			sleep(CREATE_DELAY)
+			break the inner loop
+```
+
+The `creator_task` keeps creating units of growing rounds, with the requirement that, roughly, there at least  `(2/3)*N` units in the previous round available in `D`. There is a certain delay after each unit is created -- this technicality allows us to make sure the Dag does not grow too quickly, which is problematic for pragmatic reasons: the nodes have limited bandwidth (which also varies between different nodes) and have bounded RAM memory, where we would like to store the Dag.
```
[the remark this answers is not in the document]
DamianStraszak: Bounded RAM is the problem why we make the number of rounds bounded (which is still far away from a complete fix), but if we do not have delays then the sessions would possibly last too short and we would have no control over the length of sessions. However, the reader at this point is not aware of the max_round design hence I cannot and don't want to refer to that.

The reason behind sessions in explained in another part. I think what is written here is the best one can say without digging too deep into technicalities.

rush.md
```diff
@@ -0,0 +1,374 @@
...
+			new_batch = below(U)
+			remove from new_batch the contents of batch[0], batch[1], ..., batch[r-1]
+			sort new_batch using any canonical ordering
+			batch[r] = new_batch
+			let [U_1, U_2, ..., U_k] = new_batch
+			order = order ++ [U_1.data, U_2.data, ..., U_k.data]
+		else:
+			break
+	return order
+```
```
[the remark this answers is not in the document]
DamianStraszak: Generally I agree that Some() + None is not the standard pseudocode language, but in this case it allows to formulate things succinctly and clearly and I went for it anyway (there is also some explanation of this syntax above the pseudocode). 

Atfer your comment I started changing the code but rolled back quickly, because it would require to change too many things and for worse. One example: how would you formulate the "monotonicity" property without making it long or confusing using the pythonic syntax?

rush.md
DamianStraszak: Actually "continue" is when Decide() returns Some(false)

rush.md
DamianStraszak: Good point, added some comment on that.

rush.md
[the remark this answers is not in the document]
michalseweryn: I find `if Head(r, D) = Some(U):` confusing... I have never seen a notation where the pattern of a pattern matching is on the right hand side of `=`.

How about:
```
    let head = Head(r, D)
    if head == None:
        break
    let Some(U) = head
    ...
```
?

And I don't claim that all parts should use pythonic syntax. Natural language parts are awesome.

rush.md
michalseweryn: In the current version if all `Decide(U_l, D)` return `Some(false)`, we reach the end of the function and there is no `return` there, and we should return `Option<Unit>`.

Also, in the `if b==true` case, we should wrap `U_l` with `Some`

docs/src/what_is_rush.md
```diff
@@ -0,0 +1,24 @@
...
+
+### 1.2 High level idea of Rush requirements.
+
+Let us index the nodes taking part in the protocol as `i = 0, 1, ..., N-1` and call them "the committee", below we list some high-level requirements to be able to run Rush among this nodes:
+
+1. The nodes are connected via a network (that is not assumed to be 100% reliable) allowing them to send arbitrary messages to each other,
+2. Each node knows the identities of all other nodes (via their public keys) and holds a private key allowing it to sign messages.
+3. Each node `i` holds a "DataIO" object (to be explained in detail below) that allows it 1) to receive fresh pieces of data (we refer to it as the input stream `in_i`), and 2) to check that a piece of data received from another node is "available". The availability is best understood when thinking about the blockchain example and data being block hashes. Then the availability question for a blockhash is essentially whether we locally hold a block with such a hash.
```
michalseweryn, line 15: `DataIO` is also responsible for Output. Maybe do not mention the `DataIO` trait in this section yet?

rush.md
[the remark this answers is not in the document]
DamianStraszak: Right, I think this makes sense -- I changed it according to your suggestion. Now the pattern matching is hopefully more clear.

rush.md
DamianStraszak: Ok, there were indeed some mistakes there. I added a fix: it is required that the Dag is high enough for Head to execute (i.e., at least round r+3 reached) then, the code cannot exit the for loop without returning `None`. Added also a comment at the end of the function.

docs/src/what_is_aleph.md
```diff
@@ -0,0 +1,207 @@
...
+
+
+#### 2.2.4  Disseminating Units.
+From now on we assume that whenever a unit `U` lands in a Dag `D` of an honest node `k` then all other honest nodes will eventually (maybe after some delay) receive `U` and place it in their copies of the Dag. To achieve this in practice there are several mechanisms in Rush to guarantee robustness of the process of disseminating units. First of all the creator broadcasts the unit several times (with exponentially increasing delays). Secondly, there is a request-response mechanism that allows to fetch missing units from other nodes.
+
+### 2.3 Computing the Ordering from Dag.
```
michalseweryn, line 65: This section is extremely heavy and difficult, and the recursive definition of how the nodes are ordered is quite hard to follow (at least to me). Maybe before you give the pseudoce listings it would be a good thing to add a high-level description of what is going to happen in the algorithm:

1. For every unit `U` we have the rules for virtual voting, which means that each element of `below(U)` is assigned a value of type `Option<bool>`.
2. If among the parents of a node `V` (which is beyond `U`), there is a quorum of `threshold` nodes for who voted `Some(true)` for `U`, then `V` witnesses that `U` is a valid candidate for the `Head` of the current round.
3. For each round `r = 0, ...`, as soon as some unit `U` of round `r` becomes a valid candidate for the `Head`, add all units from `below(U)` to the order (except those which are already there).

I am pretty sure that in the description above i oversimplified something to the point where it is no longer a valid description of the algorithm, but thi is the impression I got from reading it, and if I got something wrong, then likely the other readers will as well.

docs/src/what_is_aleph.md
```diff
@@ -0,0 +1,207 @@
...
+
+**`ForkAlert(sender, forker, proof, units)`**
+
+where `sender` is the node index of the sender, `forker` is the node index of a forking node, `proof` is a pair of units created by forker that constitute a fork and `units` is simply some list of forker's units. Such a message is brodcast by `sender` to all other nodes:
+1. The broadcast is performed using a **Reliable Broadcast** primitive. Roughly speaking, this means that the node does not simply multicast this message to all the nodes but runs a specialized protocol that guarantees that all the nodes receive this message (it the sender is honest) and receive the exact same version of this message (even if the sender is **dishonest**). We give more details on the Rush's implementation of reliable broadcast in the dedicated section.
+2. The intuitive meaning of such an alert is that: "I, `sender`, am aware that `forker` forked and I attach a `proof` of that. Moreover, before I realized this fact, I accepted the following list of units to my local copy of the Dag. Please accept these units as well."
```
michalseweryn, line 198: Please, specify that all units in the list are created by the forker. It is said earlier that "`units` is simply some list of forker's units", but saying it one more time is a good redundancy which minimizes probablity of a misunderstanding

docs/src/what_is_aleph.md
```diff
@@ -0,0 +1,207 @@
...
+3. Each sender can send at most one alert regarding a specific `forker`. So in the worst case, sender would send roughly `f` alerts.
+4. Each node, right after it realizes about a new node forking, sends its own alert right away, attaching a list of units by forker that it added to its Dag.
```
michalseweryn, line 200: So if I got it right, each node upon receiving an alert, broadcast its own alert consisting of about* `r` units by forker which they have in their dag where `r` is the round from the `proof`, correct?

*There might be more than that if the `sender` is behind other nodes

docs/src/what_is_rush.md
```diff
@@ -0,0 +1,24 @@
...
+From the high level perspective Rush allows a set of `N` prespecified nodes to agree on an ordering of items arriving in some unreliable streams of data that these nodes locally observe. An illustrative example to have in mind here is when these nodes all observe a growing blockchain that does not have a built-in notion of finality, and would like to finalize blocks. Then the above mentioned "streams of data" are simply the advancing sequences of blockchain "tips" that each of the nodes sees. Note that in this (and in all the interesting examples), because of possible forks, or network delays, the data streams of individual nodes might not be consistent.
+The goal of a consensus protocol, is to ensure consistency of the decisions, even though relying on an unreliable data source. Consequently, what Rush produces is a single stream of data that "combines" all the individual streams of the `N` nodes and importantly **is consistent** among the nodes. Thus, in the example above, all the nodes would produce a unique sequence of **finalized** blocks (see also below for a more detailed description on how to use Rush as a finality gadget for a blockchain).
```
mike1729, line 7: (see also below for a more detailed description on how to use Rush as a finality gadget for a blockchain) it's not belowe but in other chapter so maybe it would be good to add a link to it? 
https://stackoverflow.com/questions/51187658/markdown-reference-to-section-from-another-file/51207259

docs/src/what_is_rush.md
mike1729, line 15: To be explained in later section.

docs/src/what_is_aleph.md
```diff
@@ -0,0 +1,207 @@
+## 2. How Rush Does it? What is Aleph?
+
+Rush implements  [Aleph Consensus Protocol](https://arxiv.org/abs/1908.05156) -- we highly encourage you to read the original paper, as it offers the best explanation of the ideas and does not require much background to understand. In this documentation we try to offer a self-contained sketch of the protocol, although, because of its compactness it might be harder to understand. For those who have read the paper we would like to note that there are certain minor differences between the paper and the actual implementation that we highlight below in a dedicated section. However, these do not affect any of the fundamental ideas behind Aleph or its security guarantees and are mostly tricks and implementation details.
```
mike1729, line 3: `we highlight below in a dedicated section.` - change below to later or generate a link

docs/src/what_is_aleph.md
```diff
@@ -0,0 +1,210 @@
...
+		if can_create(r):
+			Create a unit U of round r:
+				Pick as its parents at least N-f units (from different creators) of round (r-1), including node k previous unit.
+				Poll the input stream for a fresh data item `d` and place it in `U` in the `data` field
+			Send U to all other nodes
+			sleep(CREATE_DELAY)
+			break the inner loop
+```
+
+The `creator_task` keeps creating units of growing rounds, with the requirement that, roughly, there at least  `(2/3)*N` units in the previous round available in `D`. There is a certain delay after each unit is created -- this technicality allows us to make sure the Dag does not grow too quickly, which is problematic for pragmatic reasons: the nodes have limited bandwidth (which also varies between different nodes) and have bounded RAM memory, where we would like to store the Dag.
```
mike1729, line 55: there _are_

docs/src/what_is_rush.md
DamianStraszak: Yes, fixed.

docs/src/what_is_aleph.md
DamianStraszak: Added a discussion just before Section 2.3.1.

But in general I don't want to repeat everything from the paper, this is supposed to be a TLDR for people who understand code better than math :)

docs/src/what_is_aleph.md
DamianStraszak: The `r` should be the "current round" (there is no such a thing, but think of the "most recent round") and not the round from the fork (the adversary might uncover a very deep fork at any time). Whether the fork happened in round `0` or round `2500` we need to specify all the units (or at least some number of the most recent units, but that's an optimization which is a little unsafe :) ).

docs/src/what_is_rush.md
```diff
@@ -0,0 +1,24 @@
...
+
+### 1.2 High level idea of Rush requirements.
+
+Let us index the nodes taking part in the protocol as `i = 0, 1, ..., N-1` and call them "the committee", below we list some high-level requirements to be able to run Rush among this nodes:
+
+1. The nodes are connected via a network (that is not assumed to be 100% reliable) allowing them to send arbitrary messages to each other,
+2. Each node knows the identities of all other nodes (via their public keys) and holds a private key allowing it to sign messages.
+3. Each node `i` holds a data source object (to be explained in detail below -- see `DataIO`) that allows it 1) to receive fresh pieces of data (we refer to it as the input stream `in_i`), and 2) to check that a piece of data received from another node is "available". The availability is best understood when thinking about the blockchain example and data being block hashes. Then the availability question for a blockhash is essentially whether we locally hold a block with such a hash.
```
maciejnems, line 15: I guess this should link to `DataIO` section in `rush_api.md`. Simillar to what @mike1729 suggested.

maciejnems, line 15: > (to be explained in detail below -- see `DataIO`) 

I guess this should link to `DataIO` section in `rush_api.md`. Something simillar to what @mike1729  suggested.

docs/src/what_is_rush.md
DamianStraszak: I added references in all the places I spotted.

docs/src/what_is_rush.md
DamianStraszak: Fixed.

docs/src/what_is_aleph.md
DamianStraszak: Fixed.

docs/src/what_is_aleph.md
DamianStraszak: yes.

docs/src/what_is_aleph.md
DamianStraszak: Ok, I made this more specific.

docs/src/what_is_rush.md
DamianStraszak: Yes, corrected.

docs/src/what_is_aleph.md
```diff
@@ -0,0 +1,210 @@
...
+
+The main idea here is that we cannot afford to store an arbitrary amount of units that are forks -- most of them are useless, but without adding a specific mechanism to the protocol, there is no way to discard them. That's why we introduce **Fork Alerts**. A fork alert is a message
+
+**`ForkAlert(sender, forker, proof, units)`**
+
+where `sender` is the node index of the sender, `forker` is the node index of a forking node, `proof` is a pair of units created by forker that constitute a fork and `units` is simply some list of forker's units. Such a message is brodcast by `sender` to all other nodes:
+1. The broadcast is performed using a **Reliable Broadcast** primitive. Roughly speaking, this means that the node does not simply multicast this message to all the nodes but runs a specialized protocol that guarantees that all the nodes receive this message (it the sender is honest) and receive the exact same version of this message (even if the sender is **dishonest**). We give more details on the Rush's implementation of reliable broadcast in the dedicated section.
```
mike1729, line 200: (**if** the sender ...

docs/src/rush_api.md
```diff
@@ -0,0 +1,173 @@
...
+    fn sign(&self, msg: &[u8]) -> Self::Signature;
+    fn verify(&self, msg: &[u8], sgn: &Self::Signature, index: NodeIndex) -> bool;
+}
+```
+
+A typical implementation of KeyBox would be a collection of `N` public keys, an index `i` and a single private key corresponding to the public key number `i`. The meaning of `sign` is then to produce a signature using the given private key, and `verify(msg, s, j)` is to verify whether the signature `s` under the message `msg` is correct with respect to the public key of the `j`th node.
+
+### 3.2 Examples
+
+While the implementations of `KeyBox` and `Network` are pretty much universal, the implementation of `DataIO` depends on the specifica application. We consider two examples here.
```
mike1729, line 59: specific**a**

docs/src/internals.md
```diff
@@ -0,0 +1,47 @@
...
+
+1. The unit is created by one of the node's `Creator` component -- implemented in `src/creator.rs`. Creator sends a notification to an outer component.
+2. The newly created unit is filled with data, session information and a signature. This is done in `src/member.rs`. Subsequently a recurring task of broadcasting this unit is put in the task queue. The unit will be broadcast to all other nodes a few times (with some delays in between).
+3. The unit is received by another node -- happens in `src/member.rs` and passes some validation (signature checks etc.). If all these checks pass and the unit is not detected to be a fork, then it is placed in the `UnitStore` -- the `store` field of the `Member` struct.
+4. The idea is that this store keeps only **legit unit** in the sense defined in the section on alerts. Thus it no forks will be put there unless coming from alerts.
```
maciejnems, line 8: > Thus it no forks will be put there unless coming from alerts.

Maybe it should be:

```
Thus no forks would be put there unless coming from alerts.
```

docs/src/what_is_aleph.md
DamianStraszak: Fixed.

docs/src/rush_api.md
DamianStraszak: Fixed

docs/src/internals.md
DamianStraszak: Fixed.

docs/src/reliable_broadcast.md
```diff
@@ -0,0 +1,32 @@
+## 6 Reliable Broadcast
+
+Recall that Reliable Broadcast is the primitive we use to broadcast `fork alerts` among nodes -- see [the section on alerts](what_is_aleph.md#25-alerts----dealing-with-fork-spam). There are two requirements for a protocol realizing reliable broadcast:
+
+1. If an honest sender initiates a reliable broadcast with some message `m` then the protocol terminates and as a result all honest nodes receive `m`.
+2. If a malicious sender initiates a reliable broadcast then either it terminates for all honest nodes and they receive the same message `m`, or it does not terminate for all the honest nodes.
```
mike1729, line 6: all the honest nodes -> any honest node

docs/src/reliable_broadcast.md
```diff
@@ -0,0 +1,32 @@
...
+
+So, roughly speaking, we want a broadcast primitive that is consistent even if a malicious node is the sender. There is the possibility that a malicious broadcast will not terminate, but it is not hard to see that this is the best we can hope for.
+
+### 6.1 Consistent Broadcast using multisignatures -- RMC
+
+The main idea behind the reliable broadcast implementation in Rush is the use of multisignatures. Without going too much into details, think of a multisignature over a message `m` as a list of `N-f` signatures by `N-f` different committee nodes over the same message `m` (more efficient ways to achieve such a functionality are possible, like threshold signatures or signature aggregates, but they are beyond the scope of this section). Someone holding a multisignature over `m` can be sure that a large fraction of nodes "agree" (with the meaning of "agree" depending on the particular application) on `m`.
+
+The RMC (Reliable MultiCast) protocol is a way to reliably disseminate a single hash `h` among all nodes (in the next section we explain how to extend it to disseminating arbitrary data and only a hash). The idea is as follows:
```
mike1729, line 14: and **not** only a hash

docs/src/reliable_broadcast.md
```diff
@@ -0,0 +1,32 @@
...
+
+1. Whenever a node `i` wishes  to disseminate a hash `h` it initiates a reliable multicast instance by signing `h` and sending such a signed hash `SIG(h, i, sig_i)` to all other nodes.
+2. Upon receiving such a signed hash, each  node `j` signs `h` and sends its signed hash: `SIG(h, j, sig_j)` to all other nodes.
+3. Each node keeps receiving signatures under `h` from different nodes. Upon receiving `N-f` of them, this node combines the signatures into a single multisignature `msig` and sends to all nodes a message `MULTISIG(h, msig)`.
+4. Upon receiving `MULTISIG(h, msig)` under `h`, each node passes it also to all other nodes.
+
+The moment when a node receives `MULTISIG(h, msig)`  is considered as the completion of the multicast for this node (and even though the node still keeps resubmitting messages) this instance of RMC is considered as successful. If a RMC succeeds for some honest node then it is guaranteed to succeed for all honest nodes (but maybe with some delay). We refer to the file `/src/rmc.rs` for a thorough documentation of this component.
```
mike1729, line 21: multiple whitespaces

docs/src/reliable_broadcast.md
DamianStraszak: Fixed.

docs/src/reliable_broadcast.md
DamianStraszak: Fixed.

docs/src/reliable_broadcast.md
DamianStraszak: Fixed.

mike1729 (approved):

PR #79, opened 2021-06-01
Adjusted creator's behavior w.r.t. delays.

DamianStraszak:


src/creator.rs
```diff
@@ -111,24 +111,45 @@ impl<H: Hasher> Creator<H> {
...
-        loop {
-            tokio::select! {
-                Some(u) = self.parents_rx.next() => {
-                    self.add_unit(u.round(), u.creator(), u.hash());
-                    if self.check_ready() {
-                        self.create_unit();
-                        delay_for((self.create_lag)(round)).await;
-                        round += 1;
+        let mut delay_fut = delay_for(Duration::from_millis(0));
+        'outer: loop {
```
timorl, line 117: Maybe factor out the two inner loops into something like `delay_creation` and `create_asap` (hopefully with better names...), both returning whether they received an exit signal. This might make the structure easier to understand (with good names the comments would hopefully be unnecessary) and would remove the label, to which I have a personal aversion.

timorl (changes requested):

src/creator.rs
DamianStraszak: Got rid of the double loop, and the nested loop, using @mike1729 suggestion. Hopefully this looks better now :)

src/creator.rs
```diff
@@ -111,24 +111,34 @@ impl<H: Hasher> Creator<H> {
...
                 Some(u) = self.parents_rx.next() => {
                     self.add_unit(u.round(), u.creator(), u.hash());
-                    if self.check_ready() {
-                        self.create_unit();
-                        delay_for((self.create_lag)(round)).await;
-                        round += 1;
+                }
+                _ = &mut delay_fut => {
+                    if delay_passed {
+                        error!(target: "rush-creator", "{:?} more than one hour has passed since we created the previous unit.", self.node_ix);
```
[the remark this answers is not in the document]
timorl: I'm in Zürich so I'm not sure if I should accept "Krakowskim targiem", but sure... ;p

PR #81, opened 2021-06-02
Implement actual alerts using RMC

timorl:
Commit music: https://dubiozakolektiv.bandcamp.com/track/alarm-song

### DIFF
--- a/src/alerts.rs
+++ b/src/alerts.rs
@@ -54,6 +54,12 @@ impl<H: Hasher, D: Data, S: Signature> Alert<H, D, S> {
             }
         }
     }
+
+    // Simplified forker check, should only be called for alerts that have already been checked to
+    // contain valid proofs.
+    fn forker(&self) -> NodeIndex {
+        self.proof.0.as_signable().creator()
+    }
 }
 
 impl<H: Hasher, D: Data, S: Signature> Index for Alert<H, D, S> {
@@ -245,7 +251,7 @@ impl<'a, H: Hasher, D: Data, MK: MultiKeychain> Alerter<'a, H, D, MK> {
     }
 
     fn on_own_alert(&mut self, alert: Alert<H, D, MK::Signature>) {
-        let forker = alert.proof.0.as_signable().creator();
+        let forker = alert.forker();
         self.known_forkers.insert(forker, alert.proof.clone());
         let alert = Signed::sign(alert, self.keychain);
         if let Err(e) = self.messages_for_network.unbounded_send((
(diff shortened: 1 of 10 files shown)


src/alerts.rs
```diff
@@ -100,15 +177,15 @@ impl<H: Hasher, D: Data, KB: KeyBox> Alerter<H, D, KB> {
     fn units_are_legit(
```
DamianStraszak, line 177: The term "legit" has some specific meaning (because I said so long ago :) ): roughly speaking legit = "either from a non-forker or from alert". So it would be better to call this function differently. The field in the Alert field is fine though, as it is consistent with the above semantics.

DamianStraszak (changes requested):

src/testing/byzantine.rs
```diff
@@ -223,9 +221,6 @@ async fn honest_members_agree_on_batches_byzantine(
     let mut batch_rxs = Vec::new();
     let (net_hub, mut networks) = configure_network(n_members, network_reliability);
 
-    let alert_hook = AlertHook::new();
-    net_hub.add_hook(alert_hook.clone());
-
```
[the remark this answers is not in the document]
timorl: Added a check for whether every honest node sends at least one alert to every dishonest one. I expect that problems with the alerts failing to finish would already be visible as the consensus stopping.

src/alerts.rs
timorl: Changed to `correct_commitment`.

src/alerts.rs
```diff
@@ -26,51 +40,114 @@ impl<H: Hasher, D: Data, S> Alert<H, D, S> {
...
-    alerts_from_units: Receiver<Alert<H, D, KB::Signature>>,
+struct Alerter<'a, H: Hasher, D: Data, MK: MultiKeychain> {
+    keychain: &'a MK,
+    messages_for_network: Sender<(
+        AlertMessage<H, D, MK::Signature, MK::PartialMultisignature>,
+        Recipient,
+    )>,
+    messages_from_network: Receiver<AlertMessage<H, D, MK::Signature, MK::PartialMultisignature>>,
+    notifications_for_units: Sender<ForkingNotification<H, D, MK::Signature>>,
+    alerts_from_units: Receiver<Alert<H, D, MK::Signature>>,
```
mike1729, line 98: s/units/member here and above?

mike1729 (approved):

src/alerts.rs
```diff
@@ -159,47 +236,141 @@ impl<H: Hasher, D: Data, KB: KeyBox> Alerter<H, D, KB> {
...
+            Ok(alert) => alert,
+            Err(e) => {
+                debug!(target: "alerter","{:?} We have received an incorrectly signed alert: {:?}.", self.index(), e);
+                return;
+            }
+        };
+        let contents = alert.as_signable();
+        if let Some(forker) = self.who_is_forking(&contents.proof) {
+            if self.known_rmcs.contains_key(&(contents.sender, forker)) {
+                debug!(target: "alerter","{:?} We already know about an alert by {:?} about {:?}.", self.index(), alert.as_signable().sender, forker);
```
DamianStraszak, line 274: What will happen if we initially get some alert A for the pair (i, j) (and we were the only node to receive A) and all the other nodes received A' and actually completed RMC on A'?

timorl: O, nice catch, this is definitely a bug. Will require a bit of work to fix.

PR #82, opened 2021-06-02
Adds check availiability for data stored in units

mike1729:


src/lib.rs
```diff
@@ -28,8 +28,12 @@ mod testing;
 mod units;
 
 pub trait DataIO<Data> {
-    type Error: Debug;
+    type Error: Debug + 'static;
     fn get_data(&self) -> Data;
+    fn check_availability(
```
timorl, line 33: Please add a comment explaining the semantics of `Option` here – i.e. (if I understood correctly) that returning `None` means "don't worry about availability, everything is always fine".

timorl (changes requested):

src/lib.rs
mike1729: done

timorl (approved):

PR #86, opened 2021-06-04
Remove tokio from dependencies

kostekIV:


Cargo.toml
```diff
@@ -8,18 +8,16 @@ edition = "2018"
...
+log = "0.4"
+rand = "0.8"
+async-trait = "0.1"
 codec = { package = "parity-scale-codec", version = "2", default-features = false, features = ["derive"] }
-rand = "0.8.3"
-async-trait = "0"
 
 [dev-dependencies]
 libp2p = { version ="0", features = ["mdns", "tcp-tokio"]}
+tokio = { version = "0.2.21", features = ["macros", "rt-threaded", "time"] }
```
mike1729, line 21: Do we need macros and time now?

mike1729, line 21: And the most important thing, you can bump tokio version to the newest!

mike1729 (changes requested):

Cargo.toml
kostekIV: "time" was not needed but `macros` together with `rt` and `rt-multi-thread` are needed for tests because of `#[tokio::test]`.

kostekIV: Done

mike1729 (approved):

PR #89, opened 2021-06-07
Fixes ordering of closing channels and handling errors on sending on …

mike1729:
…closed channels

### DIFF
--- a/src/creator.rs
+++ b/src/creator.rs
@@ -77,13 +77,13 @@ impl<H: Hasher> Creator<H> {
 
         let new_preunit = PreUnit::new(self.node_ix, round, control_hash);
         debug!(target: "AlephBFT-creator", "{:?} Created a new unit {:?} at round {:?}.", self.node_ix, new_preunit, self.current_round);
-        if let Err(e) = self
+        self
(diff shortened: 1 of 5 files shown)


src/creator.rs
```diff
@@ -77,12 +77,18 @@ impl<H: Hasher> Creator<H> {
...
         debug!(target: "AlephBFT-creator", "{:?} Created a new unit {:?} at round {:?}.", self.node_ix, new_preunit, self.current_round);
-        self.new_units_tx
+        if let Err(e) = self
+            .new_units_tx
             .unbounded_send(NotificationOut::CreatedPreUnit(new_preunit))
-            .expect("Channel should be open");
+        {
+            debug!(target: "AlephBFT-creator", "{:?} notification channel is closed {:?}, closing", self.node_ix, e);
+            return Err(());
+        }
```
michalseweryn, line 86: For consistency, can we replace `if let` with  `.map_err(...)?` as in the extender

mike1729: Done.

michalseweryn (approved):

PR #90, opened 2021-06-07
Add support for asynchronous signing in KeyBox

michalseweryn:
Modify the `KeyBox` trait so that the `sign` method is `async`. This is because blocking signing in deprecated in substrate.

### DIFF
--- a/src/member.rs
+++ b/src/member.rs
@@ -170,12 +170,12 @@ where
             .expect("channel to consensus should be open")
     }
 
-    fn on_create(&mut self, u: PreUnit<H>) {
+    async fn on_create(&mut self, u: PreUnit<H>) {
         debug!(target: "AlephBFT-member", "{:?} On create notification.", self.index());
         let data = self.data_io.get_data();
         let full_unit = FullUnit::new(u, data, self.config.session_id);
         let hash = full_unit.hash();
-        let signed_unit = Signed::sign(full_unit, self.keybox);
+        let signed_unit = Signed::sign(full_unit, self.keybox).await;
         self.store.add_unit(signed_unit, false);
         let curr_time = time::Instant::now();
         let task = ScheduledTask::new(Task::UnitMulticast(hash, 0), curr_time);
@@ -316,10 +316,10 @@ where
         }
     }
 
-    fn on_consensus_notification(&mut self, notification: NotificationOut<H>) {
+    async fn on_consensus_notification(&mut self, notification: NotificationOut<H>) {
         match notification {
             NotificationOut::CreatedPreUnit(pu) => {
-                self.on_create(pu);
+                self.on_create(pu).await;
             }
             NotificationOut::MissingUnits(coords) => {
                 self.on_missing_coords(coords);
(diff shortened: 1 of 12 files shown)


src/member.rs
```diff
@@ -725,7 +725,7 @@ where
         loop {
             futures::select! {
                 notification = rx_consensus.next() => match notification {
-                        Some(notification) => self.on_consensus_notification(notification),
+                        Some(notification) => self.on_consensus_notification(notification).await,
```
michalseweryn, line 728: I am not sure if we want to `await` here, we could alternatively add a vector of futures outputting signed data and poll that vector.

src/signed.rs
```diff
@@ -17,10 +17,11 @@ impl<T: Debug + Clone + Encode + Decode + Send + 'static> Signature for T {}
...
 /// and `verify(msg, s, j)` is to verify whether the signature s under the message msg is
 /// correct with respect to the public key of the jth node.
+#[async_trait]
 pub trait KeyBox: Index + Clone + Send + Sync + 'static {
     type Signature: Signature;
     /// Signs a message `msg`.
-    fn sign(&self, msg: &[u8]) -> Self::Signature;
+    async fn sign(&self, msg: &[u8]) -> Self::Signature;
     /// Verifies whether a node with `index` correctly signed the message `msg`.
     fn verify(&self, msg: &[u8], sgn: &Self::Signature, index: NodeIndex) -> bool;
```
timorl, line 26: If `sign` is already `async` shouldn't `verify` also be? The access the same resources in very similar ways, so it's quite a bit strange that one is and one isn't.

src/signed.rs
```diff
@@ -90,7 +91,7 @@ pub struct UncheckedSigned<T: Signable, S> {
 /// to upgrade this `struct` to [`Signed<'a, T, KB>`] which ensures that the signature matches the
 /// signed object, and the method [`UncheckedSigned::check_multi`] can be used to upgrade to
 /// [`Multisigned<'a, T, MK>`].
-impl<T: Signable, S: Signature> UncheckedSigned<T, S> {
```
timorl, line 93: Why did you remove that bound? Is some linter complaining that it technically isn't used?

src/signed.rs
```diff
@@ -192,9 +193,9 @@ impl<'a, T: Signable + Clone + Index, KB: KeyBox> Clone for Signed<'a, T, KB> {
 
 impl<'a, T: Signable + Index, KB: KeyBox> Signed<'a, T, KB> {
     /// Create a signed object from a signable. The index of `signable` must match the index of the `key_box`.
-    pub fn sign(signable: T, key_box: &'a KB) -> Self {
+    pub async fn sign(signable: T, key_box: &'a KB) -> Signed<'a, T, KB> {
```
timorl, line 196: Huh, `Self` doesn't work with `async`?

timorl (changes requested):
Some nits and questions and a suggestion about also making `verify` async.

src/signed.rs
michalseweryn: In substrate, public keys are stored in-memory while the private keys can be stored in the filesystem (https://crates.parity.io/sc_keystore/struct.LocalKeystore.html), this is why only signing is `async`.

src/signed.rs
michalseweryn: I don't remember why did the compiler complain. In general there is a question whether the  bounds should be in the struct definition or in the `impl` blocks. It seems like the convention in Rust is to put them in the definition, so I restored the bound and propagated it to all `impl` s

src/signed.rs
michalseweryn: The compiler complains when `Self` has lifetimes

PR #91, opened 2021-06-08
Add simple multisignatures to AlephBFT

michalseweryn:
This PR adds a `DefaultMultiKeychain<KB: KeyBox>` struct, which is a simple implementation of `MultiKeyChain` where a multisignature is simply a list of signatures.

### DIFF
--- a/src/signed.rs
+++ b/src/signed.rs
@@ -1,6 +1,9 @@
-use crate::{nodes::NodeIndex, Index};
+use crate::{
+    nodes::{NodeCount, NodeIndex, NodeMap},
+    Index,
+};
 use async_trait::async_trait;
-use codec::{Decode, Encode};
+use codec::{Decode, Encode, Error, Input, Output};
 use log::debug;
 use std::{fmt::Debug, marker::PhantomData};
 /// The type used as a signature.
@@ -20,6 +23,8 @@ impl<T: Debug + Clone + Encode + Decode + Send + Sync + 'static> Signature for T
 #[async_trait]
 pub trait KeyBox: Index + Clone + Send + Sync + 'static {
     type Signature: Signature;
+
+    fn node_count(&self) -> NodeCount;
     /// Signs a message `msg`.
     async fn sign(&self, msg: &[u8]) -> Self::Signature;
     /// Verifies whether a node with `index` correctly signed the message `msg`.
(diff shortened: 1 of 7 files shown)


timorl (approved):

src/signed.rs
```diff
@@ -397,3 +400,89 @@ impl<'a, T: Signable, MK: MultiKeychain> PartiallyMultisigned<'a, T, MK> {
...
+}
+
+/// Keybox wrapper which implements MultiKeychain such that a partial multisignature is a list of
+/// signatures and a partial multisignature is considered complete if it contains more than 2N/3 signatures.
+///
+/// Note: this way of multisigning is very inefficient, and should be used only for testing.
+#[derive(Debug, Clone)]
+pub struct DefaultMultiKeychain<KB: KeyBox> {
+    key_box: KB,
+    node_count: NodeCount,
```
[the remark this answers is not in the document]
michalseweryn: Moved `node_count` to the `KeyBox` trait

src/signed.rs
```diff
@@ -398,3 +403,90 @@ impl<'a, T: Signable, MK: MultiKeychain> PartiallyMultisigned<'a, T, MK> {
...
+        index: NodeIndex,
+    ) -> Self::PartialMultisignature {
+        SignatureSet::add_signature(SignatureSet::new(self.node_count()), signature, index)
+    }
+
+    fn is_complete(&self, msg: &[u8], partial: &Self::PartialMultisignature) -> bool {
+        let signature_count = partial
+            .signatures
+            .iter()
+            .filter(|sgn| sgn.is_some())
```
mike1729, line 482: `Option` implements `iter`, so shorter but not necessarily more readable version is to just use https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flatten

src/signed.rs
```diff
@@ -398,3 +403,90 @@ impl<'a, T: Signable, MK: MultiKeychain> PartiallyMultisigned<'a, T, MK> {
...
+            .count();
+        if signature_count <= 2 * self.node_count().0 / 3 {
```
mike1729, line 484: It would be good to have this as a threshold parameter or calculated by a global function like `minimal_quorum` or sth like that.

mike1729 (changes requested):

src/nodes.rs
```diff
@@ -49,7 +49,7 @@ impl Div<usize> for NodeCount {
 }
 
 #[derive(Clone, Debug, Eq, PartialEq, From, Encode, Decode)]
-pub(crate) struct NodeMap<T>(Vec<T>);
+pub struct NodeMap<T>(Vec<T>);
```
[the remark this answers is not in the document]
michalseweryn: I didn't mean to make the `NodeMap` public, but deriving `Decode` causes a leak of that type. I added a custom implementation od `Encode + Decode` to avoid leaking the type.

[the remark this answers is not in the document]
michalseweryn: Also, on a side note: the type `T` in `NodeMap<T>` is currently always `Option<_>`. Probably we should refactor `NodeMap<T>` so that it has an interface more similar to say `HashMap<T>` (an element for an index can be absent), and also `BoolNodeMap` should be renames to `NodeSet`

src/signed.rs
michalseweryn: I think `flatten` is readable :).

src/signed.rs
michalseweryn: Added a `quorum` method

PR #92, opened 2021-06-08
Alerter tests

timorl:


src/testing/alerts.rs
```diff
@@ -0,0 +1,295 @@
...
+enum Output {
+    Outgoing(TestMessage, Recipient),
+    Notification(TestNotification),
+}
+
+struct Scenario {
+    n_members: NodeCount,
+    inputs: Vec<Input>,
+    expected: HashMap<Output, usize>,
+}
```
michalseweryn, line 39: The `Scenario` struct seems redundant. All it does is store inputs and outputs (`n_members` is implied by the `keychains` field of `TestCase`). Can we get rid of `Scenario` and replace `TestCase::run(self, run_as: NodeIndex)` with `TestCase::run_scenario(self, run_as: NodeIndex, inputs: Vec<Input>, mut outputs: HashSet<Output>)`? (the outputs are actually unique in each of your tests)

michalseweryn (changes requested):

src/alerts.rs
```diff
@@ -91,8 +102,22 @@ impl<H: Hasher, D: Data, S: Signature> Signable for Alert<H, D, S> {
     }
 }
 
+impl<H: Hasher, D: Data, S: Signature> PartialEq for Alert<H, D, S> {
+    fn eq(&self, other: &Self) -> bool {
+        self.key() == other.key()
+    }
+}
```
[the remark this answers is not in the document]
michalseweryn: Also, there is a `derivative` crate, which allows you to add an attribute `#[derivative(PartialEq="ignore")]` to ignore a field, we can consider using it.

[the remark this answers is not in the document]
timorl: Ah, I've been looking for something like `derivative` in this context, added and thanks!

I added the method not because it was more readable (which it wasn't), but to keep the implementations of `Hash` and `PartialEq` in sync. Thankfully this is much simpler with `derivative`, so the method is unnecessary. :>

src/testing/alerts.rs
```diff
@@ -0,0 +1,295 @@
...
+            .expected
+            .entry(Output::Notification(notification))
+            .or_insert(0) += 1;
+        self
+    }
+
+    fn check_output(&mut self, output: Output) {
+        match self.expected.get_mut(&output) {
+            Some(count) => *count -= 1,
+            None => debug!("Possibly unnecessary {:?} emitted by alerter.", output),
```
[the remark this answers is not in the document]
timorl: Yes, the second test does not check for all outgoing messages that will appear.

src/testing/alerts.rs
timorl: I think this is more readable than building up vectors, especially for bigger tests. To illustrate my point and improve the quality I shall add some shortly. <_<

I also planned to use some matchers in there (which would make the logic much less trivial and allow testing for unexpected output), but I couldn't find a good crate for that...

michalseweryn: I see your point... However the mechanics of a `TestCase` returning a mutable reference to a `Scenario` is quite confusing as well. Maybe instead we can completely get rid of the `TestCase` struct, and replace its methods with top-level functions? We do not need to keep a vector of keyboxes, we can generate them when they are needed? The lifetimes should not be a problem here as all `Signed`s are immediately downgraded to `UncheckedSigned`s.

src/alerts.rs
```diff
@@ -22,12 +23,15 @@ use std::{
 
 pub(crate) type ForkProof<H, D, S> = (UncheckedSignedUnit<H, D, S>, UncheckedSignedUnit<H, D, S>);
 
-#[derive(Debug, Decode, Encode)]
+#[derive(Debug, Decode, Encode, Derivative)]
+#[derivative(PartialEq, Eq, Hash)]
```
DamianStraszak, line 27: Why do we derive Hash here (and for several other types)? Or: where is it used?

timorl: `Output` is being held in `HashMap`s in the added tests.

src/testing/alerts.rs
timorl: Now some `Signed`s are actually combined to create a multisignature, so that is no longer an option. I smashed `TestCase` and `Scenario` together, although I had to add one more small struct for some more advanced tests.

src/testing/alerts.rs
```diff
@@ -0,0 +1,525 @@
...
+    let alert = test_case.alert(alerter_index, fork_proof.clone());
+    let alert_hash = Signable::hash(&alert);
+    let signed_alert = test_case
+        .unchecked_signed(alert.clone(), alerter_index)
+        .await;
+    test_case
+        .incoming_message(AlertMessage::ForkAlert(signed_alert))
+        .outgoing_notification(ForkingNotification::Forker(fork_proof))
+        .wait();
+    for i in 1..n_members.0 - 1 {
```
michalseweryn, line 338: Here we iterate from 1 to 5. Would iterating from 1 to 4 suffice? together with the own index 0 we would have > 2*N/3 honest nodes

timorl: True, but I'm trying not to make these tests too specific – this is "all honest nodes respond". Should I make another one that checks the minimum? It feels more like a problem for `signed` testing, but I'm not sure.

src/testing/alerts.rs
```diff
@@ -0,0 +1,525 @@
...
+    for i in 1..n_members.0 {
+        test_case.unexpected_message(
+            AlertMessage::RmcMessage(
+                own_index,
+                RmcMessage::SignedHash(signed_wrong_alert_hash.clone()),
+            ),
+            Recipient::Node(NodeIndex(i)),
+        );
+    }
+    // We also make a proper alert to actually have something to wait for.
```
michalseweryn, line 405: What do you mean?

src/testing/alerts.rs
```diff
@@ -0,0 +1,525 @@
...
+        .outgoing_message(
+            AlertMessage::RmcMessage(own_index, RmcMessage::SignedHash(signed_alert_hash.clone())),
+            Recipient::Node(querier),
+        )
+        .wait()
+        .incoming_message(AlertMessage::AlertRequest(querier, alert_hash))
+        .outgoing_message(
+            AlertMessage::ForkAlert(signed_alert.clone()),
+            Recipient::Node(querier),
+        );
```
michalseweryn, line 445: In the loop belowe we test this once again. Is that intentional?

src/testing/alerts.rs
timorl: `TestCase` only waits for messages while it has any messages it expects. If we don't expect any message it will just finish successfully immediately.

src/testing/alerts.rs
timorl: Yup, we want a response to every query, including repeated ones.

DamianStraszak (approved):

src/testing/alerts.rs
michalseweryn: This is fine in the current shape, leave it as is

michalseweryn (approved):

PR #93, opened 2021-06-14
fuzz tests

fixxxedpoint:
consult the `fuzzing` section of README.md


src/testing/mock.rs
```diff
@@ -1,14 +1,17 @@
 use async_trait::async_trait;
 use codec::{Decode, Encode};
+use log::{debug, error};
+use parking_lot::Mutex;
+
+use tokio::task::yield_now;
```
timorl, line 6: I think we are trying to avoid using `tokio`, so maybe use the equivalent from `futures`?

src/testing/mock.rs
```diff
@@ -243,35 +250,34 @@ struct Peer {
...
 
 impl UnreliableRouter {
     pub(crate) fn new(peer_list: Vec<NodeIndex>, reliability: f64) -> Self {
         UnreliableRouter {
-            peers: Arc::new(Mutex::new(HashMap::new())),
-            peer_list: Arc::new(Mutex::new(peer_list)),
-            hook_list: Arc::new(Mutex::new(Vec::new())),
+            peers: HashMap::new(),
+            peer_list,
+            hook_list: vec![],
```
timorl, line 265: Same comment about `Vec::new()`.

src/testing/fuzz.rs
```diff
@@ -0,0 +1,299 @@
...
+use futures_timer::Delay;
+use log::{error, info};
+use std::{
+    io::{BufRead, BufReader, Read, Result as IOResult, Write},
+    iter::once_with,
+    time::Duration,
+};
+use tokio::runtime::{Builder, Runtime};
+
+struct ClosureHook<F> {
```
timorl, line 24: This is a repeated layer of abstraction – `NetworkHook` essentially already is this. Since you are only using this for `spy` just implement `spy` as a `NetworkHook` directly.

src/testing/fuzz.rs
```diff
@@ -0,0 +1,299 @@
...
+            Ok(v) => Some(v),
+            Err(e) => {
+                error!("Error while decoding NetworkData: {:?}.", e);
+                None
+            }
+        }
+    }
+}
+
+struct RecorderNetwork<I: Iterator<Item = NetworkData>> {
```
timorl, line 85: The name is a bit confusing, maybe `PlaybackNetwork`?

src/testing/mock.rs
```diff
@@ -520,14 +521,15 @@ pub(crate) type HonestMember<'a> = Member<'a, Hasher64, Data, DataIO, KeyBox, Sp
 pub(crate) fn configure_network(
     n_members: usize,
     reliability: f64,
-) -> (UnreliableRouter, Vec<Option<Network>>) {
+    connected_peers: impl IntoIterator<Item = NodeIndex>,
```
timorl, line 524: Consider making a separate function `configure_network_with_connected_peers` or something like that, because this argument is very boilerplatey for almost all uses of this function. Of course `configure_network` would call the other function with the `1..n_members.map(NodeIndex)` argument.

src/testing/fuzz.rs
```diff
@@ -0,0 +1,299 @@
...
+    StreamExt,
+};
+use futures_timer::Delay;
+use log::{error, info};
+use std::{
+    io::{BufRead, BufReader, Read, Result as IOResult, Write},
+    iter::once_with,
+    time::Duration,
+};
+use tokio::runtime::{Builder, Runtime};
```
timorl, line 22: Still `tokio` stuff, but I totally don't understand why it's here – are you completely sure it cannot be removed?

src/testing/mod.rs
```diff
@@ -8,7 +8,9 @@ mod consensus;
 mod crash;
 #[cfg(test)]
 mod dag;
-#[cfg(test)]
+#[cfg(feature = "fuzz")]
+pub mod fuzz;
```
timorl, line 12: This whole setup is neither here nor there. Almost the whole implementation relevant to fuzzing is in the main crate, but the binaries themselves are separated. I would say either completely move it outside, just exporting some mock/test data types, so that it actually can be coded outside of the crate, or move the fuzz binaries into the crate. The advantage of the former is no `tokio` and other dependencies in alephbft (even if these were just conditional on "fuzz"), the advantage of the latter is no need to make any mock/test types `pub`. I feel like now we have about the worst of both worlds, but I'm not sure which is actually better.

timorl (changes requested):
Main issue is that the design seems to be a rotten compromise between having everything in the main crate vs having fuzzing binaries in a separate crate – see last comment for details.

Other than that mostly minor nits.

src/testing/fuzz.rs
fixxxedpoint: mock.rs still uses tokio::spawn and so I need to provide it some tokio::Runtime. I've checked futures::Executor::block_on and it was panicing. Or perhaps I can just remove those tokio things from mock.rs instead.

fuzz/libfuzz_targets/fuzz_target_1.rs
```diff
@@ -0,0 +1,43 @@
...
+    fn read(&mut self, buf: &mut [u8]) -> IOResult<usize> {
+        Ok(buf.iter_mut().zip(&mut self.0).fold(0, |count, (b, v)| {
+            *b = v;
+            count + 1
+        }))
+    }
+}
+
+#[derive(Debug)]
+struct VecOfStoredNetworkData(Vec<NetworkData>);
```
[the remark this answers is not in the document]
fixxxedpoint: hmm...FuzzNetworkData?

src/testing/mock.rs
fixxxedpoint: I can't find anything similar in `futures`. There is something in `futures-lite`. Previous version was providing my own implementation, but the way you need to implement it when using `futures` looked awful.

fuzz/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+edition = "2018"
+
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+aleph-bft = { path = "..", features = ["fuzz"] }
+libfuzzer-sys = { version = "0.4" }
+afl = { version = "0.8", optional = true }
+structopt = { version = "0.3.8", optional = true }
```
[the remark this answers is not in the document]
fixxxedpoint: clap 3 is still in beta state. What is our policy in such case?

fuzz/libfuzz_targets/fuzz_target_1.rs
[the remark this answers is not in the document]
timorl: I thought just `StoredNetworkData` would suffice, but that seems alright too.

fuzz/Cargo.toml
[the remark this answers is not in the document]
timorl: Ah, good point. I would say fuzz tests depending on something in beta are fine, but if anyone smarter wants to override then we can switch later.

src/testing/mock.rs
timorl: Right, there might be no equivalent in `futures`. :/ In this case I would still prefer eliminating `tokio` and keeping the previous version, even if it is a bit ugly. Not a strong preference though, so do what you will.

fuzz/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "rush-fuzz"
```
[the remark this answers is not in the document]
fixxxedpoint: changed

README.md
```diff
@@ -123,6 +123,46 @@ There are many unit tests and several integration tests that may be run by stand
 `cargo test --lib` or `cargo test --lib --skip medium` if you want to run just small tests.
 Alternatively, you may run the `run_local_pipeline.sh` script.
 
+### Fuzzing
+
+There are fuzzing tests that try to crash the whole application by creating arbitrary data for the network layer
+and feeding it into the `member` implementation. To run those tests you need to install `afl` and `cargo-fuzz`.
```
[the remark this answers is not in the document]
fixxxedpoint: extended

fuzz/afl_targets/fuzz_target_1.rs
```diff
@@ -0,0 +1,11 @@
+#[macro_use]
+extern crate afl;
```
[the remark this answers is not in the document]
fixxxedpoint: after version bump of afl I was able to use the new syntax for macro import

fuzz/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+[package.metadata]
+cargo-fuzz = true
+
+[dependencies]
+aleph-bft = { path = "..", features = ["fuzz"] }
+libfuzzer-sys = { version = "0.4" }
+afl = { version = "0.8", optional = true }
+codec = { package = "parity-scale-codec", version = "2", default-features = false, features = ["derive"] }
+log = "0.4"
+structopt = "0.3.8"
```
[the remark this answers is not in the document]
fixxxedpoint: I can use it, but it's still in beta state.

fuzz/libfuzz_targets/fuzz_target_1.rs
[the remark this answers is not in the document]
fixxxedpoint: changed

src/testing/fuzz.rs
fixxxedpoint: flattened

src/testing/fuzz.rs
fixxxedpoint: changed

src/testing/mock.rs
fixxxedpoint: Removed. Droping instances of network was enough to achive similar result.

src/testing/mod.rs
fixxxedpoint: It seems like it shouldn't be part of the same package due to the way it is compile, otherwise it is messing with workspace/cargo. I moved it to a separate package.

src/testing/mock.rs
```diff
@@ -205,15 +212,15 @@ impl Spawner {
...
+pub type NetworkData = crate::NetworkData<Hasher64, Data, Signature, PartialMultisignature>;
 type NetworkReceiver = UnboundedReceiver<(NetworkData, NodeIndex)>;
 type NetworkSender = UnboundedSender<(NetworkData, NodeIndex)>;
 
 pub(crate) struct Network {
     rx: NetworkReceiver,
     tx: NetworkSender,
     peers: Vec<NodeIndex>,
-    index: NodeIndex,
+    pub(crate) index: NodeIndex,
```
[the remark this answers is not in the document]
fixxxedpoint: changed

Cargo.toml
```diff
@@ -25,8 +25,15 @@ async-trait = "0.1"
 codec = { package = "parity-scale-codec", version = "2", default-features = false, features = ["derive"] }
 parking_lot = "0.11"
 
+# dependencies used by the fuzzing binary
+tokio = { version = "1.6.1", default-features = false, optional = true }
+env_logger = { version = "0.8", default-features = false, optional = true }
```
timorl, line 30: I hate that these are non-dev dependencies here, even if they are technically optional.

Very aggressive suggestion – we already duplicate a lot of the objects you want to use from `mock` in the dummy example. Maybe instead of changing the `testing/mock` part of the repo, create another subcrate `AlephBFT-mock` (better name welcome), which would implement the things fuzz and dummy example need (copy them from mock/dummy). It would still be duplicated (we still need these in `testing/mock` for, well, testing purposes and we cannot have circular dependencies I believe), but these getting out of sync are not really a problem.

This would fix all the problems with dependencies, as well as with the strange conditional pub statements, and no need for "fuzz" as a feature.

What do you think?

src/testing/mock.rs
```diff
@@ -280,120 +170,90 @@ impl UnreliableRouter {
...
-
-    pub(crate) fn count(&self, sender: NodeIndex, recipient: NodeIndex) -> usize {
-        match self
-            .alerts_sent_by_connection
-            .lock()
-            .get(&(sender, recipient))
-        {
-            Some(count) => *count,
-            None => 0,
+async fn yield_now() {
```
timorl, line 221: Uh, implementig our own `yield_now` is even worse than depending on tokio. I would really suggest leaving `UnreliableRouter` as-was here, with the `Future` implementation – it's unwieldy, but correct and does not need dependencies or implementations of functions that clearly should come from a library.

If you go with my suggestion of creating a separate mock subcrate, there you can simply use `tokio`, as you did previously here.

src/testing/mock.rs
```diff
@@ -280,120 +170,90 @@ impl UnreliableRouter {
...
-        if let crate::NetworkData(Alert(ForkAlert(_))) = data {
-            *self
-                .alerts_sent_by_connection
-                .lock()
-                .entry((sender, recipient))
-                .or_insert(0) += 1;
-        }
-    }
+pub trait NetworkHook: Send {
+    fn update_state(&mut self, data: &mut NetworkData, sender: NodeIndex, recipient: NodeIndex);
```
timorl, line 236: I'm a bit unsure about the change from mutexes to mut refs – what was your reasoning here? I suspect it's likely correct, so the question is mostly curiosity.

timorl (changes requested):
Another huge suggestion, sorry. <_<" Should at least be relatively quick to execute, using the Копи-Пэйст method.

README.md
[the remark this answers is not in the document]
fixxxedpoint: I don't think that it needs more clarification. They are two different tools recommended by https://rust-fuzz.github.io/book/. afl seems more mature, cargo-fuzz seems to be better integrated with rust. For example grandpa is using both. afl requires non-empty correct initial data set, cargo-fuzz can start with nothing (which probably might make it "slower" in achieving some "valid" state).

src/testing/mock.rs
fixxxedpoint: One obvious reason is that it is confusing when a method named update_state doesn't use a mutable reference. The other is that it forces you to use mutexes/other_ugliness even when Network owns some hook.

README.md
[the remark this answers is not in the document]
timorl: Ehh~, technically only one of them is recommended by the book, but in general I don't think using both is reasonable. Twice as much code to support and I'm not really sure what the gain is supposed to be. I would strongly recommend picking the one you think is better and removing the other.

src/testing/mock.rs
timorl: The first argument is not particularly convincing, with Rust mutability abstractions it's expected that non-mut things change state if they have some runtime ownership thingies in them (like mutexes). But yeah, I guess if we can avoid mutexes that's convincing.

timorl (approved):
<3

README.md
```diff
@@ -129,6 +129,49 @@ There are many unit tests and several integration tests that may be run by stand
...
+```sh
+cargo install cargo-fuzz
+cargo install afl
+```
+
+#### cargo-fuzz/libfuzzer
+
+```sh
+cd fuzz
+cargo fuzz run --features="libfuzz" fuzz_target_1
```
DamianStraszak, line 149: fuzz_target instead of fuzz_target_1?

DamianStraszak (approved):

README.md
fixxxedpoint: changed

PR #97, opened 2021-06-16
Added example code for measuring alephbft performance in the blockcha…

DamianStraszak:
…in setting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ The second example: `blockchain` is meant for benchmarking AlephBFT in the block
 ```
 ./examples/blockchain/run_blockchain 4
 ```
-where `4` in the above is the number of committee members and can be replaced by any reasonable number. Running this script will result in generating log files `node0.log, node1.log, ...` corresponding to subsequent nodes. As one of the last log messages the achieved transactions per second count should be displayed.
+where `4` in the above is the number of committee members and can be replaced by any reasonable number. Running this script will result in generating log files `node0.log, node1.log, ...` corresponding to subsequent nodes. The achieved transactions per second should be among the final log messages in these files.
 
 ### Tests
 
--- a/README.md
+++ b/README.md
@@ -106,9 +106,8 @@ rid of this dependency.
 
 ### Examples
 
-There is a basic implementation of an honest committee member that is not
-cryptographically secure and serves only as a working example of what has to be
-implemented and not how it should be implemented.
+Currently we provide two basic examples of running AlephBFT. The first one: `dummy-honest` implements a committee member that is not
+cryptographically secure and serves only as a working example of what traits need to be implemented and how to implement them.
 The example may be run using:
 
     cargo run --example dummy_honest my_id n_members n_finalized
@@ -116,6 +115,13 @@ The example may be run using:
     my_id -- our index, 0-based
     n_members -- size of the committee
     n_finalized -- number of data to be finalized
+
+The second example: `blockchain` is meant for benchmarking AlephBFT in the blockchain setting. It implements a simple round-robin blockchain assuming honest participation. The simplest way to run it is to use the provided script as follows (assuming we start in the root directory)
+
+```
+./examples/blockchain/run_blockchain 4
+```
+where `4` in the above is the number of committee members and can be replaced by any reasonable number. Running this script will result in generating log files `node0.log, node1.log, ...` corresponding to subsequent nodes. As one of the last log messages the achieved transactions per second count should be displayed.
 
 ### Tests
 
--- a/examples/blockchain/chain.rs
+++ b/examples/blockchain/chain.rs
@@ -122,7 +122,7 @@ impl DataIO {
     pub(crate) fn new() -> (Self, UnboundedReceiver<OrderedBatch<Data>>) {
         let (finalized_tx, finalized_rx) = mpsc::unbounded();
         // We initialize with the "Genesis Block" => number 0.
-        let available_blocks: HashSet<_> = vec![0u64].into_iter().collect();
+        let available_blocks: HashSet<_> = [0u64].iter().cloned().collect();
         let inner = InnerDataIO {
             current_block: 0,
             available_blocks,
@@ -154,6 +154,17 @@ impl DataIO {
     }
 }
 
+
+
+// Runs a process that maintains a simple blockchain. The blocks are created every config.blocktime_ms
+// milliseconds and the block authors are determined by config.authorship_plan. The default config
+// uses round robin authorship: node k creates blocks number n if n%n_members = k.
+// A node will create a block n only if:
+// 1) it received the previous block (n-1)
+// 2) it is the nth block author
+// 3) enough time has passed -- to maintain blocktime of roughly config.blocktime_ms milliseconds.
+// This process holds two channel endpoints: block_rx to receive blocks from the network and
+// block_tx to push created blocks to the network (to send them to all the remaining nodes).
 pub(crate) async fn run_blockchain(
     config: ChainConfig,
     data_io: DataIO,
@@ -179,8 +190,8 @@ pub(crate) async fn run_blockchain(
                     data_io.add_block(block_num);
                 }
             }
-            // We tick every 50ms.
-            let mut delay_fut = Delay::new(Duration::from_millis(50)).fuse();
+            // We tick every 10ms.
+            let mut delay_fut = Delay::new(Duration::from_millis(10)).fuse();
 
             futures::select! {
                 maybe_block = block_rx.next() => {
--- a/examples/blockchain/chain.rs
+++ b/examples/blockchain/chain.rs
@@ -154,8 +154,6 @@ impl DataIO {
     }
 }
 
-
-
 // Runs a process that maintains a simple blockchain. The blocks are created every config.blocktime_ms
 // milliseconds and the block authors are determined by config.authorship_plan. The default config
 // uses round robin authorship: node k creates blocks number n if n%n_members = k.
--- a/examples/blockchain/chain.rs
+++ b/examples/blockchain/chain.rs
@@ -1,0 +1,206 @@
+use aleph_bft::{NodeIndex, OrderedBatch};
+use codec::{Decode, Encode};
+use futures::{
+    channel::{
+        mpsc::{self, UnboundedReceiver, UnboundedSender},
+        oneshot,
+    },
+    Future, FutureExt, StreamExt,
+};
+use futures_timer::Delay;
+use log::{debug, error};
+use parking_lot::Mutex;
+use std::{
+    collections::HashSet,
+    pin::Pin,
+    sync::Arc,
+    time::{self, Duration},
+};
+
+type BlockNum = u64;
+
+#[derive(Clone, Encode, Decode)]
+pub(crate) struct Block {
+    pub(crate) num: BlockNum,
+    pub(crate) data: Vec<u8>,
+}
+
+impl Block {
+    pub(crate) fn new(num: BlockNum, size: usize) -> Self {
+        debug!(target: "Blockchain-chain", "Started block creation {:?}", num);
+        // Not extremely random, but good enough.
+        let data: Vec<u8> = (0..size)
+            .map(|i| ((i + i / 999 + (i >> 12)) % 8) as u8)
+            .collect();
+        debug!(target: "Blockchain-chain", "Finished block creation {:?}", num);
+        Block { num, data }
+    }
+}
+
+pub(crate) type BlockPlan = Arc<dyn Fn(BlockNum) -> NodeIndex + Sync + Send + 'static>;
+
+pub(crate) struct ChainConfig {
+    // Our NodeIndex.
+    pub(crate) node_ix: NodeIndex,
+    // Number of random bytes to include in the block.
+    pub(crate) data_size: usize,
+    // Delay between blocks
+    pub(crate) blocktime_ms: u64,
+    // Delay before the first block should be created
+    pub(crate) init_delay_ms: u64,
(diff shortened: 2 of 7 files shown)


examples/blockchain/network.rs
```diff
@@ -0,0 +1,377 @@
...
+                node_ix,
+            };
+            SwarmBuilder::new(transport, behaviour, local_peer_id)
+                .executor(Box::new(|fut| {
+                    tokio::spawn(fut);
+                }))
+                .build()
+        };
+
+        swarm.listen_on("/ip4/0.0.0.0/tcp/0".parse()?)?;
```
timorl, line 331: Oh, just IPv4? Is v6 not worth the effort, unlikely to be supported by the host or some other reason? This is not a problem, just curiosity.

examples/blockchain/network.rs
```diff
@@ -0,0 +1,377 @@
...
+                            debug!(target: "Blockchain-network", "Authenticated peer: {:?} {:?}", node_ix, peer_id);
+                            self.peer_by_index.insert(node_ix, peer_id);
+                        }
+                        Message::Block(block) => {
+                            debug!(target: "Blockchain-network", "Received block num {:?}", block.num);
+                            self.block_tx.unbounded_send(block).unwrap();
+                        }
+                    }
+                    // Not sure if better to "send" () or just do nothing?
+                    self.rq_rp.send_response(channel, ()).unwrap();
```
mike1729, line 229: I would do nothing.

examples/blockchain/network.rs
```diff
@@ -0,0 +1,377 @@
...
+}
+
+const ALEPH_PROTOCOL_NAME: &str = "/aleph/chain/1";
+
+type NetworkData = aleph_bft::NetworkData<Hasher256, Data, Signature, PartialMultisignature>;
+
+#[derive(Clone, Encode, Decode)]
+enum Message {
+    Auth(NodeIndex),
+    Consensus(Vec<u8>),
```
timorl, line 49: I'm pretty sure the `Vec<u8>` here could just be `NetworkData` with no additional effort. This saves one `encode` call every time this is used, but more importantly just seems cleaner – consensus messages are not arbitrary byte sequences after all.

Unless I missed some reason you actually need this to be a sequence of bytes?

examples/blockchain/network.rs
```diff
@@ -0,0 +1,377 @@
...
+                        }
+                        Message::Block(block) => {
+                            debug!(target: "Blockchain-network", "Received block num {:?}", block.num);
+                            self.block_tx.unbounded_send(block).unwrap();
+                        }
+                    }
+                    // Not sure if better to "send" () or just do nothing?
+                    self.rq_rp.send_response(channel, ()).unwrap();
+                }
+                RequestResponseMessage::Response {
```
mike1729, line 231: This may be shortened to `RequestResponseMessage::Response {..} => {`.

examples/blockchain/chain.rs
```diff
@@ -0,0 +1,206 @@
...
+            .unbounded_send(data)
+            .map_err(|_| ())
+    }
+}
+
+impl DataIO {
+    pub(crate) fn new() -> (Self, UnboundedReceiver<OrderedBatch<Data>>) {
+        let (finalized_tx, finalized_rx) = mpsc::unbounded();
+        // We initialize with the "Genesis Block" => number 0.
+        let available_blocks: HashSet<_> = vec![0u64].into_iter().collect();
```
timorl, line 125: No need for the `vec!`, just `iter` over the slice. I'm also quite surprized there is no macro for that.

examples/blockchain/network.rs
```diff
@@ -0,0 +1,377 @@
...
+        let (msg_to_manager_tx, msg_to_manager_rx) = mpsc::unbounded();
+        let (msg_from_manager_tx, msg_from_manager_rx) = mpsc::unbounded();
+        let (block_to_data_io_tx, block_to_data_io_rx) = mpsc::unbounded();
+        let (block_from_data_io_tx, block_from_data_io_rx) = mpsc::unbounded();
+        let mut swarm = {
+            let mut rr_cfg = RequestResponseConfig::default();
+            rr_cfg.set_connection_keep_alive(Duration::from_secs(10));
+            rr_cfg.set_request_timeout(Duration::from_secs(4));
+            let protocol_support = ProtocolSupport::Full;
+            // ProtocolSupport::Outbound
```
mike1729, line 300: Artifact?

examples/blockchain/main.rs
```diff
@@ -0,0 +1,124 @@
...
+use chrono::Local;
+use crypto::KeyBox;
+use network::{Network, Spawner};
+use std::{io::Write, time};
+
+mod chain;
+mod crypto;
+mod network;
+
+const DATA_SIZE: usize = 1000000;
```
timorl, line 15: If this is supposed to represent a specific number of "transactions" maybe write it out as such, `const TRANSACTION_SIZE...` etc.

mike1729 (changes requested):

examples/blockchain/main.rs
```diff
@@ -0,0 +1,124 @@
...
+        .init();
+
+    let my_id = parse_arg(1);
+    let n_members = parse_arg(2);
+    let n_finalized = parse_arg(3);
+    let my_node_ix = NodeIndex(my_id);
+    let start_time = time::Instant::now();
+    info!(target: "Blockchain-main", "Getting network up.");
+    let (network, mut manager, block_from_data_io_tx, block_from_network_rx) =
+        Network::new(my_node_ix).await.unwrap();
```
timorl, line 61: Maybe `expect` instead of `unwrap`?

timorl (changes requested):
Nice, just some minor comments.

examples/blockchain/chain.rs
```diff
@@ -0,0 +1,206 @@
...
+        while data_io
+            .available_blocks
+            .contains(&(data_io.current_block + 1))
+        {
+            data_io.current_block += 1;
+        }
+    }
+}
+
+pub(crate) async fn run_blockchain(
```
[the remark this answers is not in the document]
DamianStraszak: Yes, added some explanation.

examples/blockchain/network.rs
```diff
@@ -0,0 +1,377 @@
...
+    #[behaviour(ignore)]
+    consensus_tx: mpsc::UnboundedSender<Vec<u8>>,
+    #[behaviour(ignore)]
+    block_tx: mpsc::UnboundedSender<Block>,
+    #[behaviour(ignore)]
+    node_ix: NodeIndex,
+}
+
+impl Behaviour {
+    fn send_consensus_message(&mut self, message: Vec<u8>, recipient: Option<NodeIndex>) {
```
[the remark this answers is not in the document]
DamianStraszak: Done.

examples/blockchain/network.rs
```diff
@@ -0,0 +1,377 @@
...
+        _: &Self::Protocol,
+        mut io: &mut T,
+    ) -> io::Result<Self::Request>
+    where
+        T: AsyncRead + Unpin + Send,
+    {
+        let length = unsigned_varint::aio::read_usize(&mut io)
+            .await
+            .map_err(|err| io::Error::new(io::ErrorKind::InvalidInput, err))?;
+        if length > usize::try_from(self.max_request_size).unwrap_or(usize::max_value()) {
```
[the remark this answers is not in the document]
DamianStraszak: Yeah, I removed all these stupid limits :) They are there mainly for the byzantine scenario, which is irrelevant here.

examples/blockchain/network.rs
```diff
@@ -0,0 +1,377 @@
...
+    async fn write_request<T>(
+        &mut self,
+        _: &Self::Protocol,
+        io: &mut T,
+        req: Self::Request,
+    ) -> io::Result<()>
+    where
+        T: AsyncWrite + Unpin + Send,
+    {
+        {
```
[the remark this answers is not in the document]
DamianStraszak: Actually not -- must scold these guys who I copy-pasted this from :)

examples/blockchain/network.rs
DamianStraszak: Ok, doing nothing seems to work well, so we do nothing.

examples/blockchain/network.rs
DamianStraszak: Cool, done :)

examples/blockchain/network.rs
DamianStraszak: Yeah, removed.

examples/blockchain/network.rs
```diff
@@ -0,0 +1,377 @@
...
+#[derive(Clone)]
+pub(crate) struct Spawner;
+
+impl aleph_bft::SpawnHandle for Spawner {
+    fn spawn(&self, _: &str, task: impl Future<Output = ()> + Send + 'static) {
+        tokio::spawn(task);
+    }
+}
+
+const ALEPH_PROTOCOL_NAME: &str = "/aleph/chain/1";
```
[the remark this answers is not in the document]
DamianStraszak: Done.

examples/blockchain/network.rs
```diff
@@ -0,0 +1,377 @@
...
+                RequestResponseMessage::Request {
+                    request_id: _,
+                    request,
+                    channel,
+                } => {
+                    let message = Message::decode(&mut &request[..])
+                        .expect("honest network data should decode");
+                    match message {
+                        Message::Consensus(msg) => {
+                            self.consensus_tx.unbounded_send(msg).unwrap();
```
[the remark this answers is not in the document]
DamianStraszak: Done.

examples/blockchain/network.rs
DamianStraszak: This current version works on localhost only so I didn't spent much time thinking about this line (actually I spent no time on it as I copied it from Mike's dummy-honest.rs :) ). Making it work over WAN will be some fun -- we can consider IPv6 then.

examples/blockchain/network.rs
DamianStraszak: You are absolutely right. I fixed it. 

However, after fixing it, clippy is unhappy because the Consensus() enum variant is huge -- it says 384 bytes (why lol?) and there is this lint https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant -- so I put an ignore for this here. The ignore is fine, as anyway 99% of the messages are consensus messages.

examples/blockchain/chain.rs
DamianStraszak: Ok, I did that, but needed to add `cloned()` to make it work as `[1u64, 2u64, 3u64].iter()` iterates over &u64 and not u64 :(

examples/blockchain/main.rs
DamianStraszak: Good idea, done.

examples/blockchain/main.rs
DamianStraszak: Fixed.

examples/blockchain/network.rs
timorl: Ah, interesting. That is a reason to have it be `Vec<u8>` I did not know about although, as you say, it is not a strong reason in this case.

README.md
```diff
@@ -117,6 +116,13 @@ The example may be run using:
     n_members -- size of the committee
     n_finalized -- number of data to be finalized
 
+The second example: `blockchain` is meant for benchmarking AlephBFT in the blockchain setting. It implements a simple round-robin blockchain assuming honest participation. The simplest way to run it is to use the provided script as follows (assuming we start in the root directory)
+
+```
+./examples/blockchain/run_blockchain 4
+```
+where `4` in the above is the number of committee members and can be replaced by any reasonable number. Running this script will result in generating log files `node0.log, node1.log, ...` corresponding to subsequent nodes. As one of the last log messages the achieved transactions per second count should be displayed.
```
timorl, line 124: The last sentence has somewhat forced grammar, maybe "The achieved transactions per second should be among the final log messages in these files."?

examples/blockchain/chain.rs
timorl: I thought there would be a better way of doing this, but even the docs suggest doing it this way... Oh well, good enough.

examples/blockchain/network.rs
```diff
@@ -0,0 +1,364 @@
...
+
+#[allow(clippy::large_enum_variant)]
+#[derive(Clone, Encode, Decode)]
+enum Message {
+    Auth(NodeIndex),
+    Consensus(NetworkData),
+    Block(Block),
+}
+
+enum MessageRecipient {
```
timorl, line 54: Eh, it's a bit annoying – we already have this type inside AlephBFT (with slightly different names), and maybe just exposing it would be simpler. But it doesn't really make sense to expose it, because of the `Network` interface having `broadcast` as a separate method, rather than using `send(msg, Recipient)`. Maybe we should think about refactoring it into the latter shape in the future?

For context the aleph-node network implementation also uses this abstraction, although with the approach you had here before, with `None` meaning everyone.

timorl (approved):
<3

mike1729 (approved):
:mechanical_arm:

README.md
DamianStraszak: Yeah -- now better :)

examples/blockchain/network.rs
DamianStraszak: Yeah, a little annoying, although these are internals of some example we will probably run once or a few times and forget about it. Whether we should refactor the Network trait: I like both versions actually, the one you propose probably allows the trait implementations to be a few lines shorter :)

timorl: I will add a task for that, maybe one day~.

PR #102, opened 2021-06-25
Fix of TOB-ALEPH-004 (Incorrect state rollback when removing forker’s units)

kostekIV:


src/units/store.rs
```diff
@@ -169,3 +170,140 @@ impl<'a, H: Hasher, D: Data, KB: KeyBox> UnitStore<'a, H, D, KB> {
         self.max_round
     }
 }
+
+#[cfg(test)]
```
timorl, line 174: I think we prefer having tests in the `testing` directory. At least I do, makes compiling quicker...

timorl (changes requested):

src/units/store.rs
kostekIV: We have such tests in other files as well. I thought it makes more sense for unit tests of UnitStore struct to be in the UnitStore file while in `testing` directory tests usually are less unit like and more integration like. Regarding compilation times the `#[cfg(test)]` annotation prevents compilation unless something like `cargo test` is run - so commands like `cargo build` should not be affected - https://doc.rust-lang.org/book/ch11-03-test-organization.html#the-tests-module-and-cfgtest.

timorl: If you change something in a file, then the file and anything that depends on it has to be recompiled, even if you only changed the test part. :<

And I know about the other tests in files, that's why I'm complaining, it's from experience. D:

Technically this is not a blocker, although I would like to unify the way we do tests, and preferably in the separate-file manner for reasons outlined above. The tests in the `testing` directory are definitely not only integration-like, at least the alert tests are the most unit-testy I could come up with.

mike1729: I agree with @timorl that we should standardize the place for tests. Currently, the effort was to put unit tests in corresponding files and intergration tests in testing, so this PR fits into that effort. We could move all the tests to testing but let's do it in a separate PR.

src/units/store.rs
```diff
@@ -169,3 +170,140 @@ impl<'a, H: Hasher, D: Data, KB: KeyBox> UnitStore<'a, H, D, KB> {
...
+
+        // Round in progress still has forker's unit
+        assert_eq!(NodeCount(1), store.n_units_per_round[4]);
+        assert!(store
+            .by_coord
+            .contains_key(&UnitCoord::new(4, NodeIndex(0))));
+        assert!(store.by_hash.contains_key(&forker_hashes[4]));
+
+        // Rounds after round in progress are "free" of forker's units;
+        for (round, hash) in forker_hashes.iter().enumerate().take(7).skip(5) {
```
mike1729, line 302: `forker_hashes[5..7].iter().enumerate()`

mike1729 (changes requested):

src/units/store.rs
```diff
@@ -169,3 +170,140 @@ impl<'a, H: Hasher, D: Data, KB: KeyBox> UnitStore<'a, H, D, KB> {
...
+    }
+
+    #[tokio::test]
+    async fn add_units_increment_round_test() {
+        let n_nodes = NodeCount(10);
+        let threshold = NodeCount(4);
+
+        let mut store = UnitStore::<Hasher64, Data, KeyBox>::new(n_nodes, threshold, 100);
+
+        let keyboxes = vec![
```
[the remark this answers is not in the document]
kostekIV: Done but I had to collect it due to lifetimes requirements of a keybox in create_unit function.

src/units/store.rs
```diff
@@ -169,3 +170,140 @@ impl<'a, H: Hasher, D: Data, KB: KeyBox> UnitStore<'a, H, D, KB> {
...
+            KeyBox::new(n_nodes, NodeIndex(1)),
+            KeyBox::new(n_nodes, NodeIndex(2)),
+            KeyBox::new(n_nodes, NodeIndex(3)),
+            KeyBox::new(n_nodes, NodeIndex(4)),
+        ];
+
+        assert_eq!(0, store.round_in_progress);
+        assert_eq!(NodeCount(0), store.n_units_per_round[0]);
+
+        for (i, keybox) in keyboxes.iter().enumerate() {
```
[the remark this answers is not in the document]
kostekIV: Because of need to collect the keyboxes I had to use here iter.

src/units/store.rs
```diff
@@ -169,3 +170,140 @@ impl<'a, H: Hasher, D: Data, KB: KeyBox> UnitStore<'a, H, D, KB> {
...
+    }
+
+    #[tokio::test]
+    async fn mark_forker_restore_state() {
+        let n_nodes = NodeCount(10);
+        let threshold = NodeCount(4);
+
+        let mut store = UnitStore::<Hasher64, Data, KeyBox>::new(n_nodes, threshold, 100);
+
+        let keyboxes = vec![
```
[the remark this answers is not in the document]
kostekIV: done

src/units/store.rs
kostekIV: done

mike1729 (approved):

PR #103, opened 2021-06-25
Fix of TOB-ALEPH-008 (Blocking IO in Network trait)

kostekIV:


timorl (approved):

src/network.rs
```diff
@@ -32,8 +32,14 @@ use std::fmt::Debug;
 pub trait Network<H: Hasher, D: Data, S: Signature, MS: PartialMultisignature>: Send {
     type Error: Debug;
     /// Send a message to a single node.
+    ///
+    /// Note on the implementation: this function should be implemented in a non-blocking manner.
+    /// Otherwise, the performance might be affected negatively.
```
mike1729, line 37: I would say sth stronger like "Otherwise, the performance might be affected negatively or the execution may end up in a dead lock" or just "Otherwise, the execution of the consensus may end up in a dead lock".

src/network.rs
```diff
@@ -32,8 +32,14 @@ use std::fmt::Debug;
...
     fn send(&self, data: NetworkData<H, D, S, MS>, node: NodeIndex) -> Result<(), Self::Error>;
     /// Send a message to all nodes.
+    ///
+    /// Note on the implementation: this function should be implemented in a non-blocking manner.
+    /// Otherwise, the performance might be affected negatively.
```
mike1729, line 42: Same as above.

mike1729 (changes requested):

src/network.rs
kostekIV: done

mike1729 (approved):

PR #105, opened 2021-06-25
Standardize log levels

timorl:


conventions.md
```diff
@@ -0,0 +1,18 @@
+# Conventions
+
+A basic overview of (semi-)arbitrary conventions we follown in this repository.
```
DamianStraszak, line 3: follown -> follow

src/member.rs
```diff
@@ -368,37 +368,37 @@ where
         let su = match uu.check(self.keybox) {
             Ok(su) => su,
             Err(uu) => {
-                debug!(target: "AlephBFT-member", "{:?} Wrong signature received {:?}.", self.index(), &uu);
+                trace!(target: "AlephBFT-member", "{:?} Wrong signature received {:?}.", self.index(), &uu);
```
DamianStraszak, line 371: This is somewhat controversial, because many other incorrect-message errors are logged as warn.

src/member.rs
```diff
@@ -418,7 +418,7 @@ where
         if u_round <= round_in_progress + rounds_margin {
             self.store.add_unit(su, false);
         } else {
-            debug!(target: "AlephBFT-member", "{:?} Unit {:?} ignored because of too high round {} when round in progress is {}.", self.index(), full_unit, u_round, round_in_progress);
+            warn!(target: "AlephBFT-member", "{:?} Unit {:?} ignored because of too high round {} when round in progress is {}.", self.index(), full_unit, u_round, round_in_progress);
```
DamianStraszak, line 421: This is something that might happen when joining a session late, in which case we would see tons of warnings in logs. Maybe trace is better.

conventions.md
timorl: Fixed.

src/member.rs
timorl: Good catch, no idea what I've been thinking, fixed.

src/member.rs
timorl: I'm not sure, the round margin seems to be quite large usually, so I'd expect joining _this_ late to actually be a symptom of a problem. I think in this case I would like to see it in the logs even on low levels of logging.

I'm not very convinced I'm right though, so if you are unconvinced by the above arguments I'll just change it to `trace!`.

DamianStraszak: Ok, you are right, warn is fine. This should not happen during normal execution.

DamianStraszak (approved):

maciejnems (approved):

PR #106, opened 2021-06-28
TOB-ALEPH-005

mike1729:


src/terminal.rs
```diff
@@ -236,7 +236,15 @@ impl<H: Hasher> Terminal<H> {
...
             .clone();
-        self.post_insert.iter().for_each(|f| f(u.clone()));
+        if self
+            .post_insert
+            .iter()
+            .map(|f| f(u.clone()))
+            .any(|result| result.is_err())
+        {
+            debug!(target: "AlephBFT-terminal", "{:?} error in post_insert hook, closing", self.node_id);
+            return Err(());
```
timorl, line 246: This significantly changes the semantics of this method – are we sure skipping the operations below when a hook fails is really what we want to do? I would expect these hooks should be executed _after_ the calls below.

src/consensus.rs
```diff
@@ -43,15 +43,15 @@ pub(crate) async fn run<H: Hasher + 'static>(
 
     // send a new parent candidate to the creator
     terminal.register_post_insert_hook(Box::new(move |u| {
-        if let Err(e) = parents_tx.unbounded_send(u.into()) {
+        parents_tx.unbounded_send(u.into()).map_err(|e| {
```
timorl, line 46: I'm not sure what the current consensus is – are we going to use `expect` for all channels internal to AlephBFT or not? Or was it some other criterion?

timorl (approved):

src/terminal.rs
mike1729: All the hooks are crucial for the execution so if one fails we return error and close the party.

src/consensus.rs
mike1729: That would be preferred but here we cannot as potential panic would kill just the terminal task and all the rest components wouldn't know about that. The change tries to propagate this info to all the components.

timorl (approved):

fixxxedpoint (commented):
Just a suggestion: that dyn "at runtime" hook insertion looks unnecessary, why not just use a single hook doing both actions and use generic for it.

fixxxedpoint (approved):

PR #108, opened 2021-06-29
Fix of TOB-ALEPH-009 (Inconsistent closed channel error handling) 

maciejnems:
We decided to panic when the receiving end of channel is closed. It is done by calling `.expect(...)` on send result.

### DIFF
--- a/src/creator.rs
+++ b/src/creator.rs
@@ -7,7 +7,7 @@ use crate::{
 };
 use futures::{channel::oneshot, FutureExt, StreamExt};
 use futures_timer::Delay;
-use log::{error, info, trace, warn};
+use log::{info, trace, warn};
 use std::time::Duration;
 
 /// A process responsible for creating new units. It receives all the units added locally to the Dag
@@ -66,7 +66,7 @@ impl<H: Hasher> Creator<H> {
         }
     }
 
-    fn create_unit(&mut self) -> Result<(), ()> {
+    fn create_unit(&mut self) {
         let round = self.current_round;
         let parents = {
             if round == 0 {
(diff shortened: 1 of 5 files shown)


conventions.md
```diff
@@ -0,0 +1,15 @@
+# Conventions
+
+A basic overview of (semi-)arbitrary conventions we follown in this repository.
+
+## Closed channel error handling
```
timorl, line 5: This actually standardizes only one side of the channel, maybe since you are already at it you could do the same for `next()` returning `None`?

conventions.md
```diff
@@ -0,0 +1,15 @@
...
+
+We handle closed channel errors in two ways.
+
+### expect
+
+If we have a guarantee that `rx` is still open during send (for example we own the `rx`), we call `.expect(...)` on the `Result` of send.
+
+### map_err
```
timorl, line 13: You could write here a bit more about what the guarantees/expectations are – i.e. whether this is for channels that have the other end outside of AlephBFT or some others too and why we don't panic on these.

Shouldn't we actually? for many of them it seems that if they are closed then continuing to run the consensus might be wrong? Or perhaps the idea is that the rest of the node fails, but at least it supports the consensus? But the latter wouldn't work anyway, because of data availability, would it?

This definitely needs a tad more explanation, but also some discussion as to what our actual policy should be.

conventions.md
maciejnems: In current state channel error handling is quite straight forward, so this file is no longer needed

conventions.md
maciejnems: You are right, this map_err in some cases is obsolete, so I changed it to always panic on other side of channel being closed

timorl (approved):

mike1729 (approved):

PR #112, opened 2021-07-06
FIX: TaskHandle

fixxxedpoint:
TaskHandles were not awaited by main components causing panics on exit in member and consensus.

### DIFF
--- a/src/consensus.rs
+++ b/src/consensus.rs
@@ -70,17 +70,29 @@ pub(crate) async fn run<H: Hasher + 'static>(
         .fuse();
     info!(target: "AlephBFT", "{:?} All services started.", conf.node_ix);
 
+    let mut terminal_exited = false;
+    let mut creator_exited = false;
+    let mut extender_exited = false;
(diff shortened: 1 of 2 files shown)


timorl (approved):

src/consensus.rs
```diff
@@ -70,17 +70,29 @@ pub(crate) async fn run<H: Hasher + 'static>(
...
 
+    let mut terminal_exited = false;
+    let mut creator_exited = false;
+    let mut extender_exited = false;
     futures::select! {
         _ = exit => {},
-        _ = terminal_handle => {},
-        _ = creator_handle => {},
-        _ = extender_handle => {},
+        _ = terminal_handle => terminal_exited = true,
```
DamianStraszak, line 78: I think it makes sense to add debug messages here saying something like "Terminal task terminated early". Might help a lot sometime.

fixxxedpoint: Added.

timorl (approved):

PR #114, opened 2021-07-06
EXPERIMENTAL: idle-aware spawner

fixxxedpoint:
Implementation of SpawnHandle with addidtional method `wait_idle`  that attempts to check if any spawned task called `waker` while it was executed. Should only be used paired with a single-threaded tasks runtime (tokio's default for tests). This PR also removes delay-driven code in some of our tests.

### DIFF
--- a/src/testing/mock.rs
+++ b/src/testing/mock.rs
@@ -15,10 +15,12 @@ use std::{
     collections::{hash_map::DefaultHasher, HashMap},
     hash::Hasher as StdHasher,
     pin::Pin,
-    sync::Arc,
+    sync::{atomic::AtomicBool, Arc},
     task::{Context, Poll},
     time::Duration,
 };
+
+use tokio::task::yield_now;
 
 use crate::{
     alerts,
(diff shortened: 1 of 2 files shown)


src/testing/mock.rs
```diff
@@ -209,6 +226,91 @@ impl SpawnHandle for Spawner {
...
+        name: &'static str,
+        task: impl Future<Output = ()> + Send + 'static,
+    ) -> TaskHandle {
+        let wrapped = SpawnFuture::new(task, self.wake_flag.clone());
+        self.spawner.spawn_essential(name, wrapped)
+    }
+}
+
+impl IdleAwareSpawner {
+    pub async fn wait(&self) {
```
timorl, line 277: I don't think you ever use this?

src/testing/mock.rs
```diff
@@ -209,6 +226,91 @@ impl SpawnHandle for Spawner {
...
+    pub fn new() -> Self {
+        IdleAwareSpawner {
+            spawner: Arc::new(Spawner::new()),
+            idle_mx: Arc::new(Mutex::new(())),
+            wake_flag: Arc::new(AtomicBool::new(false)),
+        }
+    }
+
+    pub async fn wait_idle(&self) {
+        let _ = self.idle_mx.lock();
```
timorl, line 290: I don't see why this mutex is necessary. Also, if it is then I believe it doesn't work – if you assign it to `_` it immediately drops, you have to at least do `_x`.

timorl (changes requested):
Some nits.

PR #115, opened 2021-07-14
Minor fixes and compatibility changes for aleph-node

DamianStraszak:


src/rmc.rs
```diff
@@ -205,7 +199,7 @@ impl<'a, H: Signable + Hash + Eq + Clone + Debug, MK: MultiKeychain> ReliableMul
             network_rx,
             network_tx,
             keychain,
-            node_count,
+            _node_count,
```
michalseweryn, line 202: Can we remove this unused field from the `ReliableMulticast` struct? We cn keep it in the constructor to not break the api.

src/rmc.rs
```diff
@@ -214,10 +208,14 @@ impl<'a, H: Signable + Hash + Eq + Clone + Debug, MK: MultiKeychain> ReliableMul
...
     /// Initiate a new instance of RMC for `hash`.
     pub async fn start_rmc(&mut self, hash: H) {
+        debug!(target: "AlephBFT-rmc", "starting rmc for {:?}", hash);
         let signed_hash = Signed::sign_with_index(hash, self.keychain).await;
+
         let message = Message::SignedHash(signed_hash.into_unchecked());
         self.handle_message(message.clone());
-        self.scheduler.add_task(Task::BroadcastMessage(message))
+        let task = Task::BroadcastMessage(message);
+        self.do_task(task.clone());
```
michalseweryn, line 217: Doing this task immediately is not as crucial as in the case of completing multisigned signature: this is veeery unlikely that we do not do the task immediately after running start_rmc, but we can keep it as is "just in case"

src/rmc.rs
```diff
@@ -278,12 +276,15 @@ impl<'a, H: Signable + Hash + Eq + Clone + Debug, MK: MultiKeychain> ReliableMul
...
-        }
+        self.network_tx
+            .unbounded_send(message)
+            .expect("Sending message should succeed");
+        // for recipient in 0..self.node_count.0 {
+        //     let recipient = NodeIndex(recipient);
+        //     self.network_tx
+        //         .unbounded_send((recipient, message.clone()))
+        //         .expect("Sending message should succeed");
+        // }
```
michalseweryn: Let us remove this comment

michalseweryn (changes requested):
Two cosmetic changes

michalseweryn (approved):

src/rmc.rs
DamianStraszak: Done.

src/rmc.rs
DamianStraszak: Ok, let's keep it.

src/rmc.rs
DamianStraszak: Done

src/rmc.rs
DamianStraszak: Removed

Yithis (approved):

PR #116, opened 2021-07-16
A0-157: Make nodes catch up if they fall behind

Yithis:


src/creator.rs
```diff
@@ -27,7 +27,7 @@ pub(crate) struct Creator<H: Hasher> {
     parents_rx: Receiver<Unit<H>>,
     new_units_tx: Sender<NotificationOut<H>>,
     n_members: NodeCount,
-    current_round: Round, // current_round is the round number of our next unit
+    current_round: Round, // current_round is the highest round number of all known units
     candidates_by_round: Vec<NodeMap<Option<H::Hash>>>,
     n_candidates_by_round: Vec<NodeCount>,
     create_lag: DelaySchedule,
```
timorl, line 33: There is a lie in the documenting comment, that was a lie even before these changes but it's a good moment to fix it – we don't create units as soon as we can, but according to a delay schedule. Could you fix that?

src/creator.rs
```diff
@@ -116,14 +114,17 @@ impl<H: Hasher> Creator<H> {
...
-        let mut round: usize = 0;
+        let mut round: Round = 0;
         let mut delay_fut = Delay::new((self.create_lag)(0)).fuse();
         let mut delay_passed = false;
         loop {
             futures::select! {
                 unit = self.parents_rx.next() => {
                     if let Some(u) = unit{
                         self.add_unit(u.round(), u.creator(), u.hash());
+                        if self.current_round < u.round() {
```
timorl, line 125: I'm pretty sure this should happen in `init_round` rather than here, especially since I believe you might have introduced a bug where when we create a unit of a higher round than `current_round` we don't actually update `current_round`.

src/creator.rs
```diff
@@ -138,9 +139,13 @@ impl<H: Hasher> Creator<H> {
                     break;
                 },
             };
-            if delay_passed && self.check_ready() {
-                self.create_unit();
-                delay_fut = Delay::new((self.create_lag)(round + 1)).fuse();
+            while delay_passed && round < self.current_round {
```
timorl, line 142: I think `delay_passed` is not necessary here – we always want to be no more than one round behind.

timorl (changes requested):
Some tests would also be nice – I know there are none currently, but it's a good opportunity to add them, especially checking whether "catching up" actually works.

michalseweryn (changes requested):

src/creator.rs
```diff
@@ -116,14 +114,17 @@ impl<H: Hasher> Creator<H> {
 
     pub(crate) async fn create(&mut self, mut exit: oneshot::Receiver<()>) {
```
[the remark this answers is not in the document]
timorl: SGTM, but two caveats:
1. We probably shouldn't exit if half an hour passed – we don't do that now and I don't think there is reason to be that defeatist.
2. Where would the catching up happen? Would `collect_parents` return immediately if there are units above `round`?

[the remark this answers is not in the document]
michalseweryn: Yes, my bad:
Ad 1. makes sense, we may loop then.
Ad 2. `collect_parents` should be a future terminating when we are ready to create a unit for that round, whatever the condition is. Thus the name is not the best, something like `ready_for_create` would be a better name,

src/creator.rs
```diff
@@ -5,10 +5,9 @@ use crate::{
...
 };
-use futures::{channel::oneshot, FutureExt, StreamExt};
-use futures_timer::Delay;
+use futures::{channel::oneshot, StreamExt};
 use log::{info, trace, warn};
-use std::time::Duration;
+use tokio::time::{self, Duration};
 
 /// A process responsible for creating new units. It receives all the units added locally to the Dag
 /// via the parents_rx channel endpoint. It creates units according to an internal strategy respecting
```
michalseweryn, line 13: Can you add a comment that a unit from round 0 has no parents? and the below condition only apply when `r >= 1`?

src/creator.rs
```diff
@@ -95,60 +98,241 @@ impl<H: Hasher> Creator<H> {
...
     }
 
-    fn check_ready(&self, round: Round) -> bool {
-        if round == 0 {
-            return true;
-        }
-        if round > self.max_round {
-            warn!(target: "AlephBFT-creator", "{:?} Maximum round reached. Not creating another unit.", self.node_ix);
-            return false;
+    async fn ready_for_create(&mut self, round: Round) {
```
michalseweryn, line 101: I think the name `wait_until_ready` would suit better.

src/creator.rs
```diff
@@ -95,60 +98,241 @@ impl<H: Hasher> Creator<H> {
...
         // Additionally, our unit from previous round must be available.
-        let prev_round_index = (round - 1) as usize;
         let threshold = (self.n_members * 2) / 3;
 
-        self.n_candidates_by_round[prev_round_index] > threshold
-            && self.candidates_by_round[prev_round_index][self.node_ix].is_some()
+        while self.n_candidates_by_round[prev_round_index] <= threshold || self.candidates_by_round[prev_round_index][self.node_ix].is_none() {
+             if let Some(u) = self.parents_rx.next().await {
+                self.add_unit(u.round(), u.creator(), u.hash());
+             }
```
michalseweryn, line 135: please, add an `else` clause logging an error and returning

src/creator.rs
```diff
@@ -95,60 +98,241 @@ impl<H: Hasher> Creator<H> {
...
+                _ = &mut delay => {
+                    break;
+                }
+            }
         }
+
         // To create a new unit, we need to have at least >floor(2*N/3) parents available in previous round.
         // Additionally, our unit from previous round must be available.
-        let prev_round_index = (round - 1) as usize;
         let threshold = (self.n_members * 2) / 3;
```
michalseweryn, line 130: throughout the repository, `threshold` is defined as `floor(2*N/3) + 1` that is the required number of parents. Can we change it for consistency?

src/creator.rs
```diff
@@ -95,60 +98,241 @@ impl<H: Hasher> Creator<H> {
...
+                    }
+                    continue;
+                }
+                _ = &mut delay => {
+                    break;
+                }
+            }
         }
+
         // To create a new unit, we need to have at least >floor(2*N/3) parents available in previous round.
```
michalseweryn, line 128: Change to either `at least floor(2*N/3) + 1`, or `> floor(2*N/3)` (wiithout "at least")

src/creator.rs
michalseweryn: Also, can you add a comment that it is required that a unit can appear in the channel only after all its parents have appeared there?

src/creator.rs
```diff
@@ -54,8 +53,12 @@ impl<H: Hasher> Creator<H> {
         }
     }
 
-    // initializes the vectors corresponding to the given round (and all between if not there)
+    // updates current_round and initializes the vectors corresponding to the given round (and all between if not there)
     fn init_round(&mut self, round: Round) {
+        if self.current_round < round {
+            self.current_round = round;
```
michalseweryn, line 59: This is the only place, where we modify `current_round`, and the `current_round` is always equal to one less than the length of `candidates_by_round`. Can we replace the `current_round` with a method calculating it from `candidates_by_round`?

src/creator.rs
```diff
@@ -54,8 +53,12 @@ impl<H: Hasher> Creator<H> {
...
+        }
+
         while self.candidates_by_round.len() <= round.into() {
```
michalseweryn, line 62: The loop can replaced with `Vec::resize_with`

michalseweryn (changes requested):
Nice! it starts to look better :). Will continue the review tomorrow.

src/creator.rs
```diff
@@ -95,60 +98,241 @@ impl<H: Hasher> Creator<H> {
...
-        loop {
-            futures::select! {
-                unit = self.parents_rx.next() => {
-                    if let Some(u) = unit{
-                        self.add_unit(u.round(), u.creator(), u.hash());
-                        if self.current_round < u.round() {
-                            self.current_round = u.round();
-                        }
+        for round in 0..self.max_round {
+            let mut ticker = time::interval(Duration::from_secs(30 * 60));
```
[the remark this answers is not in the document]
Yithis: Done

src/creator.rs
```diff
@@ -95,60 +98,241 @@ impl<H: Hasher> Creator<H> {
...
+        let prev_round_index = match round.checked_sub(1) {
+            Some(prev_round) => prev_round as usize,
+            None => return,
+        };
+
+        let delay = time::sleep((self.create_lag)(round.into()));
+        tokio::pin!(delay);
+        loop {
+            if round < self.current_round {
+                return ();
```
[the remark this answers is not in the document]
Yithis: Done

src/creator.rs
Yithis: Done

src/creator.rs
Yithis: Done

src/creator.rs
Yithis: Done

src/creator.rs
Yithis: Done

src/creator.rs
Yithis: Done

src/creator.rs
Yithis: Done

src/creator.rs
```diff
@@ -97,53 +93,244 @@ impl<H: Hasher> Creator<H> {
...
+        let prev_round_index = match round.checked_sub(1) {
+            Some(prev_round) => prev_round as usize,
+            None => return,
+        };
+
+        let delay = time::sleep((self.create_lag)(round.into()));
+        tokio::pin!(delay);
+        loop {
+            if usize::from(round) < self.n_candidates_by_round.len() - 1 {
+                return; // Since we get unit from round r, we have enough units from previous rounds to create our unit
```
DamianStraszak, line 107: Two issues:
1) This allows for a very sneaky attack, where an adversarial node does not follow the schedule and creates unit right away -- all other nodes will follow him in this and as a consequence, they will reach MAX_ROUND early and crash. This can be fixed by requiring that the round is higher by one, i.e. the condition should be:
`usize::from(round) + 2 < self.n_candidates_by_round.len()`
(Note we cannot subtract 2 from the RHS because of unsigned variable).
2) The comment is inaccurate: the point is not that we can create a unit, the point is that we can skip the delay, because we are behind.

src/creator.rs
```diff
@@ -97,53 +93,244 @@ impl<H: Hasher> Creator<H> {
...
-            return true;
-        }
-        if self.current_round > self.max_round {
-            warn!(target: "AlephBFT-creator", "{:?} Maximum round reached. Not creating another unit.", self.node_ix);
-            return false;
+    async fn wait_until_ready(&mut self, round: Round) {
+
+        let prev_round_index = match round.checked_sub(1) {
+            Some(prev_round) => prev_round as usize,
+            None => return,
```
DamianStraszak, line 100: The delay `time::sleep((self.create_lag)(0))` should be applied *before* the 0th round, it's quite important actually -- this is the period when we listen for but not create units.

src/creator.rs
```diff
@@ -97,53 +93,244 @@ impl<H: Hasher> Creator<H> {
...
+        };
+
+        let delay = time::sleep((self.create_lag)(round.into()));
+        tokio::pin!(delay);
+        loop {
+            if usize::from(round) < self.n_candidates_by_round.len() - 1 {
+                return; // Since we get unit from round r, we have enough units from previous rounds to create our unit 
+            }
+
+            tokio::select! {
```
DamianStraszak, line 110: I believe, we used to avoid having tokio as a dependency and used the `futures` crate for select and stuff. Did we change this?

DamianStraszak: And actually, the code does not compile now because tokio is not added to dependencies.

DamianStraszak (changes requested):

src/creator.rs
timorl: Very nice catch with 1., I would add that we want to have a comment explaining this specific attack there, because the `2` in the condition will certainly look curious.

src/creator.rs
```diff
@@ -76,18 +73,17 @@ impl<H: Hasher> Creator<H> {
...
 
-        self.current_round += 1;
-        self.init_round(self.current_round);
+        self.init_round(round + 1);
     }
 
     fn add_unit(&mut self, round: Round, pid: NodeIndex, hash: H::Hash) {
         // units that are too old are of no interest to us
-        if round + 1 >= self.current_round {
+        if usize::from(round + 1) >= self.n_candidates_by_round.len() - 1 {
```
timorl, line 86: Elsewhere in the code we use `as usize` for this, perhaps change for consistency?

timorl: There are a couple of instances of this.

src/creator.rs
```diff
@@ -97,53 +93,244 @@ impl<H: Hasher> Creator<H> {
...
+                }
+            }
+            self.create_unit(round);
+        }
+        warn!(target: "AlephBFT-creator", "{:?} Maximum round reached. Not creating another unit.", self.node_ix);
+    }
+}
+
+
+#[cfg(test)]
```
timorl, line 161: If I read correctly there is no test here that actually would have failed before this change. Do you think you would be able to write such a test?

timorl (approved):
Minor comments, but LGTM other than issues pointed out by Damian. The tests ended up being more integration than unit, so it might be hard to add one that checks the change without lots of work, so no pressure about writing them if you prefer to merge this quicker.

src/creator.rs
Yithis: Done

src/creator.rs
Yithis: Done

src/creator.rs
Yithis: Done

src/creator.rs
Yithis: Done

src/creator.rs
Yithis: Done

michalseweryn (changes requested):

src/creator.rs
```diff
@@ -125,7 +131,8 @@ impl<H: Hasher> Creator<H> {
         let threshold = (self.n_members * 2) / 3 + NodeCount(1);
 
         while self.n_candidates_by_round[prev_round_index] < threshold || self.candidates_by_round[prev_round_index][self.node_ix].is_none() {
-             if let Some(u) = self.parents_rx.next().await {
+            println!("{:?} {:?}", self.node_ix, self.candidates_by_round[prev_round_index][self.node_ix].is_none());
```
timorl, line 134: Leftover print-debugging?

src/creator.rs
```diff
@@ -291,6 +299,8 @@ mod tests {
         finish(killers, handles).await;
     }
 
+    // This test checks if 5 creators that start at the same time and 2 creators that starts after those 5 first create 125 blocks,
```
timorl, line 302: s/blocks/units (it's correct _almost_ everywhere :D)

Maybe mention in this comment that this is a test for catching up, it's relatively obvious from the statement, but it doesn't hurt to be more explicit.

src/creator.rs
```diff
@@ -97,17 +99,21 @@ impl<H: Hasher> Creator<H> {
...
 
-        let delay = time::sleep((self.create_lag)(round.into()));
-        tokio::pin!(delay);
+        let mut delay = Delay::new((self.create_lag)(round.into())).fuse();
         loop {
-            if usize::from(round) < self.n_candidates_by_round.len() - 1 {
-                return; // Since we get unit from round r, we have enough units from previous rounds to create our unit 
+            if ((round + 2) as usize) < self.n_candidates_by_round.len() { // We need to require a number higher by one then currently highest round 
+                                                                           // (by 2 then length) to prevent attack when a malcious node is creating
+                                                                           // units without any delay to achive max_round as soon as possible
```
timorl, line 112: Comments-to-the-right is an interesting style, but I don't really like the effect. This might be a personal preference though, so uh, @michalseweryn, @DamianStraszak what do you think?

Usually I support uniform style decisions, but comments' uniformity doesn't seem to be crucial, so I would also be fine if we settled on "any comment style is fine".

timorl (changes requested):
Very minor things.

src/creator.rs
michalseweryn: Ugh, please uniform comment style... I like the effect, but since throughout the repos we use different style, I don't think we should incorporate it. 
I am fine with the default way, where a comment is above the commented entity.

DamianStraszak: Yeah, I would prefer comment above the if -- we do it consistently this way.

src/creator.rs
timorl: Oh goodness, it got worse. xd

src/creator.rs
```diff
@@ -97,53 +97,272 @@ impl<H: Hasher> Creator<H> {
...
+        );
+
+        let mut handles = vec![];
+        let mut killers = vec![];
+
+        for node_ix in 0..n_members {
+            let (units_out, from_test_controller) = mpsc::unbounded();
+
+            let mut creator = Creator::new(
+                default_config(
```
DamianStraszak, line 269: This is the default config for deployment, but not for tests (they quite slow because of this) -- take a look at testing::mock::gen_config.

timorl (approved):

src/creator.rs
Yithis: Done

src/creator.rs
```diff
@@ -97,53 +93,244 @@ impl<H: Hasher> Creator<H> {
...
+                                self.n_candidates_by_round.push(NodeCount(0));
+                            }
+                            if self.candidates_by_round[h.round() as usize][h.creator()].is_none() {
+                                self.candidates_by_round[h.round() as usize][h.creator()] = Some([0; 8]);
+                                self.n_candidates_by_round[h.round() as usize] += NodeCount(1);
+                            }
+                            let full_unit = FullUnit::<Hasher64, Data>::new(h.clone(), Data::new(UnitCoord::new(0, 0.into()), 0), 0);
+                            for c in self.units_out.iter() {
+                                if c.unbounded_send(full_unit.unit()).is_err() {
+                                    return;
```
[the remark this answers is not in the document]
Yithis: Done

src/creator.rs
```diff
@@ -97,53 +93,244 @@ impl<H: Hasher> Creator<H> {
...
+                n_candidates_by_round: vec![NodeCount(0)],
+                n_members,
+            }
+        }
+
+        async fn control(&mut self) {
+            while self.units < self.max_units {
+                if let Some(pre_unit) = self.notifications_in.next().await {
+                    match pre_unit {
+                        NotificationOut::CreatedPreUnit(h) => {
```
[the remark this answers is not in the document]
Yithis: Done

michalseweryn (approved):

src/creator.rs
```diff
@@ -98,53 +98,265 @@ impl<H: Hasher> Creator<H> {
...
+
+        let mut delay = Delay::new((self.create_lag)(round.into())).fuse();
+        loop {
+            // We need to require a number higher by one then currently highest round
+            // (by 2 then length) to prevent attack when a malcious node is creating
+            // units without any delay to achive max_round as soon as possible
+            if ((round + 2) as usize) < self.n_candidates_by_round.len() {
+                // Since we get unit from round r, we have enough units from previous
+                // rounds to skip delay, because we are already behind
+                return;
```
DamianStraszak, line 118: This is incorrect now. Our unit from previous round must also be available to return here.

src/creator.rs
```diff
@@ -98,53 +98,265 @@ impl<H: Hasher> Creator<H> {
...
+            None => {
+                Delay::new((self.create_lag)(round.into())).await;
+                return;
+            }
+        };
+
+        let mut delay = Delay::new((self.create_lag)(round.into())).fuse();
+        loop {
+            // We need to require a number higher by one then currently highest round
+            // (by 2 then length) to prevent attack when a malcious node is creating
```
DamianStraszak, line 113: typo `malcious`

src/creator.rs
DamianStraszak: To clarify: the fact that we have created a unit is not the same thing as having it available as a parent in the Creator -- the created unit must first go through several components to "come back" to the Creator.

src/creator.rs
Yithis: Done

src/creator.rs
Yithis: Done

src/creator.rs
```diff
@@ -77,74 +76,282 @@ impl<H: Hasher> Creator<H> {
...
-        }
-        if self.current_round > self.max_round {
-            warn!(target: "AlephBFT-creator", "{:?} Maximum round reached. Not creating another unit.", self.node_ix);
-            return false;
+    async fn wait_until_ready(&mut self, round: Round) {
+        let prev_round_index = match round.checked_sub(1) {
+            Some(prev_round) => prev_round as usize,
+            None => {
+                Delay::new((self.create_lag)(round.into())).await;
+                return;
```
DamianStraszak, line 101: I don't like this `return` here. This does not allow to skip this delay. Instead one can move this piece of code to after the `loop` -- to line 128. Then you just remove the `Delay::new((self.create_lag)(round.into())).await;` and it should work fine.

Yithis: Done

PR #120, opened 2021-07-21
Add automatic version bumping

timorl:
I'm not actually sure this will work, as I'm not very familiar with GitHub actions. I also don't even know how to check it without merging. <_<"


.github/workflows/version-bump.yml
```diff
@@ -0,0 +1,15 @@
+name: Autobump version
+
+on:
+  push:
+    branches:
+    - master
```
fbielejec, line 6: it's now called main

.github/workflows/version-bump.yml
```diff
@@ -0,0 +1,15 @@
...
+
+jobs:
+  autobump:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: bump
+        run: ./bump_version.sh && git push -f origin master || echo 'No action needed.'
```
mike1729, line 15: Should be origin main

mike1729 (changes requested):

bump_version.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
```
fbielejec, line 1: maybe put the script in .github/workflows or just write it plainly in the yaml file?
```
        run: |
          echo a
          echo b         
```

.github/workflows/version-bump.yml
fbielejec, line 15: there seems to be a packaged Action for commits and pushes:
https://github.com/marketplace/actions/add-commit

maybe we could just use that? Simply define it as a step after bumping the version in the Cargo file?

.github/workflows/version-bump.yml
timorl: Oops, fixed.

.github/workflows/version-bump.yml
timorl: Oops, fixed.

timorl: I have seen that, but I didn't see a way of doing amends there – and creating an additional commit after every merge seems like a mess.

fbielejec: O would also avoid force pushing, just in case...

bump_version.sh
timorl: Added straight to the yaml file, seems somewhat strange, but I guess it should work.

.github/workflows/version-bump.yml
mike1729: If we amend then we have to force push. I don't know how to do a separate commit from this action.

timorl: This wouldn't work with `--amend`.

fbielejec (approved):
if it works +1 here

mike1729 (approved):

PR #130, opened 2021-07-26
Fixed bug with wrong control hash of self unit.

DamianStraszak:
The issue was that in a certain very unlucky situations involving forks it could happen that an honest unit creator would not be able to answer a parents request for its own unit (it wouldn't be able to decode the control hash). 

Very roughly speaking the reason for that was that in the terminal we only remember one unit variant (=fork) in the `unit_by_coord` map (Terminal), and more specifically (prior to this fix) this was always the most recent variant we received. This map is used when trying to decode parents -- if the parents units from this map do not agree with the control hash, a `WrongControlHash` notification is sent. It could happen, that a unit U was created, referring to a parent V1, and then before this unit U was added to the DAG, another variant V2 could be added to DAG (after a successful alert), replacing V1 in the `unit_by_coord` map, which would then result in a `WrongControlHash` notification for U and consequently in a serious issue -- none of the members would be able to decode the parents. The fix for this is quite easy: we just add a field `parent_hashes` to the CreatedPreUnit notification that is sent out by the Creator. This way, we can note in the store the parents of our units, so that even if it happens that Terminal reports WrongControlHash for our unit, we will have the parents in our store and will successfully decode them.

This bug fix does not have any accompanying test -- after adding another field to the notification this bug cannot be recreated, because by construction the parents are explicitly provided, so they cannot be missing anymore :) Still, I have locally run the `small_byzantine_one_forker` test 1000 times, with 0 failures, whereas previously it froze roughly 1 in 50 times.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,6 @@ codec = { package = "parity-scale-codec", version = "2", default-features = fals
 parking_lot = "0.11"
 
 [dev-dependencies]
-assert_matches = "1.5"
 sha3 = "0.9.1"
 unsigned-varint = { version = "0.7.0", features = ["futures", "asynchronous_codec"] }
 tokio = { version = "1.6.1", features = ["macros", "rt", "rt-multi-thread"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ codec = { package = "parity-scale-codec", version = "2", default-features = fals
 parking_lot = "0.11"
 
 [dev-dependencies]
+assert_matches = "1.5"
 sha3 = "0.9.1"
 unsigned-varint = { version = "0.7.0", features = ["futures", "asynchronous_codec"] }
 tokio = { version = "1.6.1", features = ["macros", "rt", "rt-multi-thread"] }
--- a/src/member.rs
+++ b/src/member.rs
@@ -68,7 +68,7 @@ pub(crate) enum NotificationIn<H: Hasher> {
 pub(crate) enum NotificationOut<H: Hasher> {
     /// Notification about a preunit created by this Consensus Node. Member is meant to
     /// disseminate this preunit among other nodes.
-    CreatedPreUnit(PreUnit<H>),
+    CreatedPreUnit(PreUnit<H>, Vec<H::Hash>),
     /// Notification that some units are needed but missing. The role of the Member
     /// is to fetch these unit (somehow).
     MissingUnits(Vec<UnitCoord>),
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -26,6 +26,7 @@ codec = { package = "parity-scale-codec", version = "2", default-features = fals
 parking_lot = "0.11"
 
 [dev-dependencies]
+assert_matches = "1.5"
```
timorl, line 29: Is this used anywhere?

src/testing/consensus.rs
```diff
@@ -56,14 +56,6 @@ async fn agree_on_first_batch() {
...
         assert_eq!(batches[0], batches[node_ix]);
     }
-
-    exits.into_iter().for_each(|tx| {
-        let _ = tx.send(());
-    });
-
-    for handle in &mut handles {
-        handle.await.expect("All nodes are honest.");
-    }
```
timorl, line 66: Why did this disappear?

timorl (changes requested):
Looks mostly good.

src/member.rs
```diff
@@ -185,13 +185,14 @@ where
...
 
-    async fn on_create(&mut self, u: PreUnit<H>) {
+    async fn on_create(&mut self, u: PreUnit<H>, parent_hashes: Vec<H::Hash>) {
         debug!(target: "AlephBFT-member", "{:?} On create notification.", self.index());
         let data = self.data_io.get_data();
         let full_unit = FullUnit::new(u, data, self.config.session_id);
         let hash = full_unit.hash();
         let signed_unit = Signed::sign(full_unit, self.keybox).await;
         self.store.add_unit(signed_unit, false);
+        self.store.add_parents(hash, parent_hashes);
```
mike1729, line 195: Maybe you could add a comment here (or some other places) highlighting the importance of this change. There is a task for optimizing how we store parents and it would be nice to make sure we won't reintroduce the bug again.

mike1729 (approved):
Just a small request but I approve nonetheless.

src/testing/consensus.rs
DamianStraszak: Once in *lots of* runs this panics at `thread 'testing::consensus::agree_on_first_batch' panicked at 'Channel for batches should be open: TrySendError { kind: Disconnected }', src/extender.rs:155:14`. Both after the changes and on main. I'm pretty sure this is because of the issues with "graceful exit", which zbyszko is trying to fix. This test is not testing graceful exits, so I removed it.
This is super hard to reproduce also, as I'm getting it only during `.\run_local_pipeline` and for instance never when I repeatedly run this single test `agree_on_first_batch` (1000 runs, 0 fails).

Cargo.toml
DamianStraszak: Oh right -- this was used in a test that I removed after changing my strategy for fixing this bug :) Removed dependency.

src/member.rs
DamianStraszak: Right, I added a comment clarifying this.

timorl (approved):

PR #132, opened 2021-08-04
A0-231: Two improvements of requests in member

michalseweryn:
Two improvements of requests in member.

* when multiple units trigger a "missing unit" notification for their common parent, multiple tasks requesting the unit with an appropriate coordinates were created. Now we have at most one task for every (creator, round) coordinate.
* when we need parents of the unit with a given hash, we used to periodically send requests to random peers. This PR optimizes it by always sending the first request for a given hash to the peer who created the corresponding unit (who is most likely to own ).

EDIT: currently we also use the latter improvement in the case when we need a unit with given coordinates

### DIFF
--- a/src/member.rs
+++ b/src/member.rs
@@ -85,17 +85,20 @@ pub(crate) enum NotificationOut<H: Hasher> {
 
 #[derive(Eq, PartialEq)]
 enum Task<H: Hasher> {
+    // Request the unit with the given hash.
     CoordRequest(UnitCoord),
-    // Request parents of the unit with a given hash. The bool flag determines whether we request parents for the first time
-    ParentsRequest(H::Hash, bool),
-    // The hash of a unit, and the number of this multicast (i.e., how many times was the unit multicast already).
-    UnitMulticast(H::Hash, usize),
+    // Request parents of the unit with the given hash.
+    ParentsRequest(H::Hash),
+    // Broadcast the unit with the given hash.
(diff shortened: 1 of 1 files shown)


src/member.rs
```diff
@@ -251,15 +259,28 @@ where
...
+        first_time: bool,
+        curr_time: time::Instant,
+    ) {
         if self.store.get_parents(u_hash).is_none() {
             let message = UnitMessage::<H, D, MK::Signature>::RequestParents(self.index(), u_hash);
-            let peer_id = self.random_peer();
+            let peer_id = if first_time {
+                self.store
+                    .unit_by_hash(&u_hash)
+                    .expect("We request parents of units fromm store")
```
DamianStraszak, line 273: Typo `fromm`

DamianStraszak (approved):

src/member.rs
```diff
@@ -80,23 +85,28 @@ pub(crate) enum NotificationOut<H: Hasher> {
 
 #[derive(Eq, PartialEq)]
 enum Task<H: Hasher> {
+    // Request the unit with the given hash.
```
Yithis, line 88: Typo

Yithis (approved):

src/member.rs
michalseweryn: fixed

src/member.rs
michalseweryn: fixed

PR #133, opened 2021-08-05
A0-212: Refactor code to allow nodes request old units

Yithis:


Yithis:
@DamianStraszak @michalseweryn  what do you think about this PR?

src/units/store.rs
```diff
@@ -99,28 +70,9 @@ impl<'a, H: Hasher, D: Data, KB: KeyBox> UnitStore<'a, H, D, KB> {
             warn!(target: "AlephBFT-unit-store", "Trying to mark the node {:?} as forker for the second time.", forker);
         }
         self.is_forker[forker] = true;
-        let forkers_units = (0..=self.round_in_progress)
+        (0..=self.max_round)
```
michalseweryn, line 73: Since `round_in_progress` is no longer there, update the comment above this function.

src/units/store.rs
```diff
@@ -152,10 +104,9 @@ impl<'a, H: Hasher, D: Data, KB: KeyBox> UnitStore<'a, H, D, KB> {
         // way beyond round_in_progress right away to Consensus. This could be perhaps corrected so that
         // we wait until the round is in progress, but this does not seem to help vs actual attacks and in
         // "accidental" forks the rounds will never be much higher than round_in_progress.
-        if alert || (round <= self.round_in_progress && !self.is_forker[creator]) {
```
michalseweryn, line 155: Again, update the comment above

src/units/store.rs
```diff
@@ -284,21 +230,19 @@ mod tests {
...
         for (round, hash) in forker_hashes[5..7].iter().enumerate() {
             let round = round as Round;
             let round = round + 5;
             let coord = UnitCoord::new(round, NodeIndex(0));
-            assert_eq!(NodeCount(0), store.n_units_per_round[round as usize]);
-            assert!(!store.by_coord.contains_key(&coord));
-            assert!(!store.by_hash.contains_key(hash));
+            assert_eq!(NodeCount(1), store.n_units_per_round[round as usize]);
+            assert!(store.by_coord.contains_key(&coord));
+            assert!(store.by_hash.contains_key(hash));
```
michalseweryn, line 245: As you mentioned to me outside gitub, this indeed is a silly way to fix test. Could you possibly instead check if the alert sent to `alerts_for_alerter` is correct? this would involve adding a setter for `alerts_for_alerter`, which is not the cleanest solution, but seems more reasonable.You could reuse the setter in `run_session`

michalseweryn (changes requested):
Nice PR, simplifies the store a lot :)

src/units/store.rs
```diff
@@ -9,9 +9,6 @@ pub(crate) struct UnitStore<'a, H: Hasher, D: Data, KB: KeyBox> {
     by_coord: HashMap<UnitCoord, SignedUnit<'a, H, D, KB>>,
     by_hash: HashMap<H::Hash, SignedUnit<'a, H, D, KB>>,
     parents: HashMap<H::Hash, Vec<H::Hash>>,
-    //this is the smallest r, such that round r-1 is saturated, i.e., it has at least threshold (~(2/3)N) units
-    round_in_progress: Round,
-    threshold: NodeCount,
     //the number of unique nodes that we hold units for a given round
     n_units_per_round: Vec<NodeCount>,
```
DamianStraszak, line 13: This is likely not needed. Perhaps one or more tests can be removed because of this.

src/units/store.rs
```diff
@@ -152,10 +104,9 @@ impl<'a, H: Hasher, D: Data, KB: KeyBox> UnitStore<'a, H, D, KB> {
         // way beyond round_in_progress right away to Consensus. This could be perhaps corrected so that
```
DamianStraszak, line 104: You can remove these 4 lines of comment. It's outdated.

michalseweryn (approved):

DamianStraszak (approved):

PR #135, opened 2021-08-11
A0-114: unit handling refactored out of member

fixxxedpoint:


src/consensus.rs
```diff
@@ -1,107 +1,139 @@
...
+        let extender_barrier = barrier.clone("consensus/extender");
+        let (extender_exit, exit_rx) = oneshot::channel();
+        let extender_handle = self
+            .spawn_handle
+            .spawn_essential("consensus/extender", async move {
+                extender.extend(exit_rx).await;
+
+                extender_barrier.wait().await;
+            });
+        let mut extender_handle = into_infinite_stream(extender_handle).fuse();
```
michalseweryn, line 65: Yuck! Converting it to a stream is so artificial, given that we will poll at most two elements... How about making it a fused future and replacing  `alerter_handle.next().await.unwrap()` at the end of run with something like 
`alerter_handle.now_or_never().unwrap_or(())`?

src/utils.rs
```diff
@@ -0,0 +1,81 @@
+use futures::{channel::oneshot, future::ready, stream::iter, Future, FutureExt, Stream};
+use log::warn;
+use std::iter::repeat;
+
+pub(crate) fn into_infinite_stream<F: Future>(f: F) -> impl Stream<Item = ()> {
+    f.then(|_| ready(iter(repeat(())))).flatten_stream()
+}
```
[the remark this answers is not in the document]
fixxxedpoint: Simplified. It's better than ad-hoc declaration or ad-hoc re-use and removes tedious and unnecessary noisy carrying of flags for each service while we are attempting to exit.

src/runway.rs
```diff
@@ -0,0 +1,855 @@
...
+        let outgoing_messages = self.outgoing_messages;
+        let incoming_messages = self.incoming_messages;
+        (
+            RunwayFacade::new(runway_exit, outgoing_messages, incoming_messages),
+            async move { runway.run(exit_stream, alerter, consensus).await },
+        )
+    }
+}
+
+pub(crate) struct Runway<H, D, MK, DP, SH>
```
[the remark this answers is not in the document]
fixxxedpoint: It's no longer owning them and they are spawned parallel to it. Further refactor will be part of another PR that is going simplify how all services all started and managed. Not going much into details about the second part: it's sounds like breaking `the single responsibility principle` which will make testing even more difficult.

src/consensus.rs
fixxxedpoint: I agree it is `yuck`, but just pretty convenient. I will try to think of some cleaner solution.

src/runway.rs
```diff
@@ -0,0 +1,855 @@
...
+use futures::{
+    channel::{mpsc, oneshot},
+    Future, StreamExt,
+};
+use log::{debug, error, info, trace, warn};
+
+/// Type for incoming notifications: Runway to Consensus.
+#[derive(Clone, PartialEq)]
+pub(crate) enum NotificationIn<H: Hasher> {
+    /// A notification carrying a single unit. This might come either from multicast or
```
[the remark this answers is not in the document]
fixxxedpoint: fixed

src/member.rs
```diff
@@ -136,86 +132,119 @@ impl<H: Hasher> PartialOrd for ScheduledTask<H> {
 /// For a detailed description of the consensus implemented in Member see
 /// [docs for devs](https://cardinal-cryptography.github.io/AlephBFT/index.html)
 /// or the [original paper](https://arxiv.org/abs/1908.05156).
-pub struct Member<'a, H, D, DP, MK, SH>
+pub struct Member<'a, D, DP, MK, SH>
```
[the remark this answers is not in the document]
fixxxedpoint: it's no longer an issue for the new version

src/consensus.rs
michalseweryn: Is thar aything wrong with the solution I proposed?
```
let extender_handle = extender_handle.fuse();
...
select! {
    ...
}
...
extender_handle.now_or_never().unwrap_or(())
```

fixxxedpoint: > Evaluates and consumes the future, returning the resulting output if the future is ready after the first call to Future::poll. If poll instead returns Poll::Pending, None is returned. This method is useful in cases where immediacy is more important than waiting for a result. It is also convenient for quickly obtaining the value of a future that is known to always resolve immediately.

My understanding is: if that `extender_handle` didn't return `Poll::Ready()` inside of that `select` statement, then that `extender_handle.now_or_never().unwrap_or(())` will not wait till it returns `Poll::Ready` and we want to exit from this method after we know that extender already finished.

timorl (changes requested):
Lots of comments, but I kind of expect that the responses to many of them will be of the nature "this will be done in the future". In some cases I would like this formalized (e.g. added to the ADR), but this won't be a merge blocker.

Ignore the comment about `'a` removal, but try finding the remaining `'a`s, if you haven't done so yet (it's an old comment and I haven't rechecked that there are still some stragglers, a `grep` on your version of the code should be sufficient <_<").

src/member.rs
```diff
@@ -111,111 +83,96 @@ impl<H: Hasher> ScheduledTask<H> {
...
 
-impl<'a, H, D, DP, MK, SH> Member<'a, H, D, DP, MK, SH>
+impl<H, D, S> Member<H, D, S>
 where
     H: Hasher,
     D: Data,
-    DP: DataIO<D>,
-    MK: MultiKeychain,
-    SH: SpawnHandle,
+    S: Signature + Eq,
```
timorl, line 121: Unnecessary `Eq`. Kinda surprized clippy does not pick that up.

src/testing/byzantine.rs
```diff
@@ -52,7 +52,7 @@ impl<'a> MaliciousMember<'a> {
         }
     }
 
-    fn unit_to_data(su: SignedUnit<'a, Hasher64, Data, KeyBox>) -> NetworkData {
+    fn unit_to_data(su: SignedUnit<Hasher64, Data, KeyBox>) -> NetworkData {
```
timorl, line 55: Uh, what? How does this not have `'a` now? Does this even compile? If not then I'm kinda worried that our CI does not catch it. <_<

timorl (changes requested):
Only one nitpick and one confusion, other than that LGTM.

Perhaps write your suggestion about removing `parking_lot` somewhere where it will not be forgotten.

src/testing/byzantine.rs
michalseweryn: [Lifetime ellision](https://doc.rust-lang.org/reference/lifetime-elision.html). Although it does compile, `SignedUnit<'_, Hasher64, Data, KeyBox>` is preferred according to rust documentation

src/member.rs
```diff
@@ -111,111 +84,96 @@ impl<H: Hasher> ScheduledTask<H> {
...
+    not_resolved_parents: HashSet<H::Hash>,
+    not_resolved_coords: HashSet<UnitCoord>,
     n_members: NodeCount,
-    unit_messages_for_network: Option<Sender<(UnitMessage<H, D, MK::Signature>, Recipient)>>,
-    alerts_for_alerter: Option<Sender<Alert<H, D, MK::Signature>>>,
-    spawn_handle: SH,
+    unit_messages_for_network: Sender<(UnitMessage<H, D, S>, Recipient)>,
+    unit_messages_from_network: Receiver<UnitMessage<H, D, S>>,
+    messages_from_network: Sender<RunwayNotificationIn<H, D, S>>,
+    messages_for_network: Receiver<RunwayNotificationOut<H, D, S>>,
```
michalseweryn, line 114: Shouldn't these two be named `notifications_for_runway` and `notifications_from_runway` respecitively?

michalseweryn (changes requested):

timorl (approved):
LGTM

src/runway.rs
```diff
@@ -0,0 +1,767 @@
...
+    NewUnit(UncheckedSignedUnit<H, D, S>),
+    Request(ROut),
+    Response(RIn),
+}
+
+pub(crate) type RunwayNotificationOut<H, D, S> =
+    RunwayNotification<H, D, S, RequestOut<H>, ResponseOut<H, D, S>>;
+
+pub(crate) type RunwayNotificationIn<H, D, S> =
+    RunwayNotification<H, D, S, RequestIn<H>, Response<H, D, S>>;
```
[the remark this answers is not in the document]
fixxxedpoint: changed

src/member.rs
fixxxedpoint: renamed

michalseweryn (approved):

PR #136, opened 2021-08-12
FIX: towards graceful exit - creator.rs was calling a possibly "finished" stream

fixxxedpoint:
- FIX: parents_rx in creator.rs was called multiple times after it returned None (calling next() after stream is empty) blocking creator from gracefully exiting.
- removed "as usize" calls from creator

### DIFF
--- a/src/creator.rs
+++ b/src/creator.rs
@@ -136,6 +136,7 @@ impl<H: Hasher> Creator<H> {
                 self.add_unit(u.round(), u.creator(), u.hash());
             } else {
                 warn!(target: "AlephBFT-creator", "{:?} get error as result from channel with parents.", self.node_ix);
+                return;
             }
         }
     }


src/creator.rs
```diff
@@ -55,7 +55,7 @@ impl<H: Hasher> Creator<H> {
 
     // initializes the vectors corresponding to the given round (and all between if not there)
     fn init_round(&mut self, round: Round) {
-        if (round + 1) as usize > self.n_candidates_by_round.len() {
+        if usize::from(round + 1) > self.n_candidates_by_round.len() {
```
Yithis, line 58: As far as I know currently old way is consistent with rest of the repo. Did you already change this in other files?

fixxxedpoint: I don't see any point in fixing it everywhere else in this PR. It's just a small local improvement. I'm not attached to it. Just changed it since quite often I'm asked to do similar improvements when somebody notice something worth to be bothered.

fixxxedpoint: I am more concerned about it's intended behavior. Is it safe to call `self.create_unit()` if this method exits before reaching its intended state?

src/creator.rs
```diff
@@ -43,7 +43,7 @@ impl<H: Hasher> Creator<H> {
         let n_members = conf.n_members;
         Creator {
             node_ix: conf.node_ix,
-            parents_rx,
+            parents_rx: parents_rx.fuse(),
```
michalseweryn, line 46: I don't think there is a need for that. `Receiver` does implement `FusedStream`: https://docs.rs/futures/0.3.16/futures/channel/mpsc/struct.UnboundedReceiver.html#impl-FusedStream . I dont think the usage of `futures::select!` macro would work if `Receiver` weren't `FusedStream`

michalseweryn (changes requested):

timorl (changes requested):
I think this PR should essentially be just the single `return` line, the rest – the `as` to `from` change and the `Fuse` change, if the latter is even necessary – should happen in separate ones that would make these approaches consistent over the codebase. They are both pretty mechanical and can be done quickly repo-wide.

src/creator.rs
fixxxedpoint: removed

timorl (approved):
Generally OK, just the introduced `usize` inconsistency really irks me. If merge this as-is I'll create a PR that makes it consistent again (by changing all the files the way you changed `creator.rs`) I guess.

michalseweryn (approved):

PR #137, opened 2021-09-01
A0-249 + A0-250: Node crash recovery

michalseweryn:
The version of https://github.com/Cardinal-Cryptography/AlephBFT/pull/134 built on top of `member.rs` refactor (https://github.com/Cardinal-Cryptography/AlephBFT/pull/135)

### DIFF
--- a/src/creator.rs
+++ b/src/creator.rs
@@ -5,9 +5,12 @@ use crate::{
     units::{ControlHash, PreUnit, Unit},
     Hasher, Receiver, Round, Sender,
 };
-use futures::{channel::oneshot, FutureExt, StreamExt};
+use futures::{
+    channel::{mpsc::TrySendError, oneshot},
+    FutureExt, StreamExt,
+};
 use futures_timer::Delay;
-use log::{info, trace, warn};
+use log::{error, info, trace, warn};
 use std::time::Duration;
 
 /// A process responsible for creating new units. It receives all the units added locally to the Dag
@@ -63,7 +66,7 @@ impl<H: Hasher> Creator<H> {
         }
     }
 
-    fn create_unit(&mut self, round: Round) {
+    fn create_unit(&mut self, round: Round) -> Result<(), TrySendError<NotificationOut<H>>> {
         let parents = {
             if round == 0 {
                 NodeMap::new_with_len(self.n_members)
@@ -77,11 +80,10 @@ impl<H: Hasher> Creator<H> {
 
         let new_preunit = PreUnit::new(self.node_ix, round, control_hash);
         trace!(target: "AlephBFT-creator", "{:?} Created a new unit {:?} at round {:?}.", self.node_ix, new_preunit, round);
-        self.new_units_tx
-            .unbounded_send(NotificationOut::CreatedPreUnit(new_preunit, parent_hashes))
-            .expect("Notification channel should be open");
-
+        let notification = NotificationOut::CreatedPreUnit(new_preunit, parent_hashes);
+        self.new_units_tx.unbounded_send(notification)?;
         self.init_round(round + 1);
+        Ok(())
     }
 
     fn add_unit(&mut self, round: Round, pid: NodeIndex, hash: H::Hash) {
@@ -129,7 +131,8 @@ impl<H: Hasher> Creator<H> {
         // Additionally, our unit from previous round must be available.
         let threshold = (self.n_members * 2) / 3 + NodeCount(1);
 
-        while self.n_candidates_by_round[prev_round_index] < threshold
+        while self.n_candidates_by_round.len() <= prev_round_index
+            || self.n_candidates_by_round[prev_round_index] < threshold
             || self.candidates_by_round[prev_round_index][self.node_ix].is_none()
         {
             if let Some(u) = self.parents_rx.next().await {
@@ -141,8 +144,9 @@ impl<H: Hasher> Creator<H> {
         }
     }
 
-    pub(crate) async fn create(&mut self, mut exit: oneshot::Receiver<()>) {
-        for round in 0..self.max_round {
+    pub(crate) async fn create(&mut self, starting_round: Round, mut exit: oneshot::Receiver<()>) {
+        log::debug!(target: "AlephBFT-creator", "Creator starting from round {}", starting_round);
+        for round in starting_round..self.max_round {
             let mut delay = Delay::new(Duration::from_secs(30 * 60)).fuse();
             loop {
                 futures::select! {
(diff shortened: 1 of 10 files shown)


src/runway.rs
```diff
@@ -440,10 +490,77 @@ where
...
+                let starting_round_candidate = checked_unit.as_signable().round() + 1;
+                self.on_unit_received(checked_unit.into(), false);
+                if starting_round_candidate > self.starting_round_value {
+                    self.starting_round_value = starting_round_candidate;
+                }
+            }
+        }
+        self.newest_unit_responders.insert(response.responder);
+        if self.is_starting_round_ready() {
+            trace!(target: "AlephBFT-runway", "sending starting round to creator: {}", self.starting_round_value);
```
DamianStraszak, line 543: Shall we send to `self.resolved_requests` somewhere here?

src/runway.rs
```diff
@@ -440,10 +490,77 @@ where
...
+        if response.salt != self.salt {
+            debug!(target: "AlephBFT-member", "Ignoring newest unit response with an unknown salt: {:?}", response);
+            return;
+        }
+
+        if let Some(unchecked_unit) = response.unit {
+            if unchecked_unit.as_signable().creator() != self.index() {
+                log::debug!(target: "AlephBFT-member", "Not our unit in a response:  {:?}", unchecked_unit.into_signable());
+                return;
+            }
```
DamianStraszak, line 521: I would suggest running `validate_unit` on the received unit to avoid some strange corner cases. For instance now it seems we forgot to check is the `session_id` is correct.

DamianStraszak (changes requested):

src/runway.rs
```diff
@@ -440,10 +490,77 @@ where
...
+            {
+                let starting_round_candidate = checked_unit.as_signable().round() + 1;
+                self.on_unit_received(checked_unit.into(), false);
+                if starting_round_candidate > self.starting_round_value {
+                    self.starting_round_value = starting_round_candidate;
+                }
+            }
+        }
+        self.newest_unit_responders.insert(response.responder);
+        if self.is_starting_round_ready() {
```
DamianStraszak, line 542: I'm not entirely sure if this will work correctly if all nodes respond right away (before the delay is over). Will this ever trigger then?

src/runway.rs
```diff
@@ -715,10 +847,12 @@ pub(crate) async fn run<H, D, MK, DP, SH>(
         unit_messages_for_network: runway_io.unit_messages_for_network,
         ordered_batch_rx,
         resolved_requests: runway_io.resolved_requests,
+        starting_round_sender,
         node_ix: config.node_ix,
         session_id: config.session_id,
         n_members: config.n_members,
         max_round: config.max_round,
+        salt: 0,
```
timorl, line 855: Shouldn't this be... random or something?

src/runway.rs
```diff
@@ -115,6 +122,11 @@ where
     rx_consensus: Receiver<NotificationOut<H>>,
     ordered_batch_rx: Receiver<Vec<H::Hash>>,
     data_io: DP,
+    after_catch_up_delay: bool,
+    starting_round_sender: Option<oneshot::Sender<Round>>,
```
timorl, line 126: I hate that all this data that is only needed during setup is part of the struct forever. I don't think we can expect to fix this now, since it might require a larger rewrite that we don't necessarily have time for, but we at least need an issue for that.

timorl (changes requested):

src/member.rs
```diff
@@ -299,8 +357,12 @@ where
 
                 event = self.resolved_requests.next() => match event {
                     Some(request) => match request {
-                        Request::RequestCoord(coord) => { self.not_resolved_coords.remove(&coord); },
-                        Request::RequestParents(u_hash) => { self.not_resolved_parents.remove(&u_hash); },
+                        Request::Coord(coord) => { self.not_resolved_coords.remove(&coord); },
+                        Request::Parents(u_hash) => { self.not_resolved_parents.remove(&u_hash); },
+                        Request::NewestUnit(_salt) => {
+                            // TODO: ensure that the salt is right?
```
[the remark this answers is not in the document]
DamianStraszak: I think the point is that, it is impossible that the salt is incorrect, that's why the question mark. Because in one run, we use one value of salt and accept only responses with this particular salt. I think keeping it this way should be fine...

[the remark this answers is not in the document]
timorl: Makes sense, although then I would prefer `_` to `_salt` and remove this TODO.

src/runway.rs
michalseweryn: Yes we should, fixed.

src/runway.rs
michalseweryn: It won't :sweat_smile:. Fixed that.

src/runway.rs
michalseweryn: Added a task in Jira

src/testing/dag.rs
```diff
@@ -131,9 +131,22 @@ async fn run_consensus_on_dag(
     let (_exit_tx, exit_rx) = oneshot::channel();
     let (batch_tx, mut batch_rx) = mpsc::unbounded();
     let spawner = Spawner::new();
+    let starting_round = {
```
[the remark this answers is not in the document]
michalseweryn: Added a `complete_oneshot` function.

src/runway.rs
michalseweryn: Fixed

DamianStraszak (approved):

src/member.rs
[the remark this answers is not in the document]
michalseweryn: Removed :>

src/testing/mock.rs
```diff
@@ -583,8 +583,8 @@ pub fn spawn_honest_member(
     (rx_batch, exit_tx, handle)
 }
 
-pub fn complete_oneshot<T>(t: T) -> oneshot::Receiver<T> {
+pub fn complete_oneshot<T: std::fmt::Debug>(t: T) -> oneshot::Receiver<T> {
```
timorl, line 586: Why does `T` need `Debug`?

timorl (approved):
Modulo the comment about `Debug` above.

src/testing/mock.rs
Yithis: There is a compilation error without it

timorl (changes requested):
Small possible problems.

src/creator.rs
```diff
@@ -65,7 +68,7 @@ impl<H: Hasher> Creator<H> {
         }
     }
 
-    fn create_unit(&mut self, round: Round) {
+    fn create_unit(&mut self, round: Round) -> Result<(), TrySendError<NotificationOut<H>>> {
```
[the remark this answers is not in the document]
michalseweryn: This was used for early exit before graceful exit was introduced. Fixed.

timorl (approved):

PR #140, opened 2021-09-06
A0-367: Fix AlephBFT example

michalseweryn:
An update of libp2p broke our example. This PR fixes that.

### DIFF
--- a/examples/blockchain/network.rs
+++ b/examples/blockchain/network.rs
@@ -10,7 +10,7 @@ use futures::{
     Future, FutureExt, StreamExt,
 };
 
-use log::{debug, info, warn};
+use log::{debug, info, trace, warn};
 
 use std::{collections::HashMap, error::Error, io, iter, time::Duration};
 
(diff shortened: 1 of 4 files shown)


timorl (dismissed):
If it works then LGTM.

examples/blockchain/network.rs
```diff
@@ -372,8 +372,14 @@ impl NetworkManager {
                     }
                 }
                 event = self.swarm.next().fuse() => {
```
DamianStraszak, line 374: There is also the second example: `dummy_honest.rs` -- if we want to maintain it, then this likely also needs to be changed there.

Another option would be to drop the `dummy_honest` example :)

timorl (approved):

examples/blockchain/network.rs
timorl: I think I support dropping it – if we are planning to simplify and improve (i.e. remove libp2p ;P) the main example then it should be enough for any documentation/tutorial purposes.

michalseweryn: Fixed :)

timorl: I think that is the most disgusting solution you could have picked. xd

Not that I object much.

PR #142, opened 2021-09-17
A0-354: Refactor creator for easier testability

timorl:


src/creation/creator.rs
```diff
@@ -0,0 +1,301 @@
...
+        if self.candidates_by_round[round as usize][pid].is_none() {
+            // passing the check above means that we do not have any unit for the pair (round, pid) yet
+            self.candidates_by_round[round as usize][pid] = Some(hash);
+            self.n_candidates_by_round[round as usize] += NodeCount(1);
+        }
+    }
+
+    /// Check whether the provided round is far behind the current round, meaning the unit of that
+    /// round should be created without delay.
+    pub(super) fn is_behind(&self, round: Round) -> bool {
```
Yithis, line 70: Maybe a better name? Or is it not intuitive just for me

src/testing/creation.rs
```diff
@@ -0,0 +1,311 @@
...
+    assert!(test_controller
+        .units_per_creator
+        .iter()
+        .all(|r| *r >= (max_round - 3) as usize));
+    assert!(
+        test_controller
+            .units_per_creator
+            .iter()
+            .map(
+                // There should be an easier way to do this...
```
Yithis, line 229: I don't have a better idea, maybe `.fold()`? Still marking this line, coz this comment should be deleted before merge. Hope other review will help with this

Yithis (commented):
All my comments are just notes and personal feelings from reading. Overall looks great!

src/creation/creator.rs
```diff
@@ -0,0 +1,301 @@
...
+        let round = 1;
+        assert!(!creator.can_create(round));
+    }
+
+    #[test]
+    fn creates_two_units_when_possible() {
+        let n_members = NodeCount(7);
+        let mut creators = creator_set(n_members);
+        let mut expected_hashes_per_round = Vec::new();
+        for round in 0..=1 {
```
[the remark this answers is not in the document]
timorl: Since we number rounds starting at 0 this was more intuitive for me, since it shows the highest round involved. I was on the fence though, so changing.

src/creation/creator.rs
timorl: I thought about it and I have no better ideas for a name. :/ Suggestions welcome.

src/testing/creation.rs
timorl: Yeah, I was hoping a reviewer would figure something better out and then I could delete this. I think `fold` is marginally better, but I just had an idea to use `filter`+`count` which looks acceptable to me.

src/consensus.rs
```diff
@@ -37,27 +37,24 @@ pub(crate) async fn run<H: Hasher + 'static>(
...
     let mut creator_handle = spawn_handle
-        .spawn_essential("consensus/creator", async move {
-            match starting_round.await {
-                Ok(round) => creator.create(round, exit_rx).await,
-                Err(e) => {
-                    error!(target: "AlephBFT-creator", "Starting round not provided: {}", e);
-                }
-            }
+        .spawn_essential("consensus/creation", async move {
+            creation::run(conf.clone().into(), io, starting_round, exit_rx).await;
```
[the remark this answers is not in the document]
timorl: `Round` doesn't really `into` `usize`, and `Config` cannot work with `as`. I think I remember you suggested some better way of `Round` conversion in some earlier PRs but I don't remember what it was, if you can remind me I can change to that (although we really should get an ADR/standard for that).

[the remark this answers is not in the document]
timorl: Although I managed to remove some `as`ses for unrelated reasons.

Yithis (approved):

fixxxedpoint (changes requested):
Just a few things:
- Is that `pub(super)` needed for anything? It seems like creator has a clear public interface which doesn't need to be restricted on this level.
- comment below about `create_unit` being a partial function
- the whole public interface of creator is not very well defined, i.e. you can call twice `create_unit` for the same round but there is a chance that you will receive two different units (it collected more units for some round meanwhile)
- it's not part of this refactor but assuming that units are added to creator in round-by-round manner, keeping every round stored in `Vecs` seems excessive. But modifying it not necessarily would make anything simpler/easier to understand.

src/creation/creator.rs
```diff
@@ -0,0 +1,301 @@
...
+            self.n_candidates_by_round.resize(new_size, NodeCount(0));
+        }
+    }
+
+    pub(super) fn create_unit(&self, round: Round) -> (PreUnit<H>, Vec<H::Hash>) {
+        let parents = {
+            if round == 0 {
+                NodeMap::new_with_len(self.n_members)
+            } else {
+                self.candidates_by_round[(round - 1) as usize].clone()
```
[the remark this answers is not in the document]
timorl: Thanks for the comments, very useful.
* If we export it as `pub` we would have to document & support it. Also, I'm pretty sure some lint will be unhappy about leaking private types in public interfaces.
* Fixed, now it checks whether it `can_create` before actually doing anything, if it cannot it returns `Err(())`.
* This is mostly by design, I didn't want `Creator` to contain knowledge about the round, since this is being controlled outside. I can redesign this, with the main tradeoff (I think?) being making the creation of forks harder vs making creator itself more complicated. I would prefer for it to stay as-is, but I'm open to arguments to the contrary.
* Removed `is_behind`, made `current_round` public. Now `run` checks the condition explicitly. I think this is actually better.
* This assumption is wrong afaik, you can receive an ancient unit after a long time and then the units will not be passed round-by-round.
Could you make comments in separate, uh, comments in the future? It's easier to keep the discussion organized then.

src/creation/creator.rs
timorl: Thanks to @fixxxedpoint's comments this has been removed completely. :3

src/creation/creator.rs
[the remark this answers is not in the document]
fixxxedpoint: `pub` method of a non-`pub`/non-`exported` type is not `pub` externally?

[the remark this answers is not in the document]
fixxxedpoint: > This assumption is wrong afaik, you can receive an ancient unit after a long time and then the units will not be passed round-by-round.
Let me reword this: round-by-round, i.e. when you receive a unit from round `r+1` in the creator, then you know that you're able to create a unit for round `r` (but you can still wait for some more unit if you want). My intention was also: if it would be redesigned this way, then it would be not-possible to create forks (data for a round in which we already created a unit is disposed).

src/creation/creator.rs
```diff
@@ -0,0 +1,277 @@
...
+        if self.candidates_by_round[round as usize][pid].is_none() {
+            // passing the check above means that we do not have any unit for the pair (round, pid) yet
+            self.candidates_by_round[round as usize][pid] = Some(hash);
+            self.n_candidates_by_round[round as usize] += NodeCount(1);
+        }
+    }
+
+    /// To create a new unit, we need to have at least floor(2*N/3) + 1 parents available in previous round.
+    /// Additionally, our unit from previous round must be available.
+    pub(super) fn can_create(&self, round: Round) -> bool {
```
fixxxedpoint, line 74: Consider making this method private since `create_unit` now returns `Result`. It seems that the only non-trivial usage is by `wait_until_ready`.

src/testing/creation.rs
```diff
@@ -0,0 +1,293 @@
...
+    );
+    finish(killers, handles).await;
+}
+
+// Catching up test
+// This test checks if 5 creators that start at the same time and 2 creators
+// that start after those first 5 reach round 25,
+// will reach at least round 47, when at least one reaches round 50.
+#[tokio::test(flavor = "multi_thread", worker_threads = 3)]
+async fn late_creators_should_create_dag() {
```
fixxxedpoint, line 240: The idea is ok but its execution is a bit tricky. We should avoid such tests. I am not sure, but it looks like it can fail non-deterministically. I think we have (almost?) all ingredients in place to write these tests `deterministically`.

fixxxedpoint (changes requested):
One nitpick regarding public interface and some bigger request for tests. Also, perhaps it would be nice to have some test like: a unit created within/after the `delay` should contain all available parents at that time (not just the threshold).

src/creation/creator.rs
timorl: That is the only usage, but I don't quite see how to remove it without significantly changing the logic, and I don't have an idea for a good version of the changed logic.

src/creation/creator.rs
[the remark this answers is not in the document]
timorl: > pub method of a non-pub/non-exported type or module is not pub externally?

Clippy still complains about leakage. It's also potentially less readable, especially if we do it differently in different modules?

I think this would work with clippy if we rewrote everything to the setup where everything is pub within private modules and then only the things we want exported are reexported in the module. This would require quite a lot of work though, and I'm not even sure it would be better. Some standards would be nice, as usual.

> Let me reword this:

Ah, I see what you mean. Yeah, we could do that although currently tests don't keep that assumption. Also, I think you are making a broader point about how you imagine the design/API of creator that I am not quite getting – if that is right, lets discuss this in a more live setting.

src/testing/creation.rs
timorl: I don't think we can test for this deterministically – after all we are essentially checking only whether specific delays are skipped. I agree that this test is not ideal (and technically can fail), but I don't see how else we can check this, especially with `is_behind` now removed.

src/creation/creator.rs
fixxxedpoint:
```
async fn wait_until_ready<H: Hasher>(
    round: Round,
    creator: &mut Creator<H>,
    create_lag: &DelaySchedule,
    mut ignore_delay: bool,
    incoming_parents: &mut Receiver<Unit<H>>,
    mut exit: &mut oneshot::Receiver<()>,
) -> Result<(PreUnit<H>, Vec<H::Hash>), ()> {
    let mut delay = Delay::new(create_lag(round.into())).fuse();
    let mut result = if ignore_delay {
        creator.create_unit(round)
    } else {
        Err(())
    };
    while !ignore_delay || result.is_err() {
        futures::select! {
           unit = incoming_parents.next() => match unit {
                Some(unit) => {
                    creator.add_unit(&unit);
                    if ignore_delay {
                        result = creator.create_unit(round);
                    }
                },
                None => {
                    info!(target: "AlephBFT-creator", "Incoming parent channel closed, exiting.");
                    return Err(());
                }
            },
            _ = &mut delay => {
                if ignore_delay {
                    warn!(target: "AlephBFT-creator", "Delay passed despite us not waiting for it.");
                } else {
                    result = creator.create_unit(round);
                }
                ignore_delay = true;
                delay = Delay::new(Duration::from_secs(30 * 60)).fuse();
            },
            _ = exit => {
                info!(target: "AlephBFT-creator", "Received exit signal.");
                return Err(());
            },
        }
    }
    result
}

/// A process responsible for creating new units. It receives all the units added locally to the Dag
/// via the `incoming_parents` channel. It creates units according to an internal strategy respecting
/// always the following constraints: if round is equal to 0, U has no parents, otherwise for a unit U of round r > 0
/// - all U's parents are from round (r-1),
/// - all U's parents are created by different nodes,
/// - one of U's parents is the (r-1)-round unit by U's creator,
/// - U has > floor(2*N/3) parents.
/// - U will appear in the channel only if all U's parents appeared there before
/// The currently implemented strategy creates the unit U according to a delay schedule and when enough
/// candidates for parents are available for all the above constraints to be satisfied.
///
/// We refer to the documentation https://cardinal-cryptography.github.io/AlephBFT/internals.html
/// Section 5.1 for a discussion of this component.
pub async fn run<H: Hasher>(
    conf: Config,
    io: IO<H>,
    starting_round: oneshot::Receiver<Round>,
    mut exit: oneshot::Receiver<()>,
) {
    let Config {
        node_id,
        n_members,
        create_lag,
        max_round,
    } = conf;
    let mut creator = Creator::new(node_id, n_members);
    let IO {
        mut incoming_parents,
        outgoing_units,
    } = io;
    let starting_round = match starting_round.await {
        Ok(round) => round,
        Err(e) => {
            error!(target: "AlephBFT-creator", "Starting round not provided: {}", e);
            return;
        }
    };
    debug!(target: "AlephBFT-creator", "Creator starting from round {}", starting_round);
    for round in starting_round..max_round {
        // Skip waiting if we are way behind.
        let ignore_delay = creator.current_round() > round + 2 || round == 0;
        let new_unit = wait_until_ready(
            round,
            &mut creator,
            &create_lag,
            ignore_delay,
            &mut incoming_parents,
            &mut exit,
        )
        .await;
        let (unit, parent_hashes) = match new_unit {
            Err(_) => return,
            Ok(u) => u,
        };
        if let Err(e) =
            outgoing_units.unbounded_send(NotificationOut::CreatedPreUnit(unit, parent_hashes))
        {
            warn!(target: "AlephBFT-creator", "Notification send error: {}. Exiting.", e);
            return;
        }
    }
    warn!(target: "AlephBFT-creator", "Maximum round reached. Not creating another unit.");
}
```

timorl: I really don't like how this mixes creation with waiting, especially with `create_unit` now being called in a couple places making it somewhat harder to reason about. :/

timorl: After further discussions rewrote more or less like this, PTAL.

src/creation/creator.rs
```diff
@@ -0,0 +1,268 @@
...
+        assert_eq!(creator.current_round(), 0);
+        let (preunit, parent_hashes) = creator
+            .create_unit(round)
+            .expect("Creation should succeed.");
+        assert_eq!(preunit.round(), round);
+        assert_eq!(parent_hashes, expected_hashes);
+    }
+
+    #[test]
+    fn creates_unit_with_minimal_parents() {
```
DamianStraszak, line 179: I think this test is very useful. Could you make it run not only for n=7 p=5, but also for (n,p) in {(1, 1), (2, 2), (3, 3), (4, 3), (5, 4), (6, 5)}?
The same for `cannot_create_unit_below_parents_threshold` :)

src/creation/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+        Err(e) => {
+            error!(target: "AlephBFT-creator", "Starting round not provided: {}", e);
+            return;
+        }
+    };
+    debug!(target: "AlephBFT-creator", "Creator starting from round {}", starting_round);
+    for round in starting_round..max_round {
+        // Skip waiting if we are way behind.
+        // We have to be two rounds behind, otherwise someone could force us to skip all delays.
+        let ignore_delay = creator.current_round() > round + 2;
```
DamianStraszak, line 120: This is correct now, but it seems to me that if it was `> round ` instead of `> round + 2` then it would be still fine?
If there exists a unit at round `round+1`, then there must be several honest nodes who created a unit at round `round`, thus by creating a unit at `round` we will not be the "fastest" honest node. And by a separate reasoning we show that the fastest honest node always waits the full delay, so it should be fine.

DamianStraszak (changes requested):
I think that the changes indeed make it easier to explicitly test some small functions better. Overall I like the changes and I think that they should be merged. 

That being said, I don't have much knowledge/experience when it comes to design patterns, good programming practices etc., so the opinions of @michalseweryn , @fixxxedpoint , @mike1729 and others should matter more here.

src/creation/mod.rs
timorl: Hm, I think you are right, changing. <_<"

src/creation/creator.rs
DamianStraszak: Hah ok :)

src/testing/creation.rs
fixxxedpoint: Since proposed solution is far from perfect and a good one requires more refactoring, we decided to leave it for future improvement - A0-453 in our jira.

DamianStraszak (approved):

fixxxedpoint (approved):
- [x] aleph-bft's build tests
- [x] aleph-bft's unit-tests
- [x] newest aleph-node with these changes is able to finalize - checked for 2 sessions

PR #143, opened 2021-09-22
A0-427: Broadcast own unit after adding to dag

michalseweryn:
This PR introduces two changes:

First, it introduces broadcasting each unit created by us immediately after it is added to dag, and secondly, it resolves a missing coordinate request at the moment of adding it to a store instead of doing it at the moment when it is added to dag

### DIFF
--- a/src/runway.rs
+++ b/src/runway.rs
@@ -255,7 +255,7 @@ where
 
     // TODO: we should return an error and handle it outside
     fn validate_unit(
-        &mut self,
+        &self,
         uu: UncheckedSignedUnit<H, D, MK::Signature>,
     ) -> Option<SignedUnit<'a, H, D, MK>> {
         let su = match uu.check(self.keybox) {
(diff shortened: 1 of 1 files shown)


src/runway.rs
```diff
@@ -619,9 +617,12 @@ where
             NotificationOut::AddedToDag(h, p_hashes) => {
                 self.store.add_parents(h, p_hashes);
                 self.resolve_missing_parents(&h);
-                let coord = self.store.unit_by_hash(&h).map(|u| u.as_signable().coord());
-                if let Some(coord) = coord {
-                    self.resolve_missing_coord(&coord)
+                if let Some(su) = self.store.unit_by_hash(&h).cloned() {
+                    let coord = su.as_signable().coord();
```
timorl, line 621: This unit does not have a `creator` method?

michalseweryn: It does, nice catch!

DamianStraszak (approved):

fixxxedpoint (approved):
- [x] aleph-bft build tests
- [x] aleph-bft's unit-tests
- [x] newest aleph-node with these changes is able to finalize - checked for 2 sessions
- [ ] crashing and restarting aleph-node in order to trigger these changes

PR #152, opened 2021-11-10
AZ-62: Fix synchronization workflow with foundation repository

maciejnems:
Now the workflow has access to pushing to foundation repository and triggers only on push to Cardinal-Cryptography repo


timorl (approved):

fbielejec (approved):
I approve these changes

PR #156, opened 2021-11-19
Readme cosmetic update

h4nsu:
Changes logo to a dedicated AlephBFT logo and fixes some typos.


README.md
```diff
@@ -4,18 +4,16 @@
 [![Apache 2.0 Licensed][license-image]][license-link]
 ![Rust Stable][rustc-image]
 
-### Aleph Consensus
-
-![aleph logo](logo.jpg "Aleph logo")
+![aleph logo](https://alephzero.org/wp-content/uploads/A0_logotype_bft_dark.png)
```
pmikolajczyk41, line 7: [optional] maybe it would look better centered, but as described [here](https://stackoverflow.com/a/12118349) it would require html tags which is not so great within markdown

pmikolajczyk41 (approved):

README.md
h4nsu: I wanted to avoid html, but you're right, it looks meh without centering. Fixed!

mike1729 (approved):

PR #161, opened 2021-11-22
AZ-331: Synchronize repositories after bumping version

pmikolajczyk41:


DamianStraszak (approved):
LGTM, although I'm not an expert here :) Better wait for @maciejnems

pmikolajczyk41:
@maciejnems please check my test workflows here:
https://github.com/pmikolajczyk41/AlephBFT/actions/workflows/wf1.yml
https://github.com/pmikolajczyk41/AlephBFT/actions/workflows/wf2.yml

it should work as expected (after adding `ref` attribute to checkout)

maciejnems (approved):
Looks good now. I don't like the fact, that we use force push to update dependencies. Maybe we should add a new commit instead? It is not in the scope of this PR ofc.

PR #165, opened 2021-12-15
AZ 219 remove git force push to main

KazyCC:
Changes:
1) in push-foundation-repo.yml automatic publish to crates.io will be disabled for now.
2) removed force-push
3) version is updating on PR branch, not main
4) new version is established before completing the PR

### DIFF
--- a/.github/workflows/prepare-version-bump.yml
+++ b/.github/workflows/prepare-version-bump.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches:
     - main
+    - test_main
 
 jobs:
   prepare-new-version:
--- a/.github/workflows/prepare-version-bump.yml
+++ b/.github/workflows/prepare-version-bump.yml
@@ -19,6 +19,8 @@ jobs:
         uses: actions/checkout@v2
         with:
           fetch-depth: 2
+          ref: ${{ github.base_ref }}
+          token: ${{ secrets.SYNCAZF }}
       - name: check-and-bump
         run: |
           old_version=`cargo next --get`
--- a/.github/workflows/prepare-version-bump.yml
+++ b/.github/workflows/prepare-version-bump.yml
@@ -1,0 +1,58 @@
+name: Autobump version
+
+on:
+  pull_request:
+    branches:
+    - main
+
+jobs:
+  prepare-new-version:
+    environment: Autobump version
+    runs-on: ubuntu-latest
+    # if: ${{ github.repository == 'Cardinal-Cryptography/AlephBFT'}}
+    steps:
+      - name: cargo-next-install
+        run: |
+          cargo install cargo-next --locked
+      - name: checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+      - name: check-and-bump
+        run: |
+          old_version=`cargo next --get`
+          # If version on PR branch and main branch have no differences in version -> bump it.
+          # Set version in README.md, commit, tag and push to PR branch.
+          # Assuming nominal version for us is the one in the top level Cargo.toml. 
+          if ! git diff ${{ github.head_ref }} ${{ github.base_ref }} -- Cargo.toml | grep -q '^+version ='; then
+            echo 'Version $old_version has NOT been changed manually between ${{ github.head_ref }} and ${{ github.base_ref }}, patching it now.'
+            cargo next --patch
+            new_version=`cargo next --get`
+            echo 'New version is $new_version.'
+            git add Cargo.toml
+          fi
+          new_version=`cargo next --get`
+          # Make sure README has the same version as Cargo.toml.
+          sed -i "s/^version = \"$version\"$/version = \"$new_version\"/" Cargo.toml
+          git add README.md
+          # Publishing changes to PR branch
+          git config user.email "<>"
+          git config user.name "Version autobump"
+          git commit -m "Setting version to $new_version"
+          git tag -a $new_version -m "Version $new_version created at PR ${{ github.base_ref }} -> ${{ github.head_ref }}"
+      - name: push-to-PR
+        run: |
+          git push origin ${{ github.head_ref }}
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+      - name: credentials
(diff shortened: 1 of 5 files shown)


KazyCC:
It looks like there is an issue with permissions -  https://github.com/Cardinal-Cryptography/AlephBFT/actions/runs/1583186193? @bartoszjedrzejewski  could you assist?

.github/workflows/push-foundation-repo.yml
```diff
@@ -1,11 +1,16 @@
-name: Sync Cardinal-Cryptography repo with Aleph-Zero-Foundation repo
+name: Publish AlephBFT main to crates.io.
```
timorl, line 1: You should probably name this something that implies it also synchronizes the foundation repo.

.github/workflows/push-foundation-repo.yml
```diff
@@ -1,11 +1,16 @@
...
+  # NOTE: changed to "manual workflow dispatch" for now.
+  # TODO: To be automated again, when workflows stabilize
+  # and pushes to main origin will be forbidden.
+  workflow_dispatch:
+
+#  workflow_run:
+#    workflows: ["Autobump version"]
+#    branches: [main]
+#    types:
+#      - completed
```
timorl, line 13: Commented code.

.github/workflows/prepare-version-bump.yml
```diff
@@ -0,0 +1,50 @@
...
+            git add Cargo.toml
+          fi
+          new_version=`cargo next --get`
+          # Make sure README has the same version as Cargo.toml.
+          sed -i "s/^version = \"$version\"$/version = \"$new_version\"/" Cargo.toml
+          git add README.md
+          # Publishing changes to PR branch
+          git config user.email "<>"
+          git config user.name "Version autobump"
+          git commit -m "Setting version to $new_version"
```
timorl, line 43: What is the workflow for reviews now? Forcepushing to own branch always and this commit appears anew (would this work?)? Or rather pulling this commit and building on top of it?

.github/workflows/version-bump.yml
```diff
@@ -52,6 +21,6 @@ jobs:
           args: ${{ secrets.CRATES_IO_TOKEN }}
       - name: publish
         run: |
-          if [ -f publishMe ]; then
-            cargo publish
-          fi
+          echo 'Uploading to crates.io'
```
timorl, line 24: Uh, I'm confused now, this _always_ pushes to crates.io?

Anyway, you probably want to rename this file? It seems there was some naming confusion here.

timorl (changes requested):
A couple nits, some naming confusion and one question. Looks nice in general!

.github/workflows/push-foundation-repo.yml
KazyCC: By purpose - see note above.

.github/workflows/prepare-version-bump.yml
KazyCC: If version was not set manually, you will need a pull.

.github/workflows/version-bump.yml
KazyCC: This workflow is invoked right now only manually, so yes, if you invoke it manually, it will publish every time it will be called.

.github/workflows/push-foundation-repo.yml
KazyCC: Good point, corrected with dc09229.

timorl (changes requested):

.github/workflows/push-foundation-repo.yml
KazyCC: It's not about history - commented code is left there, because it is the right code to be enabled in the future, when workflows stabilize.

.github/workflows/version-bump.yml
KazyCC: Ok, I did some clean up. See 4339e93. Workflow should be like this:
1. On PR creation to main  -> prepare-version-bump.yml 
2. Manually (for now) -> version-bump.yml
3. On (2) -> push-foundation-repo.yml.

timorl: Yeah, that seems good now. I still have some gripes with naming – maybe call this one something like "Publish to crates.io" rather than "Autobump version"? Similar for file names, what is now called "prepare-..." should just be "version-bump" and this one "publish" or something.

.github/workflows/push-foundation-repo.yml
timorl: Yeah, but keeping commented code is always confusing. I would rather trust the person who will enable the automation here to do this correctly. If they need help one of the options is to look in the history, that's what I meant by it staying there.

.github/workflows/push-foundation-repo.yml
```diff
@@ -1,4 +1,4 @@
-name: Sync Cardinal-Cryptography repo with Aleph-Zero-Foundation repo
+name: Sync Cardinal-Cryptography repo with Aleph-Zero-Foundation repo.
```
timorl, line 1: Nit: All the other names don't have punctuation, please make this consistent (whichever way you prefer).

timorl (changes requested):
Still some minor gripes. ^^"

.github/workflows/prepare-version-bump.yml
```diff
@@ -0,0 +1,50 @@
...
+        run: |
+          cargo install cargo-next --locked
+      - name: checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+          ref: ${{ github.head_ref }}
+          token: ${{ secrets.SYNCAZF || secrets.MY_TOKEN || github.token }}
+      - name: check-and-bump
+        run: |
```
kostekIV, line 24: nit: this block can be made into sh script (and put somewhere in the `scripts/...` folder). This way it will be more readable here and easier to review the script without the noise from the workflow file.

Marcin-Radecki:
I am closing this PR on behalf we've incorporated its inputs in https://github.com/Cardinal-Cryptography/AlephBFT/pull/168 .

Thanks for work on this :)

PR #168, opened 2022-01-11
AZ-219 Replaced statefull force push check when pull requests are merged

Marcin-Radecki:
This PR replaces a stateful force push check when pull requests are merged with a stateless version bump check when a new commit is pushed to any PR with the target branch as `main`.

The motivation of this change is to remove the solution with `git push --force`, and yet still have most of its features:
* developer synchronization: with `git push --force` it is for CI to decide to bump the last version on `main` to the next one, so developers do not need to track if the version on `main` changed meanwhile sb merged other PR. With the proposed solution we don't have this particular gain, however, it should be rare people won't figure out the version changed meanwhile,
* automatic solution prevented people from forgetting to update the version - this still holds, you can't merge PR which has source code changes without version bumped in it,
* commit atomicity, ie we want ideally to have the version bumped in the same commit as the change due to the current way of publishing `AlephBFT` to crates.io - this still holds,
* additional gain to remove `git push --force` is we can enable `main` as protected branch in GitHub, and that prevents `git push --force` from outside as well as forces PR to be raised with given a configurable minimal amount of approvers (this works on our other `aleph-node` repo).

Once this PR is merged, the other one https://github.com/Cardinal-Cryptography/AlephBFT/pull/165 would be declined (thanks for that work which influenced this PR).

Testing done: 
* `.github/workflows/check-version-bumped.yml` was tested on the fork, see  https://github.com/Marcin-Radecki/AlephBFT/pull/3 and history of commits on that PR  
* `.github/workflows/push-foundation-repo.yml` change was **not** tested due to its nature,
* `.github/workflows/push-foundation-repo.yml`was **partially** tested, with modified version (https://github.com/Marcin-Radecki/AlephBFT/commit/ff9475b31332ac0e026056f26d3083fb971c7ccc, https://github.com/Marcin-Radecki/AlephBFT/commit/13a890dfd4a8a3f4c8a34674280f47b47c056c33): on PR merged https://github.com/Marcin-Radecki/AlephBFT/pull/4 this action was run https://github.com/Marcin-Radecki/AlephBFT/runs/4787358743?check_suite_focus=true


bartoszjedrzejewski:
What if PR changes only files in github workflows and doesnt req bump in version? As I see it, now it will req the version to be bumped anyway. At least the workflow will be run unnecessarily.

Marcin-Radecki:
> What if PR changes only files in github workflows and doesnt req bump in version? As I see it, now it will req the version to be bumped anyway. At least the workflow will be run unnecessarily.

If PR changes only GitHub workflows, so in particular it does not change any files in `src/` folder, then this check would be skipped. We don't bump versions automatically now, it is for a developer to do it (or not in given circumstances, as per above). Agree workflow would be run anyway, but it's rather cheap ('just' git clone and simple O(1) checks? And I assume workflows are run in parallel, so the CI workflow is longer anyway, as `clippy` takes more time than introduced version check workflow.

.github/scripts/check-cargo-toml-version-bumped.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -e
+
+if [ -z "$(git diff HEAD origin/main -- src/)" ]; then
+  echo "No changes in the code."
+  exit 0
+fi
+
+if [ -z "$(git diff HEAD origin/main -- Cargo.toml | grep '^+version =')" ]; then
```
timorl, line 10: You don't need the `[ -z` or making this into a string, `grep` returns an error if it didn't match, so just the two piped commands are enough as the if condition.

.github/workflows/check-version-bumped.yml
```diff
@@ -0,0 +1,21 @@
+name: Check if PR has bumbed versions in Cargo.toml for code changes and if README is synchronized
```
timorl, line 1: Seems a bit excessive for a workflow name. Perhaps change to "Check version bump" and explain in a comment what exactly it does.

I've been hoping there would be a `description` field better for this kind of information, but I couldn't find anything about this, so there probably isn't. :<

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,38 @@
+name: Publish to crates.io
+
+on:
+  push:
+    branches:
+      main
+
+jobs:
+  publish:
+    # need to rename environment name after this is merged
```
timorl, line 10: Why cannot you do this now?

timorl: Ohhh, secrets, silly me. >.<

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,38 @@
...
+    runs-on: ubuntu-latest
+    if: ${{ github.repository == 'Cardinal-Cryptography/AlephBFT'}}
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+      - name: force publish if version bumped
+        run: |
+          if [ -n "$(git diff HEAD~ -- Cargo.toml | grep '^+version =')" ]; then
```
timorl, line 21: Again, you can just use the `grep` return code.

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,38 @@
...
+            echo "Version in Cargo.toml was bumped in this PR, publishing to crates.io."
+            touch publishMe
```
timorl, line 23: I believe this trick is not necessary anymore, just do `credentials` always, and move the above `if` to `publish`.

It was needed, because of some silly dependencies between this and the forced version bump.

timorl (changes requested):
Nice, much better than forcepushing.

bartoszjedrzejewski:
> > What if PR changes only files in github workflows and doesnt req bump in version? As I see it, now it will req the version to be bumped anyway. At least the workflow will be run unnecessarily.
> 
> If PR changes only GitHub workflows, so in particular it does not change any files in `src/` folder, then this check would be skipped. We don't bump versions automatically now, it is for a developer to do it (or not in given circumstances, as per above). Agree workflow would be run anyway, but it's rather cheap ('just' git clone and simple O(1) checks? And I assume workflows are run in parallel, so the CI workflow is longer anyway, as `clippy` takes more time than introduced version check workflow.

ok, but why waste github actions seconds pool for nothing. You can use this: https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-ignoring-paths
It's not a nogo for me, but would be nice to have it implemented.

.github/scripts/check-cargo-toml-version-bumped.sh
Marcin-Radecki: Good point, I'll remove `-z`.

.github/workflows/check-version-bumped.yml
Marcin-Radecki: Good point, since this name is displayed in UI like below:

image

I'll shorten the name.

.github/workflows/publish-package.yml
Marcin-Radecki: Will change.

.github/workflows/publish-package.yml
Marcin-Radecki: Right, I have not noticed that this can be simplified. Will change.

Marcin-Radecki:
> > > What if PR changes only files in github workflows and doesnt req bump in version? As I see it, now it will req the version to be bumped anyway. At least the workflow will be run unnecessarily.
> > 
> > 
> > If PR changes only GitHub workflows, so in particular it does not change any files in `src/` folder, then this check would be skipped. We don't bump versions automatically now, it is for a developer to do it (or not in given circumstances, as per above). Agree workflow would be run anyway, but it's rather cheap ('just' git clone and simple O(1) checks? And I assume workflows are run in parallel, so the CI workflow is longer anyway, as `clippy` takes more time than introduced version check workflow.
> 
> ok, but why waste github actions seconds pool for nothing. You can use this: https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-ignoring-paths It's not a nogo for me, but would be nice to have it implemented.

Let me try with those paths, agreed if we can make it work so workflow are not spawned for nothing then it's better.

kostekIV (approved):
lgtm :)

.github/workflows/publish-package.yml
Marcin-Radecki: This needs to be changed in the future as now name of environment stands out - I'll try to follow up with repo admin

Marcin-Radecki:
I've applied review remarks and it's better now, apart from cleaner bash  we don't waste Guthub resources, and it's more clean to developer than checks are run only if they modify particular files. 

Testing done:
* https://github.com/Marcin-Radecki/AlephBFT/pull/5
* https://github.com/Marcin-Radecki/AlephBFT/pull/6

timorl (approved):
<3

Marcin-Radecki:
Everything worked as expected after merge:
* Code was synced to AlephFoundation: https://github.com/Cardinal-Cryptography/AlephBFT/runs/4801326323?check_suite_focus=true
* Create was uploaded to crates.io: https://github.com/Cardinal-Cryptography/AlephBFT/runs/4801326331?check_suite_focus=true

PR #170, opened 2022-01-13
AZ-447: Make member die when killed, even early

timorl:
members-die-when-killed

### DIFF
--- a/src/creation/mod.rs
+++ b/src/creation/mod.rs
@@ -92,7 +92,7 @@ async fn create_unit<H: Hasher>(
 pub async fn run<H: Hasher>(
     conf: Config,
     io: IO<H>,
-    starting_round: oneshot::Receiver<Round>,
+    mut starting_round: oneshot::Receiver<Round>,
     mut exit: oneshot::Receiver<()>,
 ) {
     let Config {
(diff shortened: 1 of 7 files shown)


src/creation/mod.rs
```diff
@@ -106,12 +106,15 @@ pub async fn run<H: Hasher>(
...
+        round =  starting_round => match round {
+            Ok(round) => round,
+            Err(e) => {
+                error!(target: "AlephBFT-creator", "Starting round not provided: {}", e);
+                return;
+            }
+        },
+        _ = &mut exit => return,
     };
     debug!(target: "AlephBFT-creator", "Creator starting from round {}", starting_round);
```
Marcin-Radecki, line 119: Now sure why my IDE says type of `starting_round` at this line is `()`. Had it been `()`, it would not have compiled. On the other hand I'm a bit confused here, since we do assign to `starting_round` result of `select!`, which indeed returns `()`, as all its branches must return the same, and second branch return type is `()`

image

Marcin-Radecki: Ah my bad, `return` has _never_ type or `!` : https://doc.rust-lang.org/nightly/std/primitive.never.html
All good then. 

>! will coerce into any other type.

Marcin-Radecki (approved):
Change looks good.

src/creation/mod.rs
timorl: The second branch return type is actually `bot`, since it returns out of the function completely.

Your IDE might be confused by the `futures` macro? I know it used to be a problem with my vim plugin, but it got better.

src/creation/mod.rs
```diff
@@ -106,12 +106,15 @@ pub async fn run<H: Hasher>(
...
     } = io;
-    let starting_round = match starting_round.await {
-        Ok(round) => round,
-        Err(e) => {
-            error!(target: "AlephBFT-creator", "Starting round not provided: {}", e);
-            return;
-        }
+    let starting_round = futures::select! {
+        round =  starting_round => match round {
+            Ok(round) => round,
```
h4nsu, line 111: Nit: These is a bit tough to read at first due to multiple occurences of `round` in this select block. I'd personally rename one of these, but up to you.

h4nsu (approved):
LGTM ;)

PR #184, opened 2022-03-25
AZ-686: Factor validation out of runway

timorl:


consensus/src/runway/mod.rs
```diff
@@ -238,23 +293,25 @@ where
                     self.on_parents_response(u_hash, parents)
                 }
                 Response::NewestUnit(response) => {
-                    let salt = response.as_signable().salt;
-                    trace!(target: "AlephBFT-runway", "{:?} Response parents received {:?}.", self.index(), salt);
+                    trace!(target: "AlephBFT-runway", "{:?} Response parents received.", self.index());
```
maciejnems, line 296: I guess it should be "Response newest unit received"? Also maybe we should print `response.index()` ?

consensus/src/runway/mod.rs
```diff
@@ -264,40 +321,6 @@ where
...
-            // NOTE: this implies malicious behavior as the unit's session_id
-            // is incompatible with session_id of the message it arrived in.
-            warn!(target: "AlephBFT-runway", "{:?} A unit with incorrect session_id! {:?}", self.index(), full_unit);
-            return None;
-        }
-        if full_unit.round() > self.store.limit_per_node() {
-            warn!(target: "AlephBFT-runway", "{:?} A unit with too high round {}! {:?}", self.index(), full_unit.round(), full_unit);
-            return None;
-        }
-        if full_unit.creator().0 >= self.n_members.0 {
```
maciejnems, line 290: I guess this check was removed because `uu.check(self.keybox)` already checks per index?

consensus/src/units/validator.rs
```diff
@@ -0,0 +1,337 @@
...
+            return Err(ValidationError::NotDescendantOfPreviousUnit(
+                pre_unit.clone(),
+            ));
+        }
+        Ok(su)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
maciejnems, line 132: I guess there are no checks for `WrongSignature`, `RoundZeroWithParents` and `NotDescendantOfPreviousUnit`

maciejnems (approved):
LGTM. Just some small comments

consensus/src/runway/mod.rs
timorl: Yup, exactly. Good catch though, I wasn't sure anyone would notice.

consensus/src/units/validator.rs
timorl: Unfortunately. We cannot create wrong signatures with the current version of mocks (and I didn't want to modify them significantly before they get factored out), and we cannot use `Creator` to create the other scenarios. Since there were no tests at all before this refactor I hope I can sneak by with subpar coverage. <_<

consensus/src/runway/mod.rs
timorl: Better now I hope.

consensus/src/runway/mod.rs
DamianStraszak, line 290: Are we relying here on the fact that the underlying implementation of `KeyBox` in `uu.check(...)` is correctly dealing with incorrect node indices? 

No matter whether this is the case I think it's way better to have an explicit error variant `IncorrectNodeIndex` instead of relying on the signature check.

DamianStraszak (changes requested):
Good work. I don't have extensive comments, only some nitpicks.

consensus/src/units/validator.rs
```diff
@@ -0,0 +1,337 @@
...
+    NotDescendantOfPreviousUnit(PreUnit<H>),
+}
+
+impl<H: Hasher, D: Data, S: Signature> Display for ValidationError<H, D, S> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
+        use ValidationError::*;
+        match self {
+            WrongSignature(usu) => write!(f, "wrongly signed unit: {:?}", usu),
+            WrongSession(fu) => write!(f, "unit from wrong session: {:?}", fu),
+            RoundTooHigh(fu) => write!(f, "unit with too high round{}: {:?}", fu.round(), fu),
```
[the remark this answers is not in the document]
timorl: Fixed.

consensus/src/units/mod.rs
```diff
@@ -5,8 +5,14 @@ use crate::{
 };
 use codec::{Decode, Encode};
 use derivative::Derivative;
+use parking_lot::RwLock;
```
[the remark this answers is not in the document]
timorl: No idea, but it seems extremely old. The only thing that uses it is `FullUnit`, and it seems to be using it in place of a `Cell`. If you think this won't muddle the PR too much I think I can replace it with a `Cell`, shouldn't be much work.

consensus/src/runway/mod.rs
```diff
@@ -238,23 +293,25 @@ where
                     self.on_parents_response(u_hash, parents)
                 }
                 Response::NewestUnit(response) => {
-                    let salt = response.as_signable().salt;
-                    trace!(target: "AlephBFT-runway", "{:?} Response parents received {:?}.", self.index(), salt);
+                    trace!(target: "AlephBFT-runway", "{:?} Response newest unit received from {:?}.", self.index(), response.index());
```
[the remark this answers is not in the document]
timorl: The next PR will print errors on mismatched salts, so I don't think this will add much value then. Do you think I should still add this, or is this promise assurance enough?

consensus/src/runway/mod.rs
timorl: Yes, that's what I am relying on. I removed it when I noticed there is no way of testing this, since it cannot happen by design. Because of that I don't think we should put it back.

consensus/src/units/mod.rs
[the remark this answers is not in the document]
DamianStraszak: Better to leave it -- maybe there are some hidden problems. I created a task for that https://cardinal-cryptography.atlassian.net/browse/AZ-736

consensus/src/runway/mod.rs
DamianStraszak: I don't get it -- what "cannot happen by design"? 

I think that the `IncorrectNodeIndex` is semantically very different to `WrongSignature` and it is not quite intuitive for the latter to capture the former.

DamianStraszak: And what I would suggest is to move the check for correctness of the `NodeIndex` before the signature check, because without this there is indeed no point.

timorl: But it always will, so this part of the code will never be reached. Dead code ain't good.

DamianStraszak: Ok, I see, the problem is that the whole `Signed` design is not allowing us to reach the NodeIndex before validating the signature?

timorl: Ah, async communication, I haven't seen your other comment before writing my own so my response was a non sequitur, classic.

We could actually check it in this order, `UncheckedSigned` gets you access to the signed object. It is only a) doubled with the signature check and b) out of order with all the other checks. Currently, we first check the signature and only afterwards the shape of the unit, going from outside in. Doing this one check out of order to check something that will be re-checked in a moment anyway seems wrong.

Perhaps it would be nice to make this a more explicit part of the `KeyBox` interface, maybe even with different errors returned from `check`ing. `KeyBox` could also keep and make accessible the information about the node count, which would simplify some other parts of the code. It already has to have that information implicitly, so explicitly requiring it would solve this problem. What do you think about this idea? (Not as part of this PR, but as a task for the future.)

DamianStraszak: Yeah, the proposal sounds good. But one thing that I think begs to be done is to at least write in the doc comment for the `KeyBox` trait that the implementation is expected to return `false` when the index is incorrect.

DamianStraszak (approved):

PR #189, opened 2022-03-30
AZ-686: Factor initial unit collection out of runway

timorl:


mike1729 (changes requested):

consensus/src/runway/collection.rs
```diff
@@ -0,0 +1,519 @@
...
+    pub fn starting_round(&self) -> Option<Round> {
+        match self.ready() {
+            true => Some(self.starting_round),
+            false => None,
+        }
+    }
+}
+
+/// A runnable wrapper around initial unit collection.
+pub struct IO<'a, H: Hasher, D: Data, MK: KeyBox> {
```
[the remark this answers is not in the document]
timorl: Good point, I did it mostly for consistency with other modules, but it probably doesn't make much sense here. I'll change this to a method, unless the other discussion ends up making this struct more involved.

consensus/src/runway/collection.rs
```diff
@@ -0,0 +1,519 @@
...
+            round_for_creator,
+            responses_from_network,
+            resolved_requests,
+            collection,
+        }
+    }
+
+    /// Run the initial unit collection until it sends the initial round.
+    pub async fn run(mut self) {
+        let mut catch_up_delay = futures_timer::Delay::new(Duration::from_secs(5)).fuse();
```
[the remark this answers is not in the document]
timorl: `on_newest_response` is supposed to be non-`async` to keep the main collection struct easily unittestable. So it cannot depend on the timeout.

I think I could glue together the loops in the way you suggest by adding a `ready` field in the IO struct and making the `delay` an option, or something of that nature. I'm not convinced this is actually much better than the current solution though, so please let me know what you think – including whether then keeping the `IO` struct is reasonable.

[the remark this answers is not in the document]
mike1729: Why would `on_newest_response` have to depend on the timeout?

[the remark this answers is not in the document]
timorl: If you want it to return three different states then it would have to? Or perhaps I misunderstood your proposition?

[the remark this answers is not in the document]
timorl: Restructured in one loop and added the `Finished` optimization.

consensus/src/runway/collection.rs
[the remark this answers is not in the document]
timorl: After the below changes there is another method there which maybe justifies the existence of the struct.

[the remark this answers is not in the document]
mike1729: Agreed.

consensus/src/runway/collection.rs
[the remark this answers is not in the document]
mike1729: Beautiful :heart:

mike1729 (approved):

consensus/src/runway/collection.rs
```diff
@@ -0,0 +1,536 @@
...
+
+    /// The salt associated with this collection instance.
+    pub fn salt(&self) -> Salt {
+        self.salt
+    }
+
+    /// The current status of the collection.
+    pub fn status(&self) -> Status {
+        use Status::*;
+        let responders = NodeCount(self.responders.len());
```
maciejnems, line 198: Small nit: In two ifs here you compare in different way (one by creating `NodeCount`, second by taking `.0`). Maybe here could be something like (`NodeCount` has `PartialOrd` derived) 
```
let responder_count = NodeCount(self.responders.len());
if responder_count == self.keychain.node_count() {
    return Finished(self.starting_round);
}
if responder_count >= self.threshold {
    return Ready(self.starting_round);
}
```

consensus/src/runway/collection.rs
```diff
@@ -0,0 +1,536 @@
...
+        Error::WrongSignature
+    }
+}
+
+/// The status of an ongoing collection.
+#[derive(PartialEq, Debug)]
+pub enum Status {
+    /// Received less than threshold responses.
+    Pending,
+    /// Received at least threshold responses.
```
maciejnems, line 126: Nit: This comment is not quite correct. It should be "Received at least `threshold - 1` responses" (unless we count ourself as a response) similar is with `Status::Pending`. So maybe we should change those comments or remove adding `keychain.index()` to responders?

consensus/src/runway/collection.rs
```diff
@@ -0,0 +1,536 @@
...
+        if response.salt != self.salt {
+            return Err(Error::SaltMismatch(self.salt, response.salt));
+        }
+        if let Some(unchecked_unit) = response.unit {
+            let checked_signed_unit = self.validator.validate_unit(unchecked_unit)?;
+            let checked_unit = checked_signed_unit.as_signable();
+            if checked_unit.creator() != self.keychain.index() {
+                return Err(Error::ForeignUnit(checked_unit.creator()));
+            }
+            let starting_round_candidate = checked_unit.round() + 1;
```
maciejnems, line 181: Nit: Could be 
```
self.starting_round = std::cmp::max(self.starting_round, checked_unit.round() + 1);
```
but I'm not sure if it is popular to use max in rust.

consensus/src/runway/collection.rs
```diff
@@ -0,0 +1,536 @@
...
+                        return;
+                    },
+                },
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
maciejnems, line 290: Maybe we could add a test for too few messages but some messages are sent more than once?

maciejnems (commented):
Beautiful out factor, just some minor comments

consensus/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aleph-bft"
-version = "0.10.0"
+version = "0.10.1"
```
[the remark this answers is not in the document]
timorl: I think bumping the version on anything that changes the external interface is a must, and we should probably also bump it when the final version is implemented.

Anyway, changed to "0.11.0", since you are right about the feature flag.

consensus/src/runway/collection.rs
timorl: Yup, you are completely right, no idea why I thought I had to touch the `.0`s. Fixed.

consensus/src/runway/collection.rs
timorl: Mentioned counting the trivial self-response.

consensus/src/runway/collection.rs
timorl: Eh, I dunno, I think we were trying to avoid that to be more explicit, although maybe this is a strange leftover from the Go days...

timorl: Do you think this would be more readable?

consensus/src/runway/collection.rs
timorl: Good idea, added.

consensus/src/runway/collection.rs
maciejnems: I usually prefer using `max` as it is more readable to me. But I'm completely fine with current approach it's a very small nit and I'm not sure how readable it is for others.

maciejnems (approved):

consensus/src/runway/collection.rs
timorl: Added then.

PR #190, opened 2022-03-30
AZ-687: Add Unit read/write interface

maciejnems:
This PR is one of steps to add node crash recovery by using HDD. It:
* adds `std::io::Write` and `std::io::Read` arguments


consensus/src/testing/mock.rs
```diff
@@ -494,6 +494,98 @@ impl DataProvider {
...
 
+type SavedUnit = Vec<u8>;
+
+struct UnitSaver {
+    rx: UnboundedReceiver<SavedUnit>,
+    units: Arc<Mutex<Vec<SavedUnit>>>,
+}
+
+impl UnitSaver {
+    async fn run(mut self) {
```
deuszx, line 505: Personally, I would feel more comfortable with these interfaces being synchronous. It is crucial that we don't continue with the unit before it's stored and we don't start creating the new ones until we've loaded all from the disk on startup.

If we're worried about the performance then we're really saying that persisting X bytes on disk might take longer than sending them over the wire (what we will be sending over the wire will be more in practice b/c of all the message envelopes, etc.).

maciejnems: But it is a mock for tests. It has nothing in common with actual implementation of this saver. I'm not sure if this is the place where you wanted to add this comment.

deuszx: Ah yes, wrong place but the idea is the same.

Yes, I read the docs, it's just that making this interface non-async makes it more explicit that we have to wait. Even if you called `save.await` we don't gain anything from making this async as we want to always wait anyway.

types/src/units.rs
```diff
@@ -0,0 +1,114 @@
...
+///     }
+/// }
+///
+/// impl IntoBackupImpl {
+///     fn new(tx: UnboundedSender<Vec<u8>>) -> IntoBackupImpl {
+///         IntoBackupImpl { tx }
+///     }
+/// }
+/// ```
+pub trait IntoBackup {
```
timorl, line 53: Do you even need this trait? Couldn't you `impl<T: ..., B: Backup<T>> Into<B> for IntoBackupImpl`?

Same for the reader ofc.

types/src/units.rs
```diff
@@ -0,0 +1,114 @@
...
+    type Into;
+
+    /// Consume the object and return struct `Backup` used for saving elements of type T
+    fn into_backup<T: Codec + Send + Sync + 'static>(self) -> Self::Into
+    where
+        Self::Into: Backup<T>;
+}
+
+/// A type to generate a structure for reading data that was backed up.
+/// Example implementation:
```
timorl, line 64: I'm pretty sure this is a mock implementation, especially with the double `Vec`, so I'm not sure you want to write it as an example here.

If you want a proper example you would have to write out a bytes buffer and slow reading/decoding from it, not sure if it's worth it.

timorl (changes requested):
Main comment is in the middle of everything – I think the exact shape of this interface needs some more discussion.

And sorry for suggesting something like this in the first place if it turns out we don't really want to go for it. Dx (If we go for this version then I'm obviously not sorry. ;p)

consensus/src/runway/mod.rs
```diff
@@ -588,15 +589,26 @@ where
         }
     }
 
+    fn backup_data(
```
timorl: I think runway should already be getting an abstracted unit saver, implemented by a wrapper applied by member to the `Write`. Same for `Read` ofc.

I also think this is not the PR to do this in, which is why I suggested adding `_` in front of the arguments – although I probably should have suggested this in member already.

docs/src/aleph_bft_api.md
```diff
@@ -67,9 +67,43 @@ pub trait KeyBox: Index + Clone + Send {
...
 
+#### 3.1.3 IntoBackup, IntoReader.
+
+The `IntoBackup` and `IntoReader` traits are used for creating backups of Units created in a session. This is a part of crash recovery. Units created are needed for member to recover after crash during a session for Aleph to be BFT. This means that user needs to provide two traits `IntoBackup` and `IntoReader` that are used for storing and reading Unit that are created by member. At first (without any crash) `IntoReader` should return nothing. After crash it should contain all data that was stored before in this session.
+
+`IntoBackup` trait upon being consumed by `into_backup` method should return a struct implementing `Backup<T>` trait. It should implement for all `T: Codec + Send + Sync` a way of buckuping a data generated during session. **If member crashed `Backup<T>` should append data that was saved before the crash.**
```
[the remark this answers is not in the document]
maciejnems: Is it that hard? It should append to the data that was passed in `Read`. It does not sound that hard if we already have this data somewhere. Unless you mean that it's not so easy, to have it saved?

consensus/src/runway/mod.rs
```diff
@@ -246,6 +256,8 @@ where
             rx_consensus: config.rx_consensus,
             ordered_batch_rx: config.ordered_batch_rx,
             data_provider: config.data_provider,
+            unit_backup: config.unit_backup,
+            unit_reader: config.unit_reader,
```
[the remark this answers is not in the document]
maciejnems: Ok, I will add it in next PR

docs/src/aleph_bft_api.md
```diff
@@ -67,43 +67,32 @@ pub trait KeyBox: Index + Clone + Send {
 
 A typical implementation of KeyBox would be a collection of `N` public keys, an index `i` and a single private key corresponding to the public key number `i`. The meaning of `sign` is then to produce a signature using the given private key, and `verify(msg, s, j)` is to verify whether the signature `s` under the message `msg` is correct with respect to the public key of the `j`th node.
 
-#### 3.1.3 IntoBackup, IntoReader.
+#### 3.1.3 Read, Write
```
timorl: I would title this to make it more explicit that it's about intra-session recovery, perhaps "Read, Write – recovering mid session crashes".

I would also add a paragraph that this is optional if you don't want to recover crashes mid session, and if your session handling ensures AlephBFT will not run in the same session twice you can pass noop implementations here.

Preferably there should also be better documentation about turning initial unit collection off in this case, but that probably shouldn't be in this PR. >.<

timorl (approved):
Looks broadly correct, some comments nonetheless.

docs/src/aleph_bft_api.md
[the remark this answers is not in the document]
timorl: Depending on what you mean by "crash", if the whole machine crashes you cannot really guarantee anything. But I guess this is kinda clear from context...?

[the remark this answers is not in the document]
maciejnems: Yep, I think of it as if you have something saved, you should append it. It's not it should append all data generated before crash. So maybe "should append data that was saved despite the crash"?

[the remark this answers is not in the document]
timorl: Hmm, if I understand correctly what property you are trying to ensure perhaps something like "writing bytes should block until it's saved in a way that should be persistent, e.g. on a disk"? I think this is the actual property we want, since it ensures (as much as is practically possible) that stuff is saved before it's send to network.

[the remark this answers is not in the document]
maciejnems: Hmmm, about this one I thought a little bit differently. We ask to implement `Write` so it has two methods. `write` and `flush`. So I guess the user should implement it in a typical way and we use flush to make sure that all is saved

docs/src/aleph_bft_api.md
```diff
@@ -67,43 +67,32 @@ pub trait KeyBox: Index + Clone + Send {
...
-pub trait IntoBackup {
-    type Into;
+pub trait Write {
+    fn write(&mut self, buf: &[u8]) -> Result<usize>;
+    fn flush(&mut self) -> Result<()>;
 
-    fn into_backup<T: Codec + Send + Sync + 'static>(self) -> Self::Into
-    where
-        Self::Into: Backup<T>;
+	...
```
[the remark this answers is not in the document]
maciejnems: I thought this would emphasize the importance of implementing `flush`, but I can remove them.

docs/src/aleph_bft_api.md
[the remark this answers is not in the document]
timorl: `Write` doesn't necessarily write to disk on `flush` (it's implemented for many interfaces and they can do pretty arbitrary things) so the actual requirement should be that `flush` blocks until the written stuff is persistently saved.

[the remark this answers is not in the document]
maciejnems: How do you think we can emphasize this requirement? Do you think mentioning it here is enough?

consensus/src/member.rs
```diff
@@ -396,10 +397,13 @@ where
 /// For a detailed description of the consensus implemented by `run_session` see
 /// [docs for devs](https://cardinal-cryptography.github.io/AlephBFT/index.html)
 /// or the [original paper](https://arxiv.org/abs/1908.05156).
+#[allow(clippy::too_many_arguments)]
```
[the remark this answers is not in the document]
maciejnems: I added something that will contain all local IO (so all IO without Network) and named it `LocalIO`. It is something like:
```
#[derive(Clone)]
pub struct LocalIO<D: Data, DP: DataProvider<D>, FH: FinalizationHandler<D>, UB: Write, UR: Read> {
    pub data_provider: DP,
    pub unit_backup: UB,
    pub unit_reader: UR,
    pub finalization_handler: FH,
    _phantom: PhantomData<D>,
}
```
So `Member::run_session(...)` takes such arguments
```
    config: Config,
    local_io: LocalIO<D, DP, FH, UB, UR>,
    network: N,
    keybox: MK,
    spawn_handle: SH,
    mut exit: oneshot::Receiver<()>,
```
I guess similar solution would be to add `Network` to it and have just `IO` containing those things. One thing worth mentioning is that `runway::run(...)` does not take `Network` (which it should not take), but takes all other thing. So in case it was `IO` we would have to somehow parameterize it by `Network`, or create another config for runway, not to have too_many_arguments there. What do you think about it @fixxxedpoint ?

docs/src/aleph_bft_api.md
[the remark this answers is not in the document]
timorl: Yeah, I think just mentioning it here is enough, especially since this is kinda the default implementation in many cases.

consensus/src/config.rs
```diff
@@ -34,6 +37,34 @@ pub struct Config {
     pub max_round: Round,
 }
 
+#[derive(Clone)]
+pub struct LocalIO<D: Data, DP: DataProvider<D>, FH: FinalizationHandler<D>, UB: Write, UR: Read> {
```
timorl, line 41: I think this should rather be in `member.rs`. The fields also wouldn't have to be `pub`, which is nice.

consensus/src/config.rs
```diff
@@ -34,6 +37,34 @@ pub struct Config {
...
+    pub data_provider: DP,
+    pub unit_backup: UB,
+    pub unit_reader: UR,
+    pub finalization_handler: FH,
```
timorl, line 45: Nit: it would be nice to have this slightly better grouped, provider and finalization handler could be next to each other.

timorl (approved):

docs/src/aleph_bft_api.md
timorl: I probably wasn't clear – I think the paragraph about this being optional should be in this PR.

consensus/src/config.rs
lesniak43: Is the user supposed to provide their implementation of `UB` and `UR`? If so, then this fact should be noted in the `aleph-bft-types` package, possibly by moving this whole struct there?

timorl: The types are fortunately from `std`, so I don't think that is necessary.

lesniak43: on second thought (after a short chat with @timorl ), maybe not...

docs/src/aleph_bft_api.md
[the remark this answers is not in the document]
timorl: I think the second bold sentence is now redundant.

consensus/src/config.rs
timorl: Pff, sry, not args ofc, but the fields of `LocalIO`.

maciejnems: done

fixxxedpoint (approved):
lgtm

PR #193, opened 2022-04-11
A0-539: Add more tests for the alerter (fka AZ-340)

artemis-cc:
[PR 186](https://github.com/Cardinal-Cryptography/AlephBFT/pull/186) promised that tests would be added later. These are those tests.

### DIFF
--- a/consensus/src/alerts/mod.rs
+++ b/consensus/src/alerts/mod.rs
@@ -559,7 +559,7 @@ mod tests {
         let forker_index = NodeIndex(6);
         let own_keychain = Keychain::new(n_members, own_index);
         let forker_keychain = Keychain::new(n_members, forker_index);
-        let mut own_node = Alerter::new(
+        let mut this = Alerter::new(
             &own_keychain,
             AlertConfig {
                 n_members,
@@ -568,12 +568,12 @@ mod tests {
         );
         let fork_proof = make_fork_proof(forker_index, &forker_keychain, 0, n_members).await;
         let alert = Alert::new(own_index, fork_proof, vec![]);
-        let signed_alert = Signed::sign(alert.clone(), own_node.keychain)
+        let signed_alert = Signed::sign(alert.clone(), this.keychain)
             .await
             .into_unchecked();
         let alert_hash = Signable::hash(&alert);
         assert_eq!(
-            own_node.on_own_alert(alert).await,
+            this.on_own_alert(alert).await,
             (
                 AlertMessage::ForkAlert(signed_alert),
                 Recipient::Everyone,
@@ -589,7 +589,7 @@ mod tests {
         let forker_index = NodeIndex(6);
         let own_keychain = Keychain::new(n_members, own_index);
         let forker_keychain = Keychain::new(n_members, forker_index);
-        let mut own_node = Alerter::new(
+        let mut this = Alerter::new(
             &own_keychain,
             AlertConfig {
                 n_members,
@@ -599,11 +599,9 @@ mod tests {
         let fork_proof = make_fork_proof(forker_index, &forker_keychain, 0, n_members).await;
         let alert = Alert::new(own_index, fork_proof.clone(), vec![]);
         let alert_hash = Signable::hash(&alert);
-        let signed_alert = Signed::sign(alert, own_node.keychain)
-            .await
-            .into_unchecked();
+        let signed_alert = Signed::sign(alert, this.keychain).await.into_unchecked();
         assert_eq!(
-            own_node.on_network_alert(signed_alert).await,
+            this.on_network_alert(signed_alert).await,
             Some((Some(ForkingNotification::Forker(fork_proof)), alert_hash)),
         );
     }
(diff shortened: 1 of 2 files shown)


consensus/src/alerts/mod.rs
```diff
@@ -490,3 +490,367 @@ pub(crate) async fn run<H: Hasher, D: Data, MK: MultiKeychain>(
...
+            .check(&key_boxes[double_committer.0])
+            .expect("the signature is correct")
+            .into_partially_multisigned(&key_boxes[double_committer.0]);
+        for i in 1..n_members.0 - 2 {
+            let node_id = NodeIndex(i);
+            let signed_nonempty_alert_hash =
+                Signed::sign_with_index(nonempty_alert_hash, &key_boxes[node_id.0])
+                    .await
+                    .into_unchecked();
+            multisigned_nonempty_alert_hash = multisigned_nonempty_alert_hash.add_signature(
```
timorl, line 759: Something went really wrong with this test, you meticulously create this multisignature and then don't use it in any way in the test.

I think you need to use this multisignature to create an RMC message and check whether that correctly responds with a `AlerterResponse::RmcMessage`. You should probably also check that a not fully multisigned RMC message of this nonempty alert gets ignored, and a not fully multisigned RMC message of the empty alert does not get ignored (at least before the alerter receives the multisigned version, it could theoretically ignore it afterwards although the current implementation doesn't).

consensus/src/alerts/mod.rs
```diff
@@ -490,3 +490,367 @@ pub(crate) async fn run<H: Hasher, D: Data, MK: MultiKeychain>(
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
```
timorl, line 495: You don't test `alert_confirmed` at all, even though `run` uses it directly.

timorl (changes requested):
One request for more comments and two for slight test improvements, other than that nice.

consensus/src/alerts/mod.rs
DamianStraszak: Yes, I feel like @artemis-cc  tried to do two different things at once in this test.

If I understand correctly this was supposed to be a test case where we learn about two alerts `a1`, `a2`, with the same `alert_id` but the second one `a2` (the one we received later) then gets multisigned. In such a case, upon getting a `complete` rmc message for `a2` we should output `AlerterResponse::RmcMessage(message)`. There seems to be a few lines missing then if this was the intention.

consensus/src/alerts/mod.rs
```diff
@@ -490,3 +490,367 @@ pub(crate) async fn run<H: Hasher, D: Data, MK: MultiKeychain>(
...
+    }
+
+    #[tokio::test]
+    async fn who_is_forking_ok() {
+        let n_members = NodeCount(7);
+        let own_index = NodeIndex(0);
+        let forker_index = NodeIndex(6);
+        let own_key_box = KeyBox::new(n_members, own_index);
+        let forker_key_box = KeyBox::new(n_members, forker_index);
+        let own_node = Alerter::new(
```
DamianStraszak, line 781: Nit: we typically use the name "node" for a larger thing consisting of multiple components: `creator`, `terminal`, `alerter` and few others. It would be better to call it `own_alerter` or `alerter`.

DamianStraszak (approved):
Tests look solid and clean. I'm ok with merging after the suggestions of @timorl are addressed.

consensus/src/alerts/mod.rs
```diff
@@ -127,11 +127,11 @@ impl<H: Hasher, D: Data, S: Signature, MS: PartialMultisignature> AlertMessage<H
 ///
 /// Certain calls to the alerter may generate responses. It is the caller's responsibility to
 /// forward them appropriately.
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq)]
 enum AlerterResponse<H: Hasher, D: Data, S: Signature, MS: PartialMultisignature> {
     ForkAlert(UncheckedSigned<Alert<H, D, S>, S>, Recipient),
     ForkResponse(Option<ForkingNotification<H, D, S>>, H::Hash),
-    AlertRequest(AlertMessage<H, D, S, MS>, Recipient),
+    AlertRequest(H::Hash, Recipient),
```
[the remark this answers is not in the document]
artemis-cc: Added in ed5fb3fa!

consensus/src/alerts/mod.rs
artemis-cc: Added in 5d4be267 (thanks to @DamianStraszak for the help)!

consensus/src/alerts/mod.rs
artemis-cc: Yeah, that makes sense. I renamed it from `alerter` because some tests have a different mocked-up node that sends an alert and is called…alerter. So the thing that the alert comes from is not always the thing that I called “own_node”. I was confused by that for a while.

What about `this`? It kind of fits since the alerter is what's being tested

timorl (approved):

consensus/src/alerts/mod.rs
DamianStraszak: Yes, it's fine. Feel free to resolve this comment by yourself, I don't have strong opinions about names, just wanted to point that there is this specific meaning of "node" :)

DamianStraszak (approved):

consensus/src/alerts/mod.rs
artemis-cc: Fixed in 27fcea1e! 😀

PR #194, opened 2022-04-12
AZ-707 Push packages to crates.io as a part of CI 

lesniak43:
[A0-532](https://cardinal-cryptography.atlassian.net/browse/A0-532)

### DIFF
--- a/.github/workflows/publish-packages.yml
+++ b/.github/workflows/publish-packages.yml
@@ -5,7 +5,6 @@ on:
     branches:
       - main
 
-
 jobs:
   publish:
     environment: Autobump version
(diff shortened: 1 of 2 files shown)


timorl (approved):

.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,20 @@
+name: Publish to crates.io
+
+on:
+  push:
+    branches:
+      - main
+
+
```
kostekIV, line 8: usually, we have just one line between sections

kostekIV (approved):

.github/workflows/publish-packages.yml
lesniak43: fixed

PR #195, opened 2022-04-14
A0-543: Add backup loading mechanism

maciejnems:
This is a first part of adding backup loading mechanism. It introduces an implementation of backup loading mechanism that is yet to be integrated.

The idea is that backup mechanism reads Unit data from a file. After that it waits for response from unit collection. Upon receiving unit collection result it check highest round that was backed up. If unit collection had lower result, the backup mechanism sends all units to add them to dag and after that sends the starting round to creator.

### DIFF
--- a/consensus/src/runway/backup.rs
+++ b/consensus/src/runway/backup.rs
@@ -5,6 +5,7 @@ use log::{error, warn};
 use std::io::{Read, Write};
 
 #[derive(Debug)]
+/// Backup load error. Could be either caused by io error from Reader, or by decoding.
 pub enum LoaderError {
     IO(std::io::Error),
     Codec(CodecError),
@@ -22,10 +23,12 @@ impl From<CodecError> for LoaderError {
     }
 }
 
+/// Abstraction over Unit backup saving mechanism
 pub trait UnitSaver<H: Hasher, D: Data, S: Signature> {
     fn save(&mut self, unit: UncheckedSignedUnit<H, D, S>) -> Result<(), std::io::Error>;
 }
 
+/// Abstraction over Unit backup loading mechanism
 pub trait UnitLoader<H: Hasher, D: Data, S: Signature> {
     fn load(self) -> Result<Vec<UncheckedSignedUnit<H, D, S>>, LoaderError>;
 }
@@ -73,6 +76,10 @@ fn _on_shutdown(starting_round_tx: oneshot::Sender<Option<Round>>) {
     }
 }
 
+/// Loading mechanism first Unit data from `unit_loader` and awaits on response from unit collection.
+/// If loaded Units are compatible with unit collection result (meaning highest unit is from round at least
+/// unit_collection + 1) it send all loaded units by `loaded_unit_tx` and sends starting round number by
+/// `starting_round_tx`. If Units are not compatible it sends no Units and `None` by `starting_round_tx`
 pub async fn _run_loading_mechanism<H: Hasher, D: Data, S: Signature, UL: UnitLoader<H, D, S>>(
     unit_loader: UL,
     loaded_unit_tx: Sender<UncheckedSignedUnit<H, D, S>>,
--- a/consensus/src/runway/backup.rs
+++ b/consensus/src/runway/backup.rs
@@ -24,27 +24,40 @@ impl From<CodecError> for LoaderError {
 }
 
 /// Abstraction over Unit backup saving mechanism
-pub trait UnitSaver<H: Hasher, D: Data, S: Signature> {
-    fn save(&mut self, unit: UncheckedSignedUnit<H, D, S>) -> Result<(), std::io::Error>;
+pub struct _UnitSaver<W: Write> {
+    inner: W,
 }
 
 /// Abstraction over Unit backup loading mechanism
-pub trait UnitLoader<H: Hasher, D: Data, S: Signature> {
-    fn load(self) -> Result<Vec<UncheckedSignedUnit<H, D, S>>, LoaderError>;
+pub struct _UnitLoader<R: Read> {
+    inner: R,
 }
 
-impl<H: Hasher, D: Data, S: Signature, W: Write> UnitSaver<H, D, S> for W {
-    fn save(&mut self, unit: UncheckedSignedUnit<H, D, S>) -> Result<(), std::io::Error> {
-        self.write_all(&unit.encode())?;
-        self.flush()?;
+impl<W: Write> _UnitSaver<W> {
+    pub fn _new(write: W) -> Self {
+        Self { inner: write }
+    }
+
+    pub fn _save<H: Hasher, D: Data, S: Signature>(
+        &mut self,
+        unit: UncheckedSignedUnit<H, D, S>,
+    ) -> Result<(), std::io::Error> {
+        self.inner.write_all(&unit.encode())?;
+        self.inner.flush()?;
         Ok(())
     }
 }
 
-impl<H: Hasher, D: Data, S: Signature, R: Read> UnitLoader<H, D, S> for R {
-    fn load(mut self) -> Result<Vec<UncheckedSignedUnit<H, D, S>>, LoaderError> {
+impl<R: Read> _UnitLoader<R> {
+    pub fn _new(read: R) -> Self {
+        Self { inner: read }
+    }
+
+    fn _load<H: Hasher, D: Data, S: Signature>(
+        mut self,
+    ) -> Result<Vec<UncheckedSignedUnit<H, D, S>>, LoaderError> {
         let mut buf = Vec::new();
-        self.read_to_end(&mut buf)?;
+        self.inner.read_to_end(&mut buf)?;
         let input = &mut &buf[..];
         let mut result = vec![];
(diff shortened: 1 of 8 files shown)


consensus/src/runway/backup.rs
```diff
@@ -0,0 +1,331 @@
...
+    Ok((units, next_round))
+}
+
+fn _on_shutdown(starting_round_tx: oneshot::Sender<Option<Round>>) {
+    if starting_round_tx.send(None).is_err() {
+        warn!(target: "AlephBFT-runway", "Coulnd not send `None` starting round.");
+    }
+}
+
+/// Loading mechanism first Unit data from `unit_loader` and awaits on response from unit collection.
```
fbielejec, line 79: wording, I'm missing a verb here - first what?

consensus/src/runway/backup.rs
```diff
@@ -0,0 +1,331 @@
...
+/// If loaded Units are compatible with unit collection result (meaning highest unit is from round at least
```
fbielejec, line 80:
```suggestion
/// If loaded Units are compatible with the unit collection result (meaning the highest unit is from at least round
```

consensus/src/runway/backup.rs
```diff
@@ -0,0 +1,331 @@
...
+/// unit_collection + 1) it send all loaded units by `loaded_unit_tx` and sends starting round number by
```
fbielejec, line 81:
```suggestion
/// unit_collection + 1) it sends all loaded units by `loaded_unit_tx` and sends starting round number by
```

consensus/src/units/mod.rs
```diff
@@ -218,15 +218,73 @@ impl<H: Hasher> Unit<H> {
...
+    type Unit = GenericUnit<Hasher64>;
+    type FullUnit = GenericFullUnit<Hasher64, Data>;
+    type UncheckedSignedUnit = GenericUncheckedSignedUnit<Hasher64, Data, Signature>;
+
+    pub fn creator_set(n_members: NodeCount) -> Vec<Creator> {
+        let mut result = Vec::new();
+        for i in 0..n_members.0 {
+            result.push(Creator::new(NodeIndex(i), n_members));
+        }
+        result
```
fbielejec, line 250:
```suggestion
            (0..n_members.0).map(|i| Creator::new(NodeIndex(i), n_members)).collect::<Vec<Creator>>();        
```

consensus/src/units/mod.rs
```diff
@@ -218,15 +218,73 @@ impl<H: Hasher> Unit<H> {
...
+    }
+
+    pub fn create_units<'a, C: Iterator<Item = &'a Creator>>(
+        creators: C,
+        round: Round,
+    ) -> Vec<(PreUnit, Vec<<Hasher64 as Hasher>::Hash>)> {
+        let mut result = Vec::new();
```
fbielejec, line 257: similar, use functional pattern pls, `map(|_|).collect()`

consensus/src/units/mod.rs
```diff
@@ -218,15 +218,73 @@ impl<H: Hasher> Unit<H> {
...
+            result.push(
+                creator
+                    .create_unit(round)
+                    .expect("Creation should succeed."),
+            );
+        }
+        result
+    }
+
+    pub fn preunit_to_unit(preunit: PreUnit, session_id: SessionId) -> Unit {
```
fbielejec, line 268: could this be done in mor eidiomatic Rust by `impl From<PreUnit> for Unit`?

consensus/src/units/mod.rs
```diff
@@ -218,15 +218,73 @@ impl<H: Hasher> Unit<H> {
...
+        FullUnit::new(preunit, 0, session_id).unit()
+    }
+
+    pub fn add_units(creator: &mut Creator, units: &[Unit]) {
```
fbielejec, line 272: could use method chaining here instead of fugly interior mutbility:
```
impl for Creator {
  add_units(creator: &mut Creator, units: &[Unit]) -> Self {
    for unit in units {
      creator.add_unit(unit);
    }
    self.to_owned()
  }
}
```
and then:
 
```creator.add_units(units);```

File: consensus/src/units/mod.rs (hunk 6 above)
Comment (timorl) [reply consensus/src/units/mod.rs:268]: This is in tests, it's kinda strange implementing such things in tests, isn't it?

File: consensus/src/runway/backup.rs [hunk 8]
```diff
@@ -0,0 +1,331 @@
...
+}
+
+impl From<CodecError> for LoaderError {
+    fn from(err: CodecError) -> Self {
+        Self::Codec(err)
+    }
+}
+
+/// Abstraction over Unit backup saving mechanism
+pub trait UnitSaver<H: Hasher, D: Data, S: Signature> {
```
timorl, line 27: I don't think these need to be traits. You can just have them as structs which take a `W` and `R` and then implement `Write` and `Read` for the sake of tests.

consensus/src/runway/backup.rs
```diff
@@ -0,0 +1,331 @@
...
+
+    let starting_round = next_round;
+    if starting_round < highest_response {
+        error!(target: "AlephBFT-runway", "backup and unit collection missmatch. Backup got: {:?}, collection got: {:?}", starting_round, highest_response);
+        _on_shutdown(starting_round_tx);
+        return;
+    };
+
+    for u in units {
+        if let Err(e) = loaded_unit_tx.unbounded_send(u) {
```
timorl, line 115: I think this should happen before waiting for highest response, I don't see a reason to postpone it.

consensus/src/runway/mod.rs
```diff
@@ -18,8 +18,10 @@ use futures::{
 use log::{debug, error, info, trace, warn};
 use std::{collections::HashSet, convert::TryFrom};
 
+mod backup;
 mod collection;
 
+pub use backup::{UnitLoader, UnitSaver, _run_loading_mechanism as run_backup_loading_mechanism};
```
timorl, line 24: Why `pub`?

consensus/src/units/validator.rs
```diff
@@ -132,63 +132,15 @@ mod tests {
     use super::{ValidationError::*, Validator as GenericValidator};
     use crate::{
         creation::Creator as GenericCreator,
-        units::{
-            FullUnit as GenericFullUnit, PreUnit as GenericPreUnit,
-            UncheckedSignedUnit as GenericUncheckedSignedUnit, Unit as GenericUnit,
+        units::tests::{
```
timorl, line 135: These are already exported, so no `tests::` needed, right?

timorl (approved):
Nice, just a couple mostly nits.

consensus/src/runway/backup.rs
maciejnems: I thought that we could send unit only if we will start the creator. But I guess sending it earlier would be better.

consensus/src/runway/mod.rs
maciejnems: So that UnitLoader and UnitSaver could be passed to runway as arguments. Do you think passing std::io::Read/Write to runway would be better? _run_loading_mechanism should not be here, I will remove it.

timorl: Ah, yeah, all that makes sense.

consensus/src/runway/backup.rs
maciejnems: Both UnitSaver and UnitLoader don't need any internal state, just the methods provided by `W` and `R`. So I thought creating traits that have a default implementation for each `std::io::Write/Read` would be better. Otherwise i would have to create structs with only `inner: W/R`. Do you think such approach is better? I'm not sure what you mean by "implement Write and Read for the sake of tests".

consensus/src/units/mod.rs
maciejnems: I also feel like it kind of strange to implement such things for tests

consensus/src/units/mod.rs
maciejnems: I added this method for testing. I'm just not sure what for is chaining here. There is no case where this would be needed.

consensus/src/units/validator.rs
maciejnems: Yes. Also I moved those methods to module `testing`

consensus/src/runway/backup.rs
maciejnems: Changed

consensus/src/units/mod.rs
fbielejec: now you made me feel the same

fbielejec (approved):

consensus/src/runway/backup.rs
maciejnems: I changed it to have `UnitSaver/Loader<W/R>` and save/load methods are dependent on `<H,D,S>` do you think this is enough type safety or should I also parameterize `UnitSaver/Loader<W/R, H, D, S>` so that only one type of unit can be saved/loaded?

timorl: Hmm, that's a good point, I don't think it's strictly necessary, but I support doing this.

PR #196, opened 2022-04-15
AZ-693 Fix examples

lesniak43:
[A0-531](https://cardinal-cryptography.atlassian.net/browse/A0-531)


examples/dummy_honest/examples/dummy_honest.rs
```diff
@@ -0,0 +1,217 @@
+use aleph_bft::{run_session, Recipient};
```
DamianStraszak, line 1: Regarding the path to the current file: why nested examples again :P? I presume this is on purpose, but why not just put this into `main.rs`?

examples/dummy_honest/examples/dummy_honest.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    info!(target: "dummy-honest", "Getting network up.");
+    let (network, mut manager) = Network::new().await.unwrap();
+    let (close_network, exit) = oneshot::channel();
+    tokio::spawn(async move { manager.run(exit).await });
+
+    let data_provider = DataProvider::new();
+    let (finalization_handler, mut finalized_rx) = FinalizationHandler::new();
+
+    let units = Arc::new(Mutex::new(vec![]));
```
DamianStraszak, line 69: Comment for here and an analogous place in one of the other files: calling this `units` (and the things below `unit_loader` and `unit_saver`) suggests that it is a vector of units and per my understanding it is not -- it holds serialized data, units for now, but will likely also hold other things in the future.  I would call it either `buffer` or `raw_bytes` or something like that.

DamianStraszak (changes requested):
Great work, the AlephBFT repo is looking better and better!

examples/blockchain/examples/blockchain.rs
```diff
@@ -99,21 +94,14 @@ async fn main() {
...
+        let config = aleph_bft::default_config(args.n_members.into(), args.my_id.into(), 0);
+        let units = Arc::new(Mutex::new(vec![]));
+        let unit_loader = Loader::new((*units.lock()).clone());
+        let unit_saver = Saver::new(units);
+        let local_io =
+            aleph_bft::LocalIO::new(data_provider, finalization_handler, unit_saver, unit_loader);
+        run_session(config, local_io, network, keychain, Spawner {}, exit).await
     });
 
     let mut max_block_finalized = 0;
```
maciejnems, line 107: Similarly as in `dummy_honest` example, could be changed to something that would error!/panic! if finalization stream closes before we have `args.n_finalized` finalized blocks. And in future we could include it in our pipeline.

maciejnems (commented):
Very nice, examples are finally working again :partying_face:

examples/dummy_honest/examples/dummy_honest.rs
maciejnems: Right now this is called the same in `LocalIO`. For now those names are accurate (`unit_loader` and `unit_saver`). Do you think we should change it there too? In future we would also like to save/load Alerts. Maybe something like `backup_loader`, `backup_saver`?

DamianStraszak: The names of the arguments in `aleph_bft::LocalIO` should be indeed  `backup_loader`, `backup_saver` or `backup_reader/writer`. The user doesn't have to know that we will be storing units there (and actually should not assume that, because we likely will store other things there as well). On the other hand, the internal structs name `UnitSaver` and `UnitLoader` can be kept, because this is accurate for now.

examples/dummy_honest/examples/dummy_honest.rs
```diff
@@ -0,0 +1,217 @@
...
+    let unit_loader = Loader::new((*units.lock()).clone());
```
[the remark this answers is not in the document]
DamianStraszak: Yes, by doing it this way, the vector is not quite shared between loader and saver. If we wanted to make it right I think one would need to fix the mocks, so that you can do something like 
`let loader, saver = new_loader_and_saver()` and `new_loader_and_saver` would create these sharing a Mutex wrapped `Vec`.

examples/dummy_honest/examples/dummy_honest.rs
maciejnems: Ok, `aleph_bft::LocalIO::new` arguments can be changed in https://github.com/Cardinal-Cryptography/AlephBFT/pull/197. Here I would go with changing those variable names to `backup`, `backup_loader` and `backup_saver`.

examples/dummy_honest/examples/dummy_honest.rs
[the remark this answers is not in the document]
maciejnems: But saver and loader should not share a mutex. Saver should have no impact on data stored in loader and loader should have data backed up before crash, so in this case empty vector.

examples/blockchain/examples/blockchain.rs
```diff
@@ -1,40 +1,38 @@
-use futures::{channel::oneshot, StreamExt};
-use log::{debug, info};
+extern crate aleph_bft_examples_blockchain;
```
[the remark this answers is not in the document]
lesniak43: the idea is that now you can run it using `cargo run --example blockchain` - see Jira (link in PR description) for details :P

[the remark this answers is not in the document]
lesniak43: although I'm not saying that it's a good or bad idea - I'd feel much more comfortable if the decision was made here by my superiors, namely you two :D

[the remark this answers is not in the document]
DamianStraszak: Ok, I got it now. I'm not 100% convinced this is the right approach, but I can live with that :)

examples/dummy_honest/examples/dummy_honest.rs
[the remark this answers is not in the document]
DamianStraszak: Yeah, you are right, you can disregard what I said above :)

.github/workflows/publish-packages.yml
```diff
@@ -17,3 +17,4 @@ jobs:
         with:
           registry-token: ${{ secrets.CRATES_IO_TOKEN }}
           ignore-unpublished-changes: true
+          publish-delay: 30000
```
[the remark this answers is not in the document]
lesniak43: this action failed when it was run for the first time - started uploading `mock` before `crypto` was availabe or sth like that, - and I had to rerun it manually, and this time it managed to upload multiple packages with no problems, so it's really not consistent - Tomasz decided we should have a 30s delay just to be sure

examples/blockchain/examples/blockchain.rs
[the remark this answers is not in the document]
lesniak43: after talking with @DamianStraszak - leaving it as it is to have `cargo build|run --example ...` etc.

examples/blockchain/examples/blockchain.rs
```diff
@@ -1,40 +1,38 @@
...
+use futures::{channel::oneshot, StreamExt};
+use log::{debug, info};
+use parking_lot::Mutex;
+use std::{io::Write, sync::Arc, time};
 
 const TXS_PER_BLOCK: usize = 50000;
 const TX_SIZE: usize = 300;
-const BLOCK_TIME_MS: u64 = 1000;
-const INITIAL_DELAY_MS: u64 = 5000;
+const BLOCK_TIME_MS: BlockNum = 1000;
```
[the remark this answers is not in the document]
lesniak43: fixed

examples/blockchain/src/chain.rs
```diff
@@ -40,30 +36,30 @@ impl Block {
...
     // Delay before the first block should be created
-    pub(crate) init_delay_ms: u64,
+    pub init_delay_ms: BlockNum,
     // f(k) means who should author the kth block
-    pub(crate) authorship_plan: BlockPlan,
+    pub authorship_plan: BlockPlan,
 }
 
-pub(crate) fn gen_chain_config(
+pub fn gen_chain_config(
```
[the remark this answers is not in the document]
lesniak43: fixed

examples/dummy_honest/examples/dummy_honest.rs
lesniak43: as above

examples/dummy_honest/examples/dummy_honest.rs
lesniak43: renamed, also removed `units` as `Loader` and `Saver` should not (?) share data

examples/dummy_honest/examples/dummy_honest.rs
```diff
@@ -0,0 +1,217 @@
...
+
+    let (close_member, exit) = oneshot::channel();
+    tokio::spawn(async move {
+        let keychain = Keychain::new(args.n_members.into(), args.my_id.into());
+        let config = aleph_bft::default_config(args.n_members.into(), args.my_id.into(), 0);
+        run_session(config, local_io, network, keychain, Spawner {}, exit).await
+    });
+
+    let mut finalized = vec![];
+    while let Some(data) = finalized_rx.next().await {
```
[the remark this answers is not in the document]
lesniak43: fixed

examples/blockchain/examples/blockchain.rs
lesniak43: fixed

DamianStraszak (approved):

examples/blockchain/examples/blockchain.rs
```diff
@@ -125,13 +116,18 @@ async fn main() {
...
     }
+    if max_block_finalized < args.n_finalized as u32 {
+        error!(target: "Blockchain-main", "Finalization stream finished too soon. Highest finalized = {:?}, expected {:?}", max_block_finalized, args.n_finalized);
+        panic!();
+    }
+
     let stop_time = time::Instant::now();
-    let tot_millis = (stop_time - start_time).as_millis() - INITIAL_DELAY_MS as u128;
-    let tps = (n_finalized as f64) * (TXS_PER_BLOCK as f64) / (0.001 * (tot_millis as f64));
+    let tot_millis = (stop_time - start_time).as_millis() - INITIAL_DELAY.as_millis();
```
maciejnems, line 129: Here could be just 
```
let tot_millis = (stop_time - start_time - INITIAL_DELAY).as_millis();
```

examples/blockchain/examples/blockchain.rs
maciejnems: I would also add some message to the panic, so it prints something. Maybe just "Finalization stream finished too soon"?

maciejnems: I would also add some message to the panic, so it prints something. Maybe just "Finalization stream finished too soon"?

lesniak43: fixed (also in `dummy_honest.rs`)

examples/blockchain/examples/blockchain.rs
lesniak43: fixed

examples/blockchain/examples/blockchain.rs
```diff
@@ -125,13 +116,18 @@ async fn main() {
...
+        panic!("Finalization stream finished too soon.");
+    }
+
     let stop_time = time::Instant::now();
-    let tot_millis = (stop_time - start_time).as_millis() - INITIAL_DELAY_MS as u128;
-    let tps = (n_finalized as f64) * (TXS_PER_BLOCK as f64) / (0.001 * (tot_millis as f64));
+    let tot_millis = (stop_time - start_time - INITIAL_DELAY).as_millis();
+    let tps = (args.n_finalized as f64) * (TXS_PER_BLOCK as f64) / (0.001 * (tot_millis as f64));
     info!(target: "Blockchain-main", "Achieved {:?} tps.", tps);
     close_member.send(()).expect("should send");
```
maciejnems, line 132: We should probably also add some awaits here. So something like:
```
close_member.send(()).expect("should send");
member_handle.await.unwrap();
close_chain.send(()).expect("should send");
chain_handle.await.unwrap();
close_network.send(()).expect("should send");
network_handle.await.unwrap();
```

examples/dummy_honest/examples/dummy_honest.rs
```diff
@@ -0,0 +1,221 @@
...
+    for i in 0..args.n_finalized {
+        match finalized_rx.next().await {
+            Some(_) => debug!(target: "dummy-honest", "Got new batch. Finalized: {:?}", i+1),
+            None => {
+                error!(target: "dummy-honest", "Finalization stream finished too soon. Got {:?} batches, wanted {:?} batches", i+1, args.n_finalized);
+                panic!("Finalization stream finished too soon.");
+            }
+        }
+    }
+    close_member.send(()).expect("should send");
```
maciejnems, line 94: Here we should probably add some awaits
```
close_member.send(()).expect("should send");
member_handle.await.unwrap();
close_network.send(()).expect("should send");
network_handle.await.unwrap();
```

maciejnems (changes requested):
Sorry for not noticing this earlier :(

maciejnems (approved):

PR #199, opened 2022-04-25
Bump parity-scale-codec dependency to 3.0

h4nsu:


pmikolajczyk41 (approved):
there is nothing I can complain about :)

Yithis (approved):

PR #200, opened 2022-04-25
AZ-298 Fix fuzz

lesniak43:
[A0-526](https://cardinal-cryptography.atlassian.net/browse/A0-526)


timorl (approved):

fixxxedpoint (changes requested):
LGTM. I think you can remove Cargo.lock from fuzz, since it is now on the `members` list and shares project's Cargo.lock.

fixxxedpoint (approved):

PR #214, opened 2022-05-23
A0-911 Remove libp2p from examples

lesniak43:
[A0-911](https://cardinal-cryptography.atlassian.net/browse/A0-911)


examples/blockchain/src/blockchain.rs
```diff
@@ -0,0 +1,119 @@
...
+    fn add_transaction(&mut self, transaction: Transaction) -> bool {
+        self.transactions_buffer.push(transaction);
+        if self.transactions_buffer.len() == self.n_tr_per_block as usize {
+            self.create_block();
+        }
+        self.blocks.len() == self.n_blocks as usize
+    }
+
+    fn create_block(&mut self) {
+        let mut transactions: Vec<ProcessedTransaction> = vec![];
```
mike1729, line 70: `let transactions = self.transactions_buffer.drain(..).map(|tx| self.process_transaction(tx)).collect();`
This is more idiomatic and takes care of line 83.

examples/blockchain/src/network.rs
```diff
@@ -1,363 +1,346 @@
...
+                            Ok(Message::DNSClientRequest(address)) => self.dns_request(address),
+                            Ok(Message::DNSResponse(addresses)) => for (id, addr) in addresses.iter().enumerate() {
+                                if let Some(addr) = addr {
+                                    self.addresses[id as usize] = Some(addr.clone());
+                                };
+                            },
+                            Ok(Message::Crash) => if let Some(ch) = self.crash.take() {
+                                ch.send(()).unwrap_or(());
+                            },
+                            Err(_) => (),
```
mike1729, line 176: This is decode error.

examples/blockchain/src/network.rs
```diff
@@ -1,363 +1,346 @@
...
+                        },
+                        Err(_) => {
+                            error!("Could not decode incoming data");
```
mike1729, line 179: This is reading from buffer.

examples/blockchain/src/network.rs
```diff
@@ -1,363 +1,346 @@
...
+                event = self.listener.accept() => match event {
+                    Ok((mut socket, _addr)) => match socket.read_to_end(&mut buffer).await {
+                        Ok(_) => match Message::decode(&mut &buffer[..]) {
+                            Ok(Message::DNSResponse(addresses)) => for (id, addr) in addresses.iter().enumerate() {
+                                if let Some(addr) = addr {
+                                    self.addresses[id as usize] = Some(addr.clone());
+                                };
+                            },
+                            Ok(_) => (),
+                            Err(_) => (),
```
mike1729, line 325: This is decode error.

examples/blockchain/src/network.rs
```diff
@@ -1,363 +1,346 @@
...
+                        },
+                        Err(_) => {
+                            error!("Could not decode incoming data");
```
mike1729, line 328: This is read from socket error.

mike1729 (changes requested):
Two general notes:
```
match x {
  Ok(_) => (),
  Err(e) => f(e),
}
```
this pattern is frequent in your code. I would prefer either `if let Err(e) = ...` or `x.err().map(|e| f(e))` but don't have strong opinion here.
Second, use `log` instead of `println`.

maciejnems:
I don't think that rewriting blockchain example is in scope of this task. From what I understand the idea was to remove libp2p and connect old blockchain example with the new network. I would suggest to do that and leave rewriting blockchain example for different PR which solves a different task. Also I'm not sure whether this example needs complete rewrite. From what I've seen it was working correctly. That said @lesniak43 mentioned to me that there was something flaky about this example, so maybe it is broken.

examples/ordering/src/network.rs
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    fn send(&self, data: NetworkData, recipient: &Recipient) -> io::Result<()> {
+        match recipient {
+            Recipient::Everyone => {
+                for r in 0..self.addresses.len() {
+                    self.send_to_peer(data.clone(), r)?;
+                }
+            }
+            Recipient::Node(r) => self.send_to_peer(data, (*r).0)?,
```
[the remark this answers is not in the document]
lesniak43: fixed

examples/ordering/src/network.rs
```diff
@@ -0,0 +1,82 @@
...
+
+impl Network {
+    pub async fn new(my_id: usize, ports: &[usize]) -> Result<Self, Box<dyn std::error::Error>> {
+        assert!(my_id < ports.len());
+        let mut addresses = vec![];
+        for port in ports {
+            addresses.push(format!("127.0.0.1:{}", port).parse::<SocketAddr>()?);
+        }
+        let listener =
+            TcpListener::bind(format!("127.0.0.1:{}", ports.get(my_id).unwrap())).await?;
```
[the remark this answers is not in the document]
lesniak43: fixed

examples/ordering/src/network.rs
```diff
@@ -0,0 +1,82 @@
...
+            stream.write_all(&data.encode())?;
+        }
+        Ok(())
+    }
+}
+
+#[async_trait::async_trait]
+impl aleph_bft::Network<NetworkData> for Network {
+    fn send(&self, data: NetworkData, recipient: Recipient) {
+        match self.send(data, &recipient) {
```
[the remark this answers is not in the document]
lesniak43: fixed

mike1729 (approved):

examples/ordering/run.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -e
+
+cargo build --release
+
+clear
+
+for i in {0..4}; do
```
maciejnems, line 9: Maybe we could make this script also dependent on a number of participants?

examples/ordering/src/network.rs
```diff
@@ -0,0 +1,80 @@
...
+                }
+            }
+            Recipient::Node(r) => self.send_to_peer(data, r.0)?,
+        }
+        Ok(())
+    }
+
+    fn send_to_peer(&self, data: NetworkData, recipient: usize) -> io::Result<()> {
+        assert!(recipient < self.addresses.len());
+        if recipient != self.my_id {
```
maciejnems, line 48: I don't think we should hide not sending messages to ourself from AlephBFT. I would remove this if and this shouldn't break anything, right?

examples/ordering/src/network.rs
```diff
@@ -0,0 +1,80 @@
...
+            addresses,
+            listener,
+        })
+    }
+
+    fn send(&self, data: NetworkData, recipient: &Recipient) -> io::Result<()> {
+        match recipient {
+            Recipient::Everyone => {
+                for r in 0..self.addresses.len() {
+                    self.send_to_peer(data.clone(), r)?;
```
maciejnems, line 38: We could filter out here sending to ourself

maciejnems (changes requested):
Very nice change, finally no p2p. Haven't looked at blockchain example yet but here are the comments for ordering example.

examples/blockchain/src/network.rs
```diff
@@ -233,130 +95,200 @@ impl aleph_bft::Network<NetworkData> for Network {
 }
 
 pub struct NetworkManager {
-    swarm: Swarm<Behaviour>,
+    id: usize,
+    address: Address,
+    addresses: Vec<Option<Address>>,
+    bootnodes: HashMap<u32, Address>,
```
maciejnems, line 101: You could keep addresses in a `HashMap` and bootnodes in `HashSet`. I guess this would make stuff simpler. You would not have to create a vec of `None`.

examples/blockchain/src/network.rs
```diff
@@ -233,130 +95,200 @@ impl aleph_bft::Network<NetworkData> for Network {
...
-        // Create a tokio-based TCP transport use noise for authenticated
-        // encryption and Mplex for multiplexing of substreams on a TCP stream.
-        let transport = TokioTcpConfig::new()
-            .nodelay(true)
-            .upgrade(upgrade::Version::V1)
-            .authenticate(noise::NoiseConfig::xx(noise_keys).into_authenticated())
-            .multiplex(mplex::MplexConfig::new())
-            .boxed();
+        let mut addresses = vec![None; n_nodes];
+        for (id, addr) in &bootnodes {
```
maciejnems, line 127: If you change addresses to `HashMap` you could just add the new address to bootnodes and pass them ass addresses.

examples/blockchain/src/network.rs
```diff
@@ -233,130 +95,200 @@ impl aleph_bft::Network<NetworkData> for Network {
...
+
+    fn dns_response(&mut self, id: usize, address: Address) {
+        self.addresses[id] = Some(address.clone());
+        self.try_send(Message::DNSResponse(self.addresses.clone()), &address)
+            .unwrap_or(());
+    }
+
     pub async fn run(&mut self, mut exit: oneshot::Receiver<()>) {
+        let dns_ticker_delay = std::time::Duration::from_millis(1000);
+        let mut dns_ticker = Delay::new(dns_ticker_delay).fuse();
```
maciejnems, line 221: You don't need to fuse `dns_ticker` and `dns_hello_ticker`. In the loop you mutate them anyway, to a new future. Also maybe `timeout` would be better than ticker?.

maciejnems: Also you could use `tokio::time::interval` here if you would like to use a fancy way for intervals

examples/blockchain/src/network.rs
```diff
@@ -233,130 +95,200 @@ impl aleph_bft::Network<NetworkData> for Network {
...
+                    dns_ticker = Delay::new(dns_ticker_delay).fuse();
+                },
+
+                _ = &mut dns_hello_ticker => {
+                    self.send(Message::DNSHello(self.id as u32, self.address.clone()), Recipient::Everyone);
+                    debug!("Sending Hello!");
+                    dns_hello_ticker = Delay::new(dns_hello_ticker_delay).fuse();
+                },
+
                 maybe_msg = self.consensus_rx.next() => {
```
maciejnems, line 279: This can be changed to 
```
Some((consensus_msg, recipient)) = self.consensus_rx.next() => {
```

examples/blockchain/src/network.rs
```diff
@@ -233,130 +95,200 @@ impl aleph_bft::Network<NetworkData> for Network {
...
                     if let Some((consensus_msg, recipient)) = maybe_msg {
-                        let handle = &mut self.swarm.behaviour_mut();
-                        handle.send_consensus_message(consensus_msg, recipient);
+                        self.send(Message::Consensus(consensus_msg), recipient);
                     }
                 }
+
                 maybe_block = self.block_rx.next() => {
```
maciejnems, line 285: This can be changed to
```
Some(block)  = self.block_rx.next() => {
```

examples/blockchain/src/network.rs
```diff
@@ -233,130 +95,200 @@ impl aleph_bft::Network<NetworkData> for Network {
...
+                _ = &mut dns_ticker => {
+                    if self.addresses.iter().any(|a| a.is_none()) {
+                        self.send(Message::DNSRequest(self.id as u32, self.address.clone()), Recipient::Everyone);
+                        debug!("Requesting IP addresses");
+                    }
+                    dns_ticker = Delay::new(dns_ticker_delay).fuse();
+                },
+
+                _ = &mut dns_hello_ticker => {
+                    self.send(Message::DNSHello(self.id as u32, self.address.clone()), Recipient::Everyone);
```
maciejnems, line 274: I feel like right now Hello and Request are doing a lot of similar things. I think it is fine as it is in current state, but maybe gossiping Hello would be more like an actual network? Then you could also remove the Request sending.

examples/blockchain/src/network.rs
```diff
@@ -233,130 +95,200 @@ impl aleph_bft::Network<NetworkData> for Network {
...
+            self.addresses[n] = None;
+        }
+    }
+
+    fn send(&mut self, message: Message, recipient: Recipient) {
+        let addr = self.recipient_to_addresses(recipient);
+        for (n, addr) in addr.into_iter() {
+            match self.try_send(message.clone(), &addr) {
+                Ok(_) => (),
+                Err(_) => self.reset_dns(n),
```
maciejnems, line 203: Maybe some `error!` about not being able to send a message?

examples/blockchain/src/network.rs
```diff
@@ -233,130 +95,200 @@ impl aleph_bft::Network<NetworkData> for Network {
...
+        if !self.bootnodes.contains_key(&(n as u32)) {
+            error!("Reseting address of node {}", n);
+            self.addresses[n] = None;
+        }
+    }
+
+    fn send(&mut self, message: Message, recipient: Recipient) {
+        let addr = self.recipient_to_addresses(recipient);
+        for (n, addr) in addr.into_iter() {
+            match self.try_send(message.clone(), &addr) {
```
maciejnems, line 201: Could be
```
if self.try_send(message.clone(), &addr).is_err() {
```

examples/ordering/src/network.rs
lesniak43: fixed

examples/ordering/src/network.rs
lesniak43: fixed

examples/blockchain/src/network.rs
```diff
@@ -233,130 +95,200 @@ impl aleph_bft::Network<NetworkData> for Network {
...
+
+    fn reset_dns(&mut self, n: usize) {
+        if !self.bootnodes.contains_key(&(n as u32)) {
+            error!("Reseting address of node {}", n);
+            self.addresses[n] = None;
+        }
+    }
+
+    fn send(&mut self, message: Message, recipient: Recipient) {
+        let addr = self.recipient_to_addresses(recipient);
```
maciejnems, line 199: I'm not sure this function does much. If you had addresses in a `HashMap` you could just match here and have one simple get instead of what this function does in case of `Recipient::Node`. In case of `Recipient::Everyone` you could just iterate over items

maciejnems (changes requested):
Blockchain part of review

examples/ordering/run.sh
lesniak43: I'd rather leave it this way - the example requires that we specify the ports manually, and I prefer the slightly weaker assumption "ports 43000-43004 are free" rather than "ports 43000-(43000+n-1) are free"...

examples/ordering/src/network.rs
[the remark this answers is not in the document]
lesniak43: slightly reorganized `send_sth` methods

examples/ordering/src/network.rs
```diff
@@ -0,0 +1,80 @@
...
+                }
+            }
+            Recipient::Node(r) => self.send_to_peer(data, r.0)?,
+        }
+        Ok(())
+    }
+
+    fn send_to_peer(&self, data: NetworkData, recipient: usize) -> io::Result<()> {
+        assert!(recipient < self.addresses.len());
```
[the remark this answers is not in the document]
lesniak43: removed, will panic in the next line anyway...

examples/blockchain/src/network.rs
lesniak43: fixed

examples/blockchain/src/network.rs
lesniak43: fixed

examples/blockchain/src/network.rs
lesniak43: fixed

examples/blockchain/src/network.rs
lesniak43: `Hello` is a quick fix for handling bootnode crashes, but I do agree that a small p2p module could be extracted from the current network implementation. Leaving as it is.

examples/blockchain/src/network.rs
lesniak43: fixed

examples/blockchain/src/network.rs
lesniak43: yeah, but still it's ~15 lines of code, so I'd leave it as a separate function

examples/ordering/run.sh
lesniak43: on the other hand, fixed... (also, see changes in README)

maciejnems (commented):
Could you also add `.gitignore` files to both examples with `node*.log` inside?

examples/blockchain/src/network.rs
lesniak43: yes :D

examples/blockchain/src/network.rs
lesniak43: fixec

maciejnems (approved):
Very nice, good job :smile:

PR #222, opened 2022-06-08
Update toolchain

Yithis:


rust-toolchain
```diff
@@ -1 +1 @@
-nightly-2021-10-24
+nightly-2022-05-23
```
fbielejec, line 1: if you could do it for the node repo :smile:

artemis-cc (changes requested):
Looks good to me, just a couple things:
* The notes that were in `#[must_use]` attributes should go in actual documentation, not just regular comments (fbielejec touched on this).
* This feels like a lot of `Eq` impls to add so casually. Was it just an oversight that they weren't there before, or could it have been intentional? Might be good to talk about this in the final commit message also.
* There are at least four(!) different orderings of `#[derive(Debug, Eq, PartialEq)]` here. Exactly what the order is doesn't matter so much, but it's probably better to pick one and use it consistently. (On the other hand, that _is_ sort of a larger issue that would be addressed by a style guide, and would probably cause a fair bit of churn, so maybe here and now isn't the place/time to do it…)

rust-toolchain
Yithis: Its already done for aleph-node repo

consensus/src/alerts/mod.rs
```diff
@@ -270,7 +270,7 @@ impl<'a, H: Hasher, D: Data, MK: MultiKeychain> Alerter<'a, H, D, MK> {
         Some(full_unit1.creator())
     }
 
-    #[must_use = "`rmc_alert()` registers the RMC but does not actually send it; the returned hash must be passed to `start_rmc()` separately"]
+    // `rmc_alert()` registers the RMC but does not actually send it; the returned hash must be passed to `start_rmc()` separately
```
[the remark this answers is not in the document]
Yithis: Done everywhere

Yithis:
@artemis-cc 

- Done
- I really don't know what was the reason, probably someone needed only `PartailEq` and didn't add `Eq`. Now it is forced by clippy
- Agree, there should be rule about this, but I don't feel it is in the scope of those changes

fbielejec (approved):

artemis-cc (approved):

PR #224, opened 2022-06-15
Substitute `chrono` crate with `time`

pmikolajczyk41:
This is needed to unblock #215. The only difference is that the time in example logs are in UTC.

### DIFF
--- a/examples/blockchain/src/main.rs
+++ b/examples/blockchain/src/main.rs
@@ -67,7 +67,7 @@ async fn main() {
                 "{} {}: {}",
                 record.level(),
                 OffsetDateTime::now_local()
-                    .unwrap_or(OffsetDateTime::now_utc())
+                    .unwrap_or_else(|_| OffsetDateTime::now_utc())
                     .format(&time_format)
                     .unwrap(),
                 record.args()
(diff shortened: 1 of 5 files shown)


examples/blockchain/src/main.rs
```diff
@@ -50,21 +58,23 @@ struct Args {
...
+        format_description!("[year]-[month]-[day] [hour]:[minute]:[second].[subsecond digits:3]");
     env_logger::builder()
-        .format(|buf, record| {
+        .format(move |buf, record| {
             writeln!(
                 buf,
                 "{} {}: {}",
                 record.level(),
-                Local::now().format("%Y-%m-%d %H:%M:%S%.3f"),
+                OffsetDateTime::now_utc().format(&time_format).unwrap(),
```
kostekIV, line 69: there is a method `now_local`, doesn't it works here?

kostekIV (approved):

examples/blockchain/src/main.rs
pmikolajczyk41: I must have missed that (it required another feature flag)

I added it, but it doesn't work for me (throws `IndeterminateOffset` error), so I kept UTC as fallback

obrok (approved):
I take that back, missed the explanation somehow.

PR #227, opened 2022-06-21
A0-861 Improve logging v2

lesniak43:


consensus/src/units/store.rs
```diff
@@ -28,6 +36,25 @@ impl<'a, H: Hasher, D: Data, KB: KeyBox> UnitStore<'a, H, D, KB> {
...
+        let mut top_row: Vec<(usize, Round)> = self
+            .by_coord
+            .keys()
+            .map(|c| (c.creator, c.round))
+            .into_group_map()
+            .into_iter()
+            .map(|(k, v)| (k, v.into_iter().max().unwrap_or(0)))
+            .map(|(k, v)| (k.0, v))
+            .collect();
+        top_row.sort();
```
DamianStraszak, line 49: This computes "highest unit per node index", apart from that it would be good to signal if some unit is missing, i.e., suppose that we hold (for a specific node index) units from rounds `0, 1, 2, 4, 5, 6, 7, 8, 10` then the above will just tell us `10` -- but I would also like to know that `3` is the earliest round that we are missing. This information can be printed only if there is a "hole", otherwise we can skip it. 

Another thing is this `.unwrap_or(0)` -- we don't differentiate between the `0`th unit and no units?

DamianStraszak (changes requested):
Solid work -- just a few suggestions.

consensus/src/member.rs
```diff
@@ -135,6 +136,66 @@ impl<D: Data, DP: DataProvider<D>, FH: FinalizationHandler<D>, US: Write, UL: Re
...
+            self.not_resolved_parents.len()
+        )?;
+        write!(f, "; task queue content: ")?;
+        write!(
+            f,
+            "CoordRequest - {}, ParentsRequest - {}, UnitMulticast - {}, RequestNewest - {}",
+            count.0, count.1, count.2, count.3
+        )?;
+        if !pending_tasks.is_empty() {
+            write!(f, "; pending tasks - {:?}", pending_tasks)?;
```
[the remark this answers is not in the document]
lesniak43: Maybe we'd leave it and increase the counter lower bound in line 171? The idea was to show/count tasks that spend too much time in the queue.

consensus/src/runway/mod.rs
```diff
@@ -629,6 +671,24 @@ where
...
     }
 
+    fn status_report(&self) {
+        let UnitStoreStatus {
+            forkers,
+            size,
+            height,
+            top_row,
+        } = self.store.get_status();
+        let runway_status: RunwayStatus<H> = RunwayStatus {
```
timorl, line 681: It might be nicer to create this struct from a `UnitStoreStatus` and the two remaining refs. You could then implement `Display` for `UnitStoreStatus` and use it in `RunwayStatus`'s implementation. Perhaps you could then also use constructors properly, `pub` fields are sorta icky. <_<

consensus/src/member.rs
```diff
@@ -135,6 +136,66 @@ impl<D: Data, DP: DataProvider<D>, FH: FinalizationHandler<D>, US: Write, UL: Re
     }
 }
 
+struct MemberStatus<'a, H, D, S>
```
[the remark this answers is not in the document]
timorl: Eh, I saw the other instance and perhaps for consistency this can stay as a struct, but then please make a constructor rather than `pub`ing all fields.

timorl (changes requested):
Looks mostly ok, some comments nonetheless.

Kinda unsure whether the gain of using itertools justifies adding an additional dependency, but my feeling of ambivalence is weak enough that I did not even make this into a comment.

consensus/src/member.rs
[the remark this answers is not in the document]
DamianStraszak: Ok, you can probably do that (`counter > 5` ?), but if you choose so, then:
1. The `scheduled_time` display is pretty useless in the form `Instant { tv_sec: 1974910, tv_nsec: 936698995 }` -- the best fix is probably to not display this field at all.
2. `pending tasks` could be changed to something like `long time pending tasks` to describe the intention better.

consensus/src/runway/collection.rs
```diff
@@ -236,6 +237,11 @@ impl<'a, H: Hasher, D: Data, MK: KeyBox> IO<'a, H, D, MK> {
         {
             warn!(target: "AlephBFT-runway", "unable to send resolved request:  {}", e);
         }
+        info!(target: "AlephBFT-runway", "Finished initial unit collection with status: {:?}", self.collection.status());
+    }
+
+    fn status_report(&self) {
+        info!(target: "AlephBFT-runway", "Initial unit collection status report: {:?}", self.collection.status());
```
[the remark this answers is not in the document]
DamianStraszak: The rationale here is as follows: we want this logged in all production nodes (ours and then also external), because it's crucial diagnostic information. So how to achieve it? I think the only reasonable way is to log it as `info`. If this is debug, and one must add special flags to enable this log then I'm 100% sure that when the time comes and the logs will be needed, then by some strange coincidence the logs will turn out to be disabled.

consensus/src/member.rs
```diff
@@ -135,6 +136,66 @@ impl<D: Data, DP: DataProvider<D>, FH: FinalizationHandler<D>, US: Write, UL: Re
...
+                Task::UnitMulticast(_) => false,
+                _ => st.counter >= 1,
+            })
+            .collect();
+        write!(f, "Member status report: ")?;
+        write!(
+            f,
+            "not_resolved_coords.len() - {}",
+            self.not_resolved_coords.len()
+        )?;
```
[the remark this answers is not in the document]
lesniak43: fixed

consensus/src/member.rs
```diff
@@ -135,6 +136,66 @@ impl<D: Data, DP: DataProvider<D>, FH: FinalizationHandler<D>, US: Write, UL: Re
...
+        write!(
+            f,
+            "; not_resolved_parents.len() - {}",
+            self.not_resolved_parents.len()
+        )?;
+        write!(f, "; task queue content: ")?;
+        write!(
+            f,
+            "CoordRequest - {}, ParentsRequest - {}, UnitMulticast - {}, RequestNewest - {}",
+            count.0, count.1, count.2, count.3
```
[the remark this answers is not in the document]
lesniak43: renamed

consensus/src/member.rs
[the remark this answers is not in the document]
lesniak43: so how should I approach this? I'd like to implement `Display` for `ScheduledTask`, which prints the counter value and `Debug`-prints the inner Task, and then print all `ScheduledTask`s in a loop (with a trailing comma, because convenience...) - is this ok?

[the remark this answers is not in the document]
lesniak43: @DamianStraszak approved, fixed

consensus/src/runway/collection.rs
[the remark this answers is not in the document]
lesniak43: done

consensus/src/runway/mod.rs
lesniak43: removed more `pub`s

consensus/src/member.rs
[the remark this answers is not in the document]
lesniak43: removed `pub`s

consensus/src/units/store.rs
lesniak43: 0 was just for safety :P and I use `Vec<(usize, Round)>` (for better readability), so in case of no units there will be no entry

lesniak43: added holes

consensus/src/units/store.rs
```diff
@@ -28,6 +79,46 @@ impl<'a, H: Hasher, D: Data, KB: KeyBox> UnitStore<'a, H, D, KB> {
...
+        let mut first_missing_rounds: Vec<(usize, Round)> = gm
+            .collect::<Vec<_>>()
+            .into_iter()
+            .map(|(id, rounds)| (id, first_missing(rounds)))
+            .filter(|(id, round)| match top_row.get(id) {
+                Some(top_round) => round < top_round,
+                None => false,
+            })
+            .collect();
+        let mut top_row: Vec<(usize, Round)> = top_row.into_iter().collect();
```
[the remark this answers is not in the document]
timorl: Ah, I think I see, you want this sorted...

It might be better to keep these as `HashMap`s and print them out more explicitly in the `Display` implementation, e.g. by using something like `(1..node_count).for_each(|node_id| top_row[node_id]...)`.

timorl (changes requested):
A bit of a mess in the store. :<

consensus/src/runway/collection.rs
```diff
@@ -132,6 +133,7 @@ pub struct Collection<'a, MK: KeyBox> {
     validator: &'a Validator<'a, MK>,
     starting_round: Round,
     responders: HashSet<NodeIndex>,
+    collected_responses: Vec<Option<Round>>,
```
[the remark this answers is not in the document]
lesniak43: fixed

consensus/src/member.rs
```diff
@@ -135,6 +146,95 @@ impl<D: Data, DP: DataProvider<D>, FH: FinalizationHandler<D>, US: Write, UL: Re
...
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let mut count_coord_request: usize = 0;
+        let mut count_parents_request: usize = 0;
+        let mut count_unit_multicast: usize = 0;
+        let mut count_request_newest: usize = 0;
+        for task in self.task_queue.iter().map(|st| &st.task) {
+            match task {
+                Task::CoordRequest(_) => count_coord_request += 1,
+                Task::ParentsRequest(..) => count_parents_request += 1,
```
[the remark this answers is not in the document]
lesniak43: https://github.com/rust-dev-tools/fmt-rfcs/issues/49
let's leave it that way?

consensus/src/units/store.rs
```diff
@@ -1,5 +1,56 @@
 use super::*;
+use itertools::Itertools;
 use log::{trace, warn};
+use std::fmt;
+
+pub(crate) struct UnitStoreStatus<'a> {
```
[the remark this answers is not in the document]
lesniak43: changed to `pub`, don't understand the second part

consensus/src/units/store.rs
```diff
@@ -28,6 +79,46 @@ impl<'a, H: Hasher, D: Data, KB: KeyBox> UnitStore<'a, H, D, KB> {
         }
     }
 
+    pub fn get_status(&self) -> UnitStoreStatus {
```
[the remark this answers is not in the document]
lesniak43: it won't be easier...

consensus/src/units/store.rs
```diff
@@ -28,6 +79,46 @@ impl<'a, H: Hasher, D: Data, KB: KeyBox> UnitStore<'a, H, D, KB> {
...
+            }
+        }
+        let gm = self
+            .by_coord
+            .keys()
+            .map(|c| (c.creator.0, c.round))
+            .into_grouping_map();
+        let top_row: HashMap<usize, Round> = gm.clone().max();
+        let mut first_missing_rounds: Vec<(usize, Round)> = gm
+            .collect::<Vec<_>>()
```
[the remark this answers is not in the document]
lesniak43: rewritten to use `NodeMap`, check if it's okay in the current state

consensus/src/units/store.rs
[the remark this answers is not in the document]
lesniak43: added `Display` to `NodeMap`

consensus/src/member.rs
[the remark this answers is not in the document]
timorl: Oh, then all the other ones should also be `..`, since you don't care about the number of fields there as well.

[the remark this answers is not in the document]
lesniak43: we need a style guide... or do we have one?...

consensus/src/runway/collection.rs
```diff
@@ -168,15 +168,26 @@ impl<'a, MK: KeyBox> Collection<'a, MK> {
...
             }
-            self.starting_round = max(self.starting_round, checked_unit.round() + 1);
+            None => 0,
+        };
+        let current_round = *self
+            .collected_starting_rounds
+            .get(response.responder)
+            .unwrap_or(&0);
+        if current_round != round {
+            info!(target: "AlephBFT-runway", "Warning: node {} responded with starting unit {}, but now says {}", response.responder.0, current_round, round);
```
timorl, line 187: I really don't think this is necessary, if you want something here I would put `warn` here, but only if `rounds < current_round`, because that is actually unexpected.

crypto/src/node.rs
```diff
@@ -81,6 +83,18 @@ impl<T> NodeMap<T> {
         NodeMap(v)
     }
 
+    pub fn with_hashmap(len: NodeCount, hashmap: HashMap<NodeIndex, T>) -> Self
```
timorl, line 86: Should probably be `from_hashmap` (unless clippy complains :/), afaik `with` implies using the thing for some behind-the-scenes stuff, like allocators for `Vec`s etc.

consensus/src/units/store.rs
```diff
@@ -28,6 +77,44 @@ impl<'a, H: Hasher, D: Data, KB: KeyBox> UnitStore<'a, H, D, KB> {
...
+                None => v.len() as u16,
+            }
+        }
+        let n_nodes: NodeCount = self.is_forker.size().into();
+        let gm = self
+            .by_coord
+            .keys()
+            .map(|c| (c.creator, c.round))
+            .into_grouping_map();
+        let top_row: HashMap<NodeIndex, Round> = gm.clone().max();
```
timorl, line 99: You can immediately make this a `NodeMap` using the `*_hashmap` function and this would also let you skip any type annotations here.

timorl: Actually same the line below.

consensus/src/units/store.rs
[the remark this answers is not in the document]
timorl: This was part of a different problem, not the `NodeMap` one. I thought about this a tad and lemme write the code I claim is better:
```
gm.
 .collect<BTreeSet<_>>()
 .into_iter()
 .filter_map(|(id, rounds)| rounds.into_iter().zip((0..)).position(|(round, index)| round != index).map(|round| (id, round)))
 .collect()
```

And then get rid of `first_missing`.

timorl (approved):
Mostly complaining about cosmetic stuff, other than that looks decent.

crypto/src/node.rs
lesniak43: possibly the best way would be to implement `Extend` and then construct the `NodeMap` like that: `NodeMap::with_size(len).extend(hashmap)` - what do you think?

consensus/src/runway/collection.rs
lesniak43: @DamianStraszak - what do you think?

consensus/src/member.rs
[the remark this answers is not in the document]
lesniak43: changed to `(_, _)` - it seems that we do have an "`_` everywhere" policy...

crypto/src/node.rs
timorl: That does look quite cool, but I would use a different name (maybe `fill`?) – `extend` doesn't make much sense for fixed size objects.

consensus/src/units/store.rs
lesniak43: fixed

consensus/src/units/store.rs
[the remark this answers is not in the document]
timorl: `BTreeSet` is sorted out of the box.

crypto/src/node.rs
lesniak43: but isn't the fixed "size" just an implementation detail? for me, `NodeMap` pretends to be a `HashMap` with additional constraints (other than the type) on keys?

have you seen any similar structs in other crates? e.g. `ArrayVec` and `FcHashMap` require the capacity to be known at compile time, so it's not quite it

then again, we probably _do_ know the number of nodes at compile time, don't we?

consensus/src/units/store.rs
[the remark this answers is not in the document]
lesniak43: then it's slower (at least in theory...), I don't like it!
@DamianStraszak - please be the judge

crypto/src/node.rs
timorl: We don't know it at compile time actually, only at object creation time. But after the object is created, its size cannot change, so `extend` doesn't make sense. In particular, this method should drop any thinks in the `NodeMap` that have indices associated with nodes outside of the acceptable range, which is not expected behaviour for `extend`.

This would be a perfect usecase for dependent types, but unfortunately we are not writing in Agda. :'<

lesniak43: I meant that we probably know how many committee members we have at compile time :P

And again, here "size" is somehow misleading, as it should be called "capacity". Both structs I've mentioned above have fixed capacity, but also implement `len` and `extend`. Imho `extend` makes perfect sense if `len` can change.

consensus/src/runway/collection.rs
DamianStraszak: You can either remove it, or make it a `warn` (and remove the inner `Warning`). I would say it's rather unexpected, to the extent that I prefer to keep it as an interesting bit of info.

crypto/src/node.rs
DamianStraszak: if `from_hashmap` is not possible then any of these is fine...

consensus/src/units/store.rs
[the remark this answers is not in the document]
DamianStraszak: Actually I find `first_missing` easier to understand than these kind of one-liners. Also this creates a heavy `BTreeSet` only to sort stuff.

DamianStraszak (approved):

crypto/src/node.rs
DamianStraszak: `len` cannot really change after a `NodeMap` object is created, so `extend` does not sound good to me.

lesniak43: ok, that's bad, because I've just added `len` that _can_ and _will_ change - please look at the implementation and give some feedback before I merge this PR

timorl: The whole abstraction here is "we are looking at something indexed by a set which we don't know at compile time, but at runtime is fixed". This doesn't quite map to what any of the structs you mentioned are trying to do, and I would argue that using language standardized for them ("capacity", "length", "extend" etc.) is likely to be confusing.

And no, we don't know the committee size at compile time, even now we could change it without recompiling anything and it definitely should stay that way...

consensus/src/runway/collection.rs
lesniak43: changed `info` to `warn`, keeping the condition unchanged

crypto/src/node.rs
lesniak43: also, `from_hashmap` is possible, but the point is that I'd personally expect this method to have only a single argument, whereas here we'd also need to provide the capacity

consensus/src/units/store.rs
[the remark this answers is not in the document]
lesniak43: @timorl - are you ok with this?

crypto/src/node.rs
lesniak43: > The whole abstraction here is "we are looking at something indexed by a set which we don't know at compile time, but at runtime is fixed". This doesn't quite map to what any of the structs you mentioned are trying to do

I totally agree - but then could you provide any better examples?

Also, I think that for a struct implementing `iter` any other implementation of `len` will be much more confusing...

lesniak43: (by "other" I mean "not equivalent to `.iter().count()`")

consensus/src/units/store.rs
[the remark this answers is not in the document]
timorl: How about
```
gm.
 .collect<HashSet<_>>()
 .into_iter()
 .filter_map(|(id, rounds)| (0..top_row.get(id)).position(|round| !rounds.contains(round)).map(|round| (id, round)))
 .collect()
```
?

I don't like `first_missing` mostly because such "purely algorithmic" code appearing in a codebase is usually a red flag (esp. with single letter variable names) and often hard to understand – the call itself is relatively easy to understand, but the code of the function raises questions. I would argue that while my previous attempt wasn't easy to understand the one above is about as simple as the `first_missing` version, especially without the additional `filter`.

consensus/src/runway/collection.rs
timorl: That's not in accordance with our logging standards – this `warn` can now appear during normal operation with no adversaries. From the AlephBFT point of view because we cannot be sure that the other node did not belatedly receive a higher unit and, more practically, from the aleph-node point of view because of data availability. I would say if `round > current_round` then this cannot be a warning, and preferably would be a `debug` at most, if not completely ignored (my preferred approach).

crypto/src/node.rs
timorl: If `from_hashmap` had only one argument it should be implemented as `From<HashMap>`, so I would say multiple args are exactly the usecase we want for a function with such a name.

`len` being equivalent to `iter().count` sounds like a reasonable policy to me.

consensus/src/runway/collection.rs
DamianStraszak: I'm ok with removing or turning into `debug`.

consensus/src/units/store.rs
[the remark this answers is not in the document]
DamianStraszak: Ok, when it comes to efficiency all this anyway uses `into_grouping_map()` so this is not making it so much slower. I'm ok with that version.

If we wanted this to actually cause low overhead (and note that for 100 nodes this will already be quite heavy) one would need to maintain `top_row` and `first_missing` directly in the store. You can easily update in amortized `O(1)` time per inserted unit. But it's not worth it at this point.

crypto/src/node.rs
DamianStraszak: I don't think you will find examples among other structs. `NodeMap` is special -- it represents the mathematical concept of a partial function [N] -> T. When creating a `NodeMap` you naturally need to specify `N` (a number) and `T` (a type). 

What I don't like is that we added implementations of `iter()`, `len()`, `size()` to it, and now these things are pairwise contradictory and not consistent with the usual intuition behind these names. I would get rid of these implementations and just use custom names like `ground_set_size()` or `iter_pairs` etc.

DamianStraszak: On top of that, I like `from_hashmap` with two args.

timorl: The most obvious replacement for `size` is `node_count`, I think we either have a trait for this or at least are using it very consistently around the codebase. For the rest I don't have strong opinions, but making the names more obviously unrelated to the common usage sounds good.

lesniak43: I'll go with `from_hashmap`, but I need to have `len` and `is_empty` (at least clippy was complaining about the second one) - @DamianStraszak how should I name those new methods then?

consensus/src/units/store.rs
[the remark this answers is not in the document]
lesniak43: fixed
now one could say that `first_missing` is... missing

consensus/src/runway/collection.rs
lesniak43: leaving this message as `debug`

consensus/src/runway/collection.rs
```diff
@@ -175,6 +179,9 @@ impl<'a, MK: KeyBox> Collection<'a, MK> {
                 return Err(Error::ForeignUnit(checked_unit.creator()));
             }
             self.starting_round = max(self.starting_round, checked_unit.round() + 1);
+            if let Some(round) = self.collected_responses.get_mut(response.responder.0) {
```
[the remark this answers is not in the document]
lesniak43: is it ok now? I've also changed `.unwrap_or(&0)` to `.unwrap_or(&round)`, so we don't see the debug logs too often..

consensus/src/units/store.rs
```diff
@@ -99,12 +87,12 @@ impl<'a, H: Hasher, D: Data, KB: KeyBox> UnitStore<'a, H, D, KB> {
...
             n_nodes,
-            gm.collect::<Vec<_>>()
+            gm.collect::<HashSet<_>>()
                 .into_iter()
-                .map(|(id, rounds)| (id, first_missing(rounds)))
-                .filter(|(id, round)| match top_row.get(*id) {
-                    Some(top_round) => round < top_round,
-                    None => false,
+                .filter_map(|(id, rounds)| {
+                    (0..*top_row.get(id).unwrap())
```
timorl, line 93: Oh no, not `unwrap`! D:

Wrap this in a `match top_row.get(id)` and `None => None,`? Not sure what clippy will have to say about that tho. <_<

consensus/src/units/store.rs
```diff
@@ -99,12 +87,12 @@ impl<'a, H: Hasher, D: Data, KB: KeyBox> UnitStore<'a, H, D, KB> {
...
+                        .position(|round| !rounds.contains(&round))
+                        .map(|round| (id, round as u16))
```
timorl, line 95: Isn't `round as Round` clearer?

crypto/src/node.rs
timorl: The version I'm seeing has both `len` and `is_empty`. <_<

timorl: Wait, nvm, this is `NodeSubset`. ^^"

consensus/src/units/store.rs
timorl: `unwrap_or(0)` would also be acceptable, although annoying implicit.

timorl (approved):
Looks good, only two tiny gripes.

consensus/src/units/store.rs
lesniak43: matched

consensus/src/units/store.rs
lesniak43: much clearer!

PR #228, opened 2022-06-22
Broadcast all top units

obrok:


obrok:
I wasn't sure what docs to update with this change. Can you point them out to me?

consensus/src/member.rs
```diff
@@ -70,8 +71,10 @@ enum Task<H: Hasher, D: Data, S: Signature> {
     CoordRequest(UnitCoord),
     // Request parents of the unit with the given hash and Recipient.
     ParentsRequest(H::Hash, Recipient),
-    // Broadcast the given unit.
+    // Broadcast the given unit. This happens only once when the unit is created.
     UnitMulticast(UncheckedSignedUnit<H, D, S>),
```
DamianStraszak, line 75: I think the existence of this task is controversial as it is performed only once and the idea of `Task` was rather to capture recurring tasks. Instead of implementing all the handling of this one-shot task, it's probably better to just send a broadcast right away in `on_create` and forget about it :P

consensus/src/member.rs
```diff
@@ -170,9 +190,17 @@ where
         resolved_requests: Receiver<Request<H>>,
     ) -> Self {
         let n_members = config.n_members;
+
+        let tasks = (0usize..n_members.into())
+            .filter(|node| NodeIndex(*node) != config.node_ix)
```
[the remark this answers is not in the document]
DamianStraszak: If we include our index we also need to make sure `top_units` is updated appropriately on `NewSelfUnit`.

consensus/src/member.rs
```diff
@@ -183,6 +211,7 @@ where
             notifications_from_runway,
             resolved_requests,
             exiting: false,
+            top_units: HashMap::with_capacity(n_members.into()),
```
DamianStraszak, line 214: We have a struct called `NodeMap` which should be convenient to use here.

DamianStraszak (changes requested):
A few small suggestions, but I like the changes in general.

consensus/src/member.rs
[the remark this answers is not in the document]
obrok: Good catch, this is unintentional. I think I did this when I was fiddling with the tests and suspected that broadcasting own unit too often was the cause of the problems. `NewAnyUnit` is generated on `NotificationOut::AddedToDag`, so I believe it should already happen for own units as well?

consensus/src/member.rs
obrok: :+1:

obrok:
@DamianStraszak I think I addressed all of your comments.

maciejnems (commented):
Very nice PR. I still have some thinking to do on this but I have one question.

consensus/src/runway/mod.rs
```diff
@@ -521,9 +526,15 @@ where
                 self.store.add_parents(h, p_hashes);
                 self.resolve_missing_parents(&h);
                 if let Some(su) = self.store.unit_by_hash(&h).cloned() {
+                    self.send_message_for_network(RunwayNotificationOut::NewAnyUnit(
```
DamianStraszak, line 529: I have thought about it for quite a bit and it now seems to me that this should be triggered not on `AddedToDag` but on each call to `self.store.add_unit`. The difference is that `AddedToDag` only happens after we have complete "history" of the unit, i.e. its parents and their parents, etc. so all its "lower cone". But `self.store.add_unit` is called whenever we get a new, valid unit. There is no reason why we should not rebroadcast units without full history.

For this reason I would like you to move sending this message from here to the (3 ?) places where `self.store.add_unit(_, _)` is called and send it from there. I think the best approach would be add a `bool` return value for `add_unit` to return a `bool` of whether the unit was added or was a duplicate (so that we can ignore duplicates).

DamianStraszak (commented):
Thanks for addressing the comments. I have added a few more -- also there is a discussion on slack (I marked you) that is kind of important for this PR, so please take a look, we must finish the discussion before this is merged.

consensus/src/runway/mod.rs
timorl: I would be wary of this approach, before a unit is properly inserted in the DAG it might be malicious in some way, and rebroadcasting it then might not be a good idea. In practice units get inserted quickly enough that this shouldn't matter for performance reasons, right?

DamianStraszak: For malicious units it shouldn't matter much, I think.
But my impression was that if we rebroadcast only DAG units then we violate the `ES` property even more... Maybe that's not true?

timorl: That would be quite strange, but maybe because I very much operate on the assumption that only in-DAG units are actually important (clearly for the consensus, but I'm not sure about other things). I cannot think of a specific attack with a malicious unit (create a unit of round 2137, and afterwards no one broadcasts your actual units?), but dropping that assumption seems strange to me.

DamianStraszak: Ok, I'm convinced. I think that by "induction" one should be able to argue that the set of all "honest" units will eventually be disseminated, because let's say `r` is the lowest round on which there is a unit `U` which has issues with dissemination, but then all units below it (having round `<=r-1`) have been disseminated correctly, hence they are in DAG, hence `U` itself in the DAG of the honest node holding it, and thus will be rebroadcast. 

@obrok -- you can ignore my above comment then, the current version is OK :P I'm leaving this discussion here. Feel free to resolve.

obrok: Heh, I already changed it, will revert.

timorl (approved):

DamianStraszak (approved):

maciejnems (approved):

PR #232, opened 2022-07-01
A0-1046: Rename KeyBox to Keychain

artemis-cc:


DamianStraszak (approved):

consensus/src/member.rs
```diff
@@ -555,7 +555,7 @@ pub async fn run_session<
     config: Config,
     local_io: LocalIO<D, DP, FH, US, UL>,
     network: N,
-    keybox: MK,
+    keychain: MK,
```
maciejnems, line 558: Maybe if that is keychain, we should change every generic from `MK` to `K`?

consensus/src/units/validator.rs
```diff
@@ -56,17 +56,17 @@ impl<H: Hasher, D: Data, S: Signature> From<SignatureError<FullUnit<H, D>, S>>
     }
 }
 
-pub struct Validator<'a, KB: KeyBox> {
+pub struct Validator<'a, KB: Keychain> {
```
maciejnems, line 59: Also instead of `KB` probably should be `K`

maciejnems (changes requested):

consensus/src/member.rs
maciejnems: Oh yeah, you are right. My bad.

maciejnems (approved):

PR #235, opened 2022-07-08
A0-847 fix false errors on termination

krzysztofziobro:
PR introduces a struct for a synchronized termination which helps with the problem of different components exiting at different times and using broken channels as a result.


krzysztofziobro:
One decision that I'm not sure about is whether we want to incorporate sending exits/creating exit channels into this new struct. What do you think?

consensus/src/terminator.rs
```diff
@@ -0,0 +1,121 @@
+use futures::channel::oneshot;
+use log::debug;
+
+pub type TerminatorConnection = (oneshot::Sender<()>, oneshot::Receiver<()>);
+pub type ShutdownConnection = (oneshot::Receiver<()>, Option<TerminatorConnection>);
```
timorl, line 5: A suggestion for a possible improvement, I won't insist, but I think it would make a lot of things look much cleaner.

I'm pretty sure you could add one more `struct` (say `ShutdownManager` for the rest of this comment, but feel free to come up with a cooler/more appropriate name) and work only on these two structs.

`ShutdownManager` would be passed to every component. It would have a method `async pub fn exit(&mut self) -> &mut Terminator` which would wait for an exit signal and return the terminator when it receives one – then the terminator could be used to `terminate_sync` just as now. It would also have a method `add_offspring(&mut self, component_name: &'static str) -> ShutdownManager` which would let you immediately pass the appropriate `ShutdownManager` to a subtask (internally it would obviously call the appropriate method on its `Terminator`). Crucially `Terminator` would now contain the send exit channels as well and `terminate_sync` would start by sending exit messages to all offspring.

This would eliminate the typedefs above which result in slightly mysterious deconstruction into pairs (someone complained when I did something like this recently, and I think this is a valid complaint) and explicit passing of exit channels. It would also make a lot of the exit code in various components cleaner.

timorl: Oh, I haven't noticed your comment that suggests something in this vein anyways, nice. ^^"

krzysztofziobro: Yes, I thought about hiding exits in Terminator, but haven't found a way to make passing channels to components look clean - this wrapper seems like a good idea. Will think it through and implement soon.

consensus/src/terminator.rs
```diff
@@ -0,0 +1,121 @@
...
+
+        // Make sure that all descendants recieved exit and won't be communicating with other components
+        for (receiver, name) in offspring_receivers {
+            if receiver.await.is_err() {
+                debug!(
+                    target: &self.component_name,
+                    "Terminator failed to receive from {}. Clean shutdown unsuccessful.",
+                    name,
+                );
+                return;
```
fixxxedpoint, line 65: Reconsider all of those `early exits` - perhaps it would be better to log all failed services. Probably requires some verification of whether it won't block forever.

fixxxedpoint (commented):
Really good work. Left some minor comments.

consensus/src/terminator.rs
```diff
@@ -22,24 +23,40 @@ impl Terminator {
...
+
         let (sender, offspring_recv) = oneshot::channel();
         let (offspring_sender, recv) = oneshot::channel();
 
         let endpoint = (sender, recv);
         let offspring_endpoint = (offspring_sender, offspring_recv);
 
         self.offspring_connections.insert(name, endpoint);
-        offspring_endpoint
+        (exit_recv, Some(offspring_endpoint))
```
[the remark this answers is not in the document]
timorl: Ah, yes, that's what I get for being overzelous, I'll update the review shortly...

consensus/src/terminator.rs
```diff
@@ -0,0 +1,141 @@
+use std::collections::HashMap;
+
+use futures::channel::oneshot::{channel, Receiver, Sender};
+use log::debug;
+
+type TerminatorConnection = (Sender<()>, Receiver<()>);
+type ShutdownConnection = (Receiver<()>, Option<TerminatorConnection>);
```
timorl, line 7: Now you only use it once, as an argument to a method, making the calls less readable, so preferably just remove and split the arguments.

consensus/src/terminator.rs
```diff
@@ -0,0 +1,141 @@
...
+pub struct Terminator {
+    component_name: &'static str,
+    parent_exit: Receiver<()>,
+    parent_connection: Option<TerminatorConnection>,
+    offspring_exits: HashMap<&'static str, Sender<()>>,
+    offspring_connections: HashMap<&'static str, TerminatorConnection>,
+}
+
+impl Terminator {
+    pub fn new(
```
timorl, line 20: Remove `pub`?

consensus/src/terminator.rs
```diff
@@ -0,0 +1,141 @@
...
+            component_name,
+            parent_exit: parent_shutdown_connection.0,
+            parent_connection: parent_shutdown_connection.1,
+            offspring_exits: HashMap::<_, _>::new(),
+            offspring_connections: HashMap::<_, _>::new(),
+        }
+    }
+
+    /// Creates a terminator for the root component
+    pub fn create_root_terminator(exit: Receiver<()>, name: &'static str) -> Self {
```
timorl, line 34: `create_root` should be ok, since you usually call this as `Terminator::create_root` anyway.

timorl (approved):
LGreatTM

consensus/src/terminator.rs
krzysztofziobro: Pretty sure it won't block - got rid of them.

consensus/src/terminator.rs
```diff
@@ -62,16 +59,15 @@ impl Terminator {
             "Terminator preparing for shutdown.",
         );
 
+        let mut offspring_senders = Vec::<_>::new();
+        let mut offspring_receivers = Vec::<_>::new();
```
timorl, line 63: More empty fishies.

consensus/src/testing/dag.rs
```diff
@@ -139,7 +139,7 @@ async fn run_consensus_on_dag(
             batch_tx,
             spawner.clone(),
             starting_round,
-            Terminator::create_root_terminator(exit_rx, "consensus"),
+            Terminator::create_root(exit_rx, "consensus"),
```
timorl, line 142: "AlephBFT-" for consistency?

timorl (approved):

consensus/src/terminator.rs
krzysztofziobro: Following our recent conversation with @fixxxedpoint: there are cases where it ends with early exits and blocks without them - it can happen when some component calls terminate_sync after it's child task exits with an error (which can block the system forever - at least in principle).

fixxxedpoint (commented):
Comments after last discussion:
- `get_exit` can be a method, not just returning a Receiver. It will allow to add some more logic when exit is triggered.
- suggestion (not very pretty): root_terminator can just hold `self` channel - in order to remove that Option
- consider early exit from `terminate_sync` if `get_exit` didn't finish before we call it - it should trigger a cascade of exits instead of potential deadlock if we call `terminate_sync`. Alternatively, `terminate_sync` might contain some mechanism allowing to warn other tasks/root of its early exit, to avoid potential deadlocks. This will allow to simplify usage of this api - it should always safe to call `terminate_sync`. Now calling it on some non-triggered exit can deadlock. Another option: `get_exit` returns something on what we can call `teminate_sync` - terminator doesn't have `terminate_sync` in its api directly.
- add few unit tests checking if this contract actually is implemented correctly - `spawn` (can be single threaded?) few tasks -> call `terminate_sync` from internal without calling `get_exit` -> success if exited

consensus/src/creation/mod.rs
```diff
@@ -69,7 +69,7 @@ async fn create_unit<H: Hasher>(
             },
             _ = exit => {
                 info!(target: "AlephBFT-creator", "Received exit signal.");
-                return Err(());
+                return Ok(None);
```
timorl, line 72: Oh no, this is definitely bad. If you really need to disambiguate the cases here, you should introduce another error type, but I suspect rebuilding the system to avoid deadlocks with `Terminator` as @fixxxedpoint suggested should be the better approach.

timorl:
>     * [...] Another option: `get_exit` returns something on what we can call `teminate_sync` - terminator doesn't have `terminate_sync` in its api directly.

I had a similar idea in one of my previous comments, the only problem is we have to (as you also mention above) be able to get this when we want to exit for internal reasons. So we would need two methods `get_exit` (as you describe it) plus `force_exit` returning the same thing, but probably working by sending some exit signal upwards (can a `oneshot` have multiple `Sender`s?) and then calling `get_exit`, waiting for the component above to start exiting.

Although now that I described it I see that this would mean, that any attempt at exiting should be propagated both ways... So `terminate_sync` attempts to send an exit message upwards, an at most `trace`s any errors, since in the usual setup this should fail?

Eh, this task is a trap every time someone tries solving the problem.

krzysztofziobro:
To be honest I don't think changing `get_exit` to a method which awaits for exit is the right call - then interface becomes highly nontrivial to use:
- If we just want to await for exit and set some flag that we will check at the beginning of `terminate_sync` for the same Terminator, then we have problems with borrowing (would need to do strange things, like putting selects in separate scope so the ref will drop before terminate_sync)
- If we want it to return something that we can call `terminate_sync` on, then we need this separate method (like your `force_exit`) that would be called at every termination caused by some error, also functions like `create_unit` in `creation/mod.rs` become problematic

One idea is that we could just check if exit was recieved at the beginning of `terminate_sync`:
- first option: just check what `self.parent_exit.is_terminated()` returns (this does not differentiate between `Ok(())` and `Canceled` on the exit channel, but we actually don't do that anywhere)
- second option: we make a wrapper for `oneshot::Reciever` that additionally keeps the info about reason of termination, and sync only if we actually received exit (for this we can implement FusedFuture and Pin, so there won't be problems with borrowing)

timorl (approved):

fixxxedpoint (approved):
lgtm

PR #246, opened 2022-07-29
A0-1045: Rewrite network in ordering example to UDP

obrok:
This solves the problem of establishing a connection with every network
send which seemed to blow up the system's capability to make more
connections in some circumstances. At the same time, the mock network
remains simple with no tracking of open connections and reconnection
attempts. A potential downside is that the UDP datagrams may be lost,
however this should rarely happen locally and the protocol should be
resistant to that in any case.

Some results:

* `./run.sh -n 15 -m 20 -c 6 -o 30 -d 15` - now passes while it hangs at some point on `main`
* `./run.sh -n 30 -m 5 -c 6 -o 30 -d 15` - still doesn't work (because of lack of censorship resistance, it seems)

### DIFF
--- a/examples/ordering/src/network.rs
+++ b/examples/ordering/src/network.rs
@@ -18,6 +18,9 @@ pub struct Network {
     my_id: usize,
     addresses: Vec<SocketAddr>,
     socket: UdpSocket,
+    /// Buffer for incoming data.
+    ///
+    /// It's allocated on the heap, because otherwise it overflows the stack when used inside a future.
     buffer: Box<[u8; MAX_UDP_DATAGRAM_BYTES]>,
 }
 
--- a/examples/ordering/src/network.rs
+++ b/examples/ordering/src/network.rs
@@ -3,19 +3,22 @@ use aleph_bft::{NodeIndex, Recipient};
 use aleph_bft_mock::{Hasher64, PartialMultisignature, Signature};
 use codec::{Decode, Encode};
 use log::error;
-use std::{io::Write, net::SocketAddr};
+use std::net::SocketAddr;
 use tokio::{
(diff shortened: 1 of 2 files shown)


examples/ordering/src/network.rs
```diff
@@ -3,19 +3,22 @@ use aleph_bft::{NodeIndex, Recipient};
...
+const MAX_UDP_DATAGRAM_BYTES: usize = 65536;
+
 pub type NetworkData = aleph_bft::NetworkData<Hasher64, Data, Signature, PartialMultisignature>;
 
 pub struct Network {
     my_id: usize,
     addresses: Vec<SocketAddr>,
-    listener: TcpListener,
+    socket: UdpSocket,
+    buffer: Box<[u8; MAX_UDP_DATAGRAM_BYTES]>,
```
DamianStraszak, line 21: Hmm, why do we need this `Box`? 
I see that we need it, as when I remove it, the thing compiles but then crashes, but why :P ?

DamianStraszak (approved):

examples/ordering/src/network.rs
obrok: I think what happens in that case is that there is a lot of state (65kb) for the futures to track in their state machine, so it blows up the stack quickly. When it's allocated on the heap, that's not a problem.

obrok: Another option would be to use a `Vec`, but I wanted to make the size explicit in the type so that there is never any error where the `Vec` is too small by accident.

obrok: Added a short comment with explanation.

lesniak43 (approved):

PR #247, opened 2022-07-29
A0-1200: Spawn member components

obrok:
I tried to spawn everything under member with the exception of the initial unit collection - this one seems like it only happens once, so it doesn't really matter?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -26,7 +26,7 @@ dependencies = [
 
 [[package]]
 name = "aleph-bft"
-version = "0.18.2"
+version = "0.19.0"
 dependencies = [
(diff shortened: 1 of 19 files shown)


lesniak43 (approved):
nice!

some nitpicks - please remove lifetimes from docs:

- `./crypto/src/signature.rs:263`
- `./crypto/src/signature.rs:296`
- `./rmc/src/lib.rs:163`

also I have a question - why do we need to pass cloned `keychain` instead of a reference?

obrok:
@lesniak43 lifetimes removed from docs.

DamianStraszak (approved):
good stuff :)

PR #255, opened 2022-08-09
A0-1253 Remove unnecessary wait selects from member and runway

krzysztofziobro:
Earlier we introduced selects which wait for unspawned tasks, in order to enable synchronized termination. Now, since all our components are spawned, this is no longer needed.

### DIFF
--- a/consensus/src/member.rs
+++ b/consensus/src/member.rs
@@ -1,4 +1,5 @@
 use crate::{
+    handle_task_termination,
     member::Task::{CoordRequest, ParentsRequest, RequestNewest, UnitBroadcast},
     network,
     runway::{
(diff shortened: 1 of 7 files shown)


fixxxedpoint (approved):
Good work!

consensus/src/member.rs
```diff
@@ -700,5 +683,19 @@ pub async fn run_session<
         debug!(target: "AlephBFT-member", "{:?} Network stopped.", index);
     }
 
+    if !runway_handle.is_terminated() {
+        if let Err(()) = runway_handle.await {
+            warn!(target: "AlephBFT-member", "{:?} Runway task stopped with an error", index);
+        }
+        debug!(target: "AlephBFT-member", "{:?} Runway stopped.", index);
+    }
```
obrok, line 691: Suggestion: consider extracting this as a function now? It's repeated 5-6 times, and I think it would be really nice to start chipping away at these multi-hundred-line functions that setup the components.

obrok (approved):

PR #266, opened 2022-09-21
A0-1398 Improve delays in unit coord requests in AlephBFT

obrok:
Changes how both CoordRequests and ParentRequests are managed:

CoordRequests:

1. CoordRequests are sent to a configurable number of random nodes
   (default 3, 3, 3, 1, 1, 1, 1, 1, ...).
2. Make the initial retries more aggressive with a default schedule of
   100, 1000, 3000, 6000, 9000, ... ms.

These changes aim at making the initial response faster in case the
network is spotty but not completely down.

ParentRequests:

ParentRequests are sent to a single random node from the first try
instead of the creator. Given that this mostly happens when we suspect
someone of cheating it might not even make sense to query the creator.


consensus/src/config.rs
```diff
@@ -78,27 +88,48 @@ pub fn exponential_slowdown(
...
-        unit_creation_delay,
-        // 5000, 500, 500, 500, ... (till step 3000), 500, 500*1.005, 500*(1.005)^2, 500*(1.005)^3, ..., 10742207 (last step)
-    };
     Config {
         node_ix,
         session_id,
         n_members,
-        delay_config,
+        delay_config: DelayConfig {
+            tick_interval: Duration::from_millis(100),
```
DamianStraszak, line 96: If I remember correctly the scheduling of tasks has precision `+/- tick_interval` hence we should probably decrease it to be like `10ms`, otherwise we might have delays of around `200ms` instead of the planned `100ms`.

consensus/src/config.rs
```diff
@@ -78,27 +88,48 @@ pub fn exponential_slowdown(
...
+        } else {
+            exponential_slowdown(t, 500.0, 3000, 1.005)
+        }
+    })
+}
+
+/// 100, 1000, 3000, 6000, 9000, ...
+fn default_coord_request_delay() -> DelaySchedule {
+    Arc::new(|t| {
+        if t <= 1 {
```
DamianStraszak, line 122: It seems to me that it is more like `100, 100, 1000, 3000, ...` -- because the first reschedule is with counter `0`?

fuzz/src/lib.rs
```diff
@@ -192,9 +192,12 @@ pub fn gen_config(node_ix: NodeIndex, n_members: NodeCount) -> Config {
         requests_interval: Duration::from_millis(50),
         unit_rebroadcast_interval_min: Duration::from_millis(400),
         unit_rebroadcast_interval_max: Duration::from_millis(500),
-        //100, 100, 300, 900, 2700, ...
-        unit_creation_delay: Arc::new(|t| exponential_slowdown(t, 50.0, usize::MAX, 1.000)),
         //50, 50, 50, 50, ...
+        unit_creation_delay: Arc::new(|t| exponential_slowdown(t, 50.0, usize::MAX, 1.000)),
```
DamianStraszak, line 196: This is a strange way to write `Arc::new(|_| Duration::from_millis(50)),` -- I know it's some old code, but you can change it on this occasion :P

consensus/src/member.rs
```diff
@@ -327,22 +331,25 @@ where
...
-    fn random_peer(&self) -> Recipient {
-        let node = rand::thread_rng()
-            .gen_range(0..self.n_members.into())
-            .into();
-        Recipient::Node(node)
+    fn random_peers(&self, n: usize) -> Vec<Recipient> {
+        self.peers
+            .iter()
+            .cloned()
+            .choose_multiple(&mut rand::thread_rng(), n)
```
DamianStraszak, line 352: This uses `choose_multiple` on iterators and not slices and thus has `O(self.peers.len())` complexity :P Iterators don't give random access to individual indices, hence it cannot be possibly better.  I guess you need to shuffle the ordering of the calls here to make it right.

DamianStraszak (approved):
Looks great, small comments only.

One idea I have is that maybe we should also do the same with `ParentRequest` what was done with `CoordRequest`, i.e., make delays and recpient numbers configurable. This would add consistency... What do you think @obrok ?

consensus/src/member.rs
obrok: :+1:

obrok:
@DamianStraszak I think I addressed all of your points.

consensus/src/config.rs
DamianStraszak: I think this is still unresolved?

lesniak43 (approved):

consensus/src/config.rs
obrok: I did decrease the timeouts to `0, 50, ...`, so it should in practice always be "on next tick". I think changing the default `tick_interval` has potentially far reaching consequences and would like to avoid it - going from `100` to `10` would basically mean the worker is woken up 10x more often, right? WDYT?

DamianStraszak: Btw. this is the default config, so it doesn't matter THAT MUCH, since it is not used in aleph-node anyway, but just wanted to emphasize that having some timeouts larger than `tick_interval` is probably not ideal.

PR #277, opened 2022-11-28
A0-1651: Type complexity check enabled

maciejzelaszczyk:
This enables the `type_complexity` check from `clippy`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,4 +73,4 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --all-targets --all-features -- -D warnings -A clippy::type_complexity
+          args: --all-targets --all-features -- -D warnings -D clippy::type_complexity
--- a/run_local_pipeline.sh
+++ b/run_local_pipeline.sh
@@ -2,6 +2,6 @@
 
 set -e
 
-cargo clippy --all-targets --all-features -- -D warnings -A clippy::type_complexity
+cargo clippy --all-targets --all-features -- -D warnings -D clippy::type_complexity
 cargo fmt --all
 cargo test --lib -- --skip medium


.github/workflows/ci.yml
```diff
@@ -73,4 +73,4 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --all-targets --all-features -- -D warnings -A clippy::type_complexity
+          args: --all-targets --all-features -- -D warnings -D clippy::type_complexity
```
timorl, line 76: I don't think you have to specify the `-D` flag here, `type_complexity` is enabled by default.

consensus/src/alerts/mod.rs
```diff
@@ -398,13 +406,19 @@ impl<'a, H: Hasher, D: Data, MK: MultiKeychain> Alerter<'a, H, D, MK> {
     }
 }
 
+pub(crate) type MessagesForNetwork<H, D, MK> = Sender<(
+    AlertMessage<H, D, <MK as Keychain>::Signature, <MK as MultiKeychain>::PartialMultisignature>,
```
timorl, line 410: Would it be enough to alias the inner type, instead of the whole `Sender`? It would be a bit cleaner. Same for `Receiver`.

consensus/src/runway/mod.rs
```diff
@@ -117,6 +120,19 @@ impl<H: Hasher, D: Data, S: Signature> TryFrom<UnitMessage<H, D, S>>
     }
 }
 
+type ResponsesForCollection<H, D, MK> = Sender<
+    UncheckedSigned<
```
timorl, line 124: Again, same as before.

consensus/src/runway/packer.rs
```diff
@@ -113,16 +113,16 @@ mod tests {
     const NODE_ID: NodeIndex = NodeIndex(0);
     const N_MEMBERS: NodeCount = NodeCount(4);
 
-    fn prepare(
-        keychain: Keychain,
-    ) -> (
+    type Preliminaries = (
```
timorl, line 116: Cute!

timorl: Although it would probably be nicer as a `struct`, then you could use `..` when unpacking it and the names would get standardized between tests, improving quality.

consensus/src/testing/creation.rs
```diff
@@ -63,15 +63,15 @@ impl TestController {
     }
 }
 
-fn setup_test(
-    n_members: NodeCount,
-) -> (
+type TestSetup = (
```
timorl, line 66: Same comment as above about a `struct`.

consensus/src/testing/mod.rs
```diff
@@ -65,18 +65,20 @@ pub fn gen_config(node_ix: NodeIndex, n_members: NodeCount) -> Config {
     }
 }
 
+type HonestMemeber = (
```
timorl, line 68: Same comment as above, although if it turns out that in this case it's a lot of work around very unrelated files, then maybe it can be skipped.

timorl: Eh, nvm, 4 objects are probably not worth it. Leaving comment in case you disagree.

consensus/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aleph-bft"
-version = "0.20.0"
+version = "0.21.0"
```
timorl, line 3: I think you haven't technically changed any interfaces, so this should be just a patch version bump. Same for all(?) other bumps.

timorl (changes requested):
Cool!

.github/workflows/ci.yml
maciejzelaszczyk: Agrred.

consensus/src/alerts/mod.rs
maciejzelaszczyk: Sure, done.

consensus/src/runway/collection.rs
```diff
@@ -212,11 +212,17 @@ impl<'a, MK: Keychain> Collection<'a, MK> {
     }
 }
 
+type ResponsesFromNetwork<H, D, MK> = Receiver<
+    UncheckedSigned<
```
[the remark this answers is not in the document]
maciejzelaszczyk: Changed.

consensus/src/runway/mod.rs
maciejzelaszczyk: Reworked.

consensus/src/testing/creation.rs
maciejzelaszczyk: Updated.

consensus/src/runway/mod.rs
```diff
@@ -117,6 +120,15 @@ impl<H: Hasher, D: Data, S: Signature> TryFrom<UnitMessage<H, D, S>>
     }
 }
 
+type ResponsesForCollection<H, D, MK> = UncheckedSigned<
```
timorl, line 123: Noe these are _definitely_ identical. :D

consensus/src/runway/packer.rs
```diff
@@ -133,25 +132,30 @@ mod tests {
             keychain,
             SESSION_ID,
         );
-        let (exit_tx, exit_rx) = oneshot::channel();
+        let (_, exit_rx) = oneshot::channel();
```
timorl, line 135: You need to keep the sender, otherwise everything immediately closes. <_<

There is a crucial and unintuitive difference between `_` and `_variable_name`.

timorl (approved):
Some minor problems, at least one of which should turn up in the pipelines anyway.

consensus/src/alerts/mod.rs
maciejzelaszczyk: No problem.

consensus/src/runway/packer.rs
maciejzelaszczyk: Rather important, thanks for catching this!

maciejzelaszczyk: Rather important, thanks for catching this!

consensus/src/alerts/mod.rs
```diff
@@ -406,18 +406,13 @@ impl<'a, H: Hasher, D: Data, MK: MultiKeychain> Alerter<'a, H, D, MK> {
     }
 }
 
-pub type MessagesForNetwork<H, D, MK> = (
-    AlertMessage<H, D, <MK as Keychain>::Signature, <MK as MultiKeychain>::PartialMultisignature>,
-    Recipient,
-);
-
-pub type MessagesFromNetwork<H, D, MK> =
+pub type NetworkMessages<H, D, MK> =
```
timorl, line 409: Agh, sorry I keep being nitpicky, but this probably should be singular not plural.

consensus/src/runway/mod.rs
```diff
@@ -120,11 +118,7 @@ impl<H: Hasher, D: Data, S: Signature> TryFrom<UnitMessage<H, D, S>>
     }
 }
 
-type ResponsesForCollection<H, D, MK> = UncheckedSigned<
-    NewestUnitResponse<H, D, <MK as Keychain>::Signature>,
-    <MK as Keychain>::Signature,
->;
-type ResponsesFromCollection<H, D, MK> = UncheckedSigned<
+type CollectionResponses<H, D, MK> = UncheckedSigned<
```
timorl, line 121: Same here. <_<"

timorl (approved):
Sorry, one more nit (in two instances). ^^"

consensus/src/alerts/mod.rs
maciejzelaszczyk: Naming is important, changed.

consensus/src/runway/mod.rs
maciejzelaszczyk: Changed here too.

PR #286, opened 2022-12-23
creator.rs

fixxxedpoint:
1. slightly refactored creator.rs: added `UnitsCollector`; added more detailed `trace` logging
2. setting initial delay to `30` minutes in creator if `can_create` flag is true - it should help in situations when we are still waiting for our own unit

### DIFF
--- a/consensus/Cargo.toml
+++ b/consensus/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aleph-bft"
-version = "0.20.4"
+version = "0.20.5"
 edition = "2021"
 authors = ["Cardinal Cryptography"]
 categories = ["algorithms", "data-structures", "cryptography", "database"]
--- a/consensus/Cargo.toml
+++ b/consensus/Cargo.toml
@@ -25,7 +25,7 @@ itertools = "0.10"
 log = "0.4"
 parking_lot = "0.12"
 rand = "0.8"
-thiserror = "1.0.38"
+thiserror = "1.0"
 
 [dev-dependencies]
 aleph-bft-mock = { path = "../mock" }
--- a/consensus/Cargo.toml
+++ b/consensus/Cargo.toml
@@ -15,6 +15,7 @@ description = "AlephBFT is an asynchronous and Byzantine fault tolerant consensu
 [dependencies]
 aleph-bft-rmc = { path = "../rmc", version = "0.6" }
 aleph-bft-types = { path = "../types", version = "0.8" }
+anyhow = "1.0"
 async-trait = "0.1"
 codec = { package = "parity-scale-codec", version = "3.0", default-features = false, features = ["derive"] }
 derivative = "2.2.0"
@@ -24,6 +25,7 @@ itertools = "0.10"
 log = "0.4"
 parking_lot = "0.12"
 rand = "0.8"
+thiserror = "1.0.38"
 
 [dev-dependencies]
 aleph-bft-mock = { path = "../mock" }
--- a/consensus/src/creation/creator.rs
+++ b/consensus/src/creation/creator.rs
@@ -3,7 +3,6 @@ use crate::{
     Hasher, NodeCount, NodeIndex, NodeMap, Round,
 };
 use anyhow::Result;
-use log::trace;
 use thiserror::Error;
 
 #[derive(Error, Debug)]
@@ -49,7 +48,6 @@ impl<H: Hasher> UnitsCollector<H> {
         if self.candidates.get(pid).is_none() {
             self.candidates.insert(pid, hash);
             self.n_candidates += NodeCount(1);
-            trace!(target: "AlephBFT-creator", "Added a new unit: {:#?}.", unit);
         }
     }
 
@@ -78,7 +76,6 @@ fn create_unit<H: Hasher>(
     let parent_hashes = parents.into_values().collect();
 
     let new_preunit = PreUnit::new(node_id, round, control_hash);
-    trace!(target: "AlephBFT-creator", "Created a new unit {:?} at round {:?}.", new_preunit, round);
     (new_preunit, parent_hashes)
 }
 
--- a/consensus/src/creation/creator.rs
+++ b/consensus/src/creation/creator.rs
@@ -17,7 +17,7 @@ enum ConstraintError {
 enum CreateError {
     #[error("Unfulfilled constraint: {0}")]
     MissingConstraint(ConstraintError),
-    #[error("Missing units collector for given round.")]
+    #[error("Missing units collector for given round: {0}.")]
     MissingCollector(Round),
 }
 
@@ -42,25 +42,25 @@ impl<H: Hasher> UnitsCollector<H> {
     }
 
     pub fn add_unit(&mut self, unit: &Unit<H>) {
-        let node_ix = unit.creator();
+        let node_id = unit.creator();
         let hash = unit.hash();
 
-        if self.candidates.get(node_ix).is_none() {
-            self.candidates.insert(node_ix, hash);
+        if self.candidates.get(node_id).is_none() {
+            self.candidates.insert(node_id, hash);
             self.n_candidates += NodeCount(1);
         }
     }
 
     pub fn prospective_parents(
         &self,
-        node_ix: NodeIndex,
+        node_id: NodeIndex,
     ) -> Result<&NodeMap<H::Hash>, ConstraintError> {
         let threshold = (self.candidates.size() * 2) / 3 + NodeCount(1);
 
         if self.n_candidates < threshold {
             return Err(ConstraintError::NotEnoughParents);
         }
-        if self.candidates.get(node_ix).is_none() {
+        if self.candidates.get(node_id).is_none() {
             return Err(ConstraintError::MissingOwnParent);
         }
         Ok(&self.candidates)
@@ -68,27 +68,27 @@ impl<H: Hasher> UnitsCollector<H> {
 }
 
 fn create_unit<H: Hasher>(
-    node_ix: NodeIndex,
+    node_id: NodeIndex,
     parents: NodeMap<H::Hash>,
     round: Round,
 ) -> (PreUnit<H>, Vec<H::Hash>) {
     let control_hash = ControlHash::new(&parents);
     let parent_hashes = parents.into_values().collect();
 
-    let new_preunit = PreUnit::new(node_ix, round, control_hash);
+    let new_preunit = PreUnit::new(node_id, round, control_hash);
     (new_preunit, parent_hashes)
 }
 
 pub struct Creator<H: Hasher> {
     round_collectors: Vec<UnitsCollector<H>>,
-    node_ix: NodeIndex,
+    node_id: NodeIndex,
     n_members: NodeCount,
 }
 
 impl<H: Hasher> Creator<H> {
-    pub fn new(node_ix: NodeIndex, n_members: NodeCount) -> Self {
+    pub fn new(node_id: NodeIndex, n_members: NodeCount) -> Self {
         Creator {
-            node_ix,
+            node_id,
             n_members,
             round_collectors: vec![UnitsCollector::new(n_members)],
         }
@@ -114,7 +114,7 @@ impl<H: Hasher> Creator<H> {
     pub fn create_unit(&self, round: Round) -> Result<(PreUnit<H>, Vec<H::Hash>)> {
         if round == 0 {
             let parents = NodeMap::with_size(self.n_members);
-            return Ok(create_unit(self.node_ix, parents, round));
+            return Ok(create_unit(self.node_id, parents, round));
         }
         let prev_round = usize::from(round - 1);
 
@@ -122,9 +122,9 @@ impl<H: Hasher> Creator<H> {
             .round_collectors
             .get(prev_round)
             .ok_or(CreateError::MissingCollector(round - 1))?
-            .prospective_parents(self.node_ix)?;
+            .prospective_parents(self.node_id)?;
 
-        Ok(create_unit(self.node_ix, parents.clone(), round))
+        Ok(create_unit(self.node_id, parents.clone(), round))
     }
 
     pub fn add_unit(&mut self, unit: &Unit<H>) {
--- a/consensus/src/creation/creator.rs
+++ b/consensus/src/creation/creator.rs
@@ -2,7 +2,31 @@ use crate::{
     units::{ControlHash, PreUnit, Unit},
     Hasher, NodeCount, NodeIndex, NodeMap, Round,
 };
+use anyhow::Result;
 use log::trace;
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+enum ConstraintError {
+    #[error("Not enough parents.")]
+    NotEnoughParents,
+    #[error("Missing own parent.")]
+    MissingOwnParent,
+}
+
+#[derive(Error, Debug)]
+enum CreateError {
+    #[error("Unfulfilled constraint: {0}")]
+    MissingContraint(ConstraintError),
+    #[error("Missing units collector for given pid.")]
+    MissingCollector,
+}
+
+impl From<ConstraintError> for CreateError {
+    fn from(value: ConstraintError) -> Self {
+        CreateError::MissingContraint(value)
+    }
+}
 
 #[derive(Clone)]
 struct UnitsCollector<H: Hasher> {
@@ -29,18 +53,19 @@ impl<H: Hasher> UnitsCollector<H> {
         }
     }
 
-    pub fn prospective_parents(&self, node_id: NodeIndex) -> Option<&NodeMap<H::Hash>> {
+    pub fn prospective_parents(
+        &self,
+        node_id: NodeIndex,
+    ) -> Result<&NodeMap<H::Hash>, ConstraintError> {
         let threshold = (self.candidates.size() * 2) / 3 + NodeCount(1);
 
         if self.n_candidates < threshold {
-            trace!(target: "AlephBFT-creator", "Unable to create a unit: not enough parents available.");
-            return None;
+            return Err(ConstraintError::NotEnoughParents);
         }
         if self.candidates.get(node_id).is_none() {
-            trace!(target: "AlephBFT-creator", "Unable to create a unit: missing our own unit in parents.");
-            return None;
+            return Err(ConstraintError::MissingOwnParent);
         }
-        Some(&self.candidates)
+        Ok(&self.candidates)
     }
 }
 
@@ -90,19 +115,20 @@ impl<H: Hasher> Creator<H> {
     /// Returns `None` if a unit cannot be created.
     /// To create a new unit, we need to have at least floor(2*N/3) + 1 parents available in previous round.
     /// Additionally, our unit from previous round must be available.
-    pub fn create_unit(&self, round: Round) -> Option<(PreUnit<H>, Vec<H::Hash>)> {
+    pub fn create_unit(&self, round: Round) -> Result<(PreUnit<H>, Vec<H::Hash>)> {
         if round == 0 {
             let parents = NodeMap::with_size(self.n_members);
-            return create_unit(self.node_id, parents, round).into();
+            return Ok(create_unit(self.node_id, parents, round));
         }
         let prev_round = usize::from(round - 1);
 
         let parents = self
             .round_collectors
-            .get(prev_round)?
+            .get(prev_round)
+            .ok_or(CreateError::MissingCollector)?
             .prospective_parents(self.node_id)?;
 
-        create_unit(self.node_id, parents.clone(), round).into()
+        Ok(create_unit(self.node_id, parents.clone(), round))
     }
 
     pub fn add_unit(&mut self, unit: &Unit<H>) {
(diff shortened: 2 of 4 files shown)


consensus/src/creation/mod.rs
```diff
@@ -55,16 +55,23 @@ async fn create_unit<H: Hasher>(
     round: Round,
     creator: &mut Creator<H>,
     create_lag: &DelaySchedule,
-    mut can_create: bool,
+    mut try_create: bool,
```
DamianStraszak, line 58: Both `can_create` and `try_create` are actually slightly confusing. This bool captures whether we need to wait for the delay to pass or not. A better name would be `delay_passed` or something similar.

DamianStraszak: Ok, after thinking a little more... This is wrong because we use a single bool for two different things. What should be done:
1) Call this bool `skip_delay` and remove `mut`
2) Create a new bool in the function body `delay_passed`.
3) Adjust the rest of the code inside the function.

fixxxedpoint: How about passing explicitly `initial_delay` instead of that flag: `create_lag` if normal, 0 if catching up, and move create as last step then?

DamianStraszak: I don't think this will do what we want. The problem here is that there should be really two functions:
1) create_unit_no_earlier_than_delay
2) create_unit_asap

And we are trying to capture both with one function.

consensus/src/creation/creator.rs
```diff
@@ -4,85 +4,109 @@ use crate::{
...
+            candidates: NodeMap::with_size(n_members),
+        }
+    }
+
+    pub fn add_unit(&mut self, unit: &Unit<H>) {
+        let pid = unit.creator();
+        let hash = unit.hash();
+
+        if self.candidates.get(pid).is_none() {
+            self.candidates.insert(pid, hash);
```
kostekIV, line 26: not a comment to this pr but I feel like we should check whether the pid is in range of candidates size (or change the signature of nodemap to return bool whether insert worked)

consensus/src/creation/mod.rs
```diff
@@ -75,11 +82,11 @@ async fn create_unit<H: Hasher>(
                 }
             },
             _ = &mut delay => {
-                if can_create {
-                    warn!(target: "AlephBFT-creator", "Delay passed despite us not waiting for it.");
+                if try_create {
+                    warn!(target: "AlephBFT-creator", "Delay passed at round {} despite us not waiting for it.", &round);
```
kostekIV, line 86: arent we waiting for it in both cases?

consensus/src/creation/creator.rs
```diff
@@ -4,85 +4,109 @@ use crate::{
...
+            self.n_candidates += NodeCount(1);
+            trace!(target: "AlephBFT-creator", "Added a new unit: {:#?}.", unit);
+        }
+    }
+
+    pub fn can_create(&self, node_id: NodeIndex) -> Option<&NodeMap<H::Hash>> {
```
kostekIV, line 32: this could return an Err with variants like: `NotEnoughParents` & `MissingOwnUnit`. wdyt?

kostekIV (commented):
I like the new UnitsCollector :)

consensus/src/creation/mod.rs
```diff
@@ -55,16 +55,23 @@ async fn create_unit<H: Hasher>(
...
     create_lag: &DelaySchedule,
-    mut can_create: bool,
+    mut try_create: bool,
     incoming_parents: &mut Receiver<Unit<H>>,
     mut exit: &mut oneshot::Receiver<()>,
 ) -> Result<(PreUnit<H>, Vec<H::Hash>), ()> {
-    let mut delay = Delay::new(create_lag(round.into())).fuse();
+    const THIRTY_MINUTES: Duration = Duration::from_secs(30 * 60);
+
+    let initial_delay = match try_create {
```
timorl, line 64: I would rather follow @DamianStraszak 's solution and then it's somewhat more natural to set the initial delay to zero if `skip_delay` is true.

This would also remove the new named constant, although maybe it should stay... anyway, if it stays I would suggest a different name for it, maybe something like `VERY_LONG_TIME`, 'cause the whole idea is that this should never pass after the first time, and the fact that it's set to 30mins is only a bit of a coincidence.

consensus/src/creation/mod.rs
```diff
@@ -55,16 +55,23 @@ async fn create_unit<H: Hasher>(
...
+        true => THIRTY_MINUTES,
+    };
+    let mut delay = Delay::new(initial_delay).fuse();
     loop {
-        if can_create {
+        if try_create {
             if let Some(result) = creator.create_unit(round) {
                 return Ok(result);
             }
+            debug!(target: "AlephBFT-creator", "Creator unable to create a new unit at round {:?}.", round);
```
timorl, line 74: If you follow @kostekIV 's suggestion about errors, this would be a very nice place to print them.

timorl (changes requested):
Very cool changes.

consensus/src/creation/creator.rs
fixxxedpoint: I can do that, but it's not like we can do anything with that information except logging it, so why bother with a custom error. :)

consensus/src/creation/creator.rs
```diff
@@ -4,85 +4,109 @@ use crate::{
...
-        let parent_hashes = parents.into_values().collect();
+        let parents = self
+            .round_collectors
+            .get(prev_round)?
+            .can_create(self.node_id)?;
 
-        let new_preunit = PreUnit::new(self.node_id, round, control_hash);
-        trace!(target: "AlephBFT-creator", "Created a new unit {:?} at round {:?}.", new_preunit, round);
-        Some((new_preunit, parent_hashes))
+        create_unit(self.node_id, parents.clone(), round).into()
```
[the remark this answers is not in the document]
fixxxedpoint: Where is the `compose` operator?!

consensus/src/creation/creator.rs
kostekIV: I feel like it makes it easier to log it at the callee side (like @timorl suggested) plus we never know if we won't use it later:
NotEnoughParents -> waitForParents
MissingOwnUnit -> waitForOwnUnits :P.
Also it looks better to return result here rather than option IMHO :P

fixxxedpoint: I also prefer it to return something more descriptive than `Option`. Just not a big fun of breaking KISS unnecessarily. My counterargument: errors should be used for recoverable situations, not just logging + those obvious values are going to change, but it's creating useless new dependency on them + it's not an error if we asked too early creator, so `Result` is not the correct one here. Any better suggestions?

consensus/src/creation/mod.rs
fixxxedpoint: Now it should be less confusing.

consensus/src/creation/creator.rs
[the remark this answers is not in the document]
fixxxedpoint: Just like the potential of `.into()` when we decide to switch it to `Result` ;)

consensus/src/creation/mod.rs
timorl: I strongly prefer a zero delay over an `Option`, especially because of the match at the beginning of the method becoming completely unnecessary.

consensus/src/creation/creator.rs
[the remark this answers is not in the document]
timorl: Then we will change it to `Ok`. :P We never use `into` in this manner in any other parts of code, and in general I would rather we avoid it when the alternative is a short wrapper. An `into` suggests there is something more mysterious happening than just a `Some`/`Ok`.

consensus/src/creation/creator.rs
DamianStraszak: You can change the name of the function, if you want this to look more like an error (like `try_create`). 
Alternatively, you can return one custom enum which has 3 options.

I'm also OK with the current version, but I think the code is going to evolve soon and more precise return types might actually help in the rewrite.

timorl: I feel like two of the arguments you present contradict each other. This is exactly a recoverable error, because creation doesn't have to succeed in this context, we recover from it by logging and ignoring the error.

If we use errors, we can have one fewer logline, we can avoid using logging in `creator.rs` completely (for the price of using `Display`), and we can have more precise tests. Plus, when we rewrite unit creation these two errors will likely remain and we will just add a couple more, if anything.

consensus/src/creation/mod.rs
```diff
@@ -51,20 +51,28 @@ pub struct IO<H: Hasher> {
...
     creator: &mut Creator<H>,
-    create_lag: &DelaySchedule,
-    mut can_create: bool,
+    delay: Option<Delay>,
     incoming_parents: &mut Receiver<Unit<H>>,
     mut exit: &mut oneshot::Receiver<()>,
 ) -> Result<(PreUnit<H>, Vec<H::Hash>), ()> {
-    let mut delay = Delay::new(create_lag(round.into())).fuse();
+    let (initial_delay, mut delay_passed) = match delay {
+        None => (very_long_delay(), true),
```
DamianStraszak, line 66: I don't like it at all now, it's longer and more confusing than before.
Two bools and computing the delay inside this function is better.
The best is imho to stop pretending that avoiding code duplication with this artificial delay is cool. Just make two similar functions, one for the case when we wait for the delay, or consider two cases with very similar `select`s.

DamianStraszak (changes requested):

consensus/Cargo.toml
```diff
@@ -24,6 +25,7 @@ itertools = "0.10"
 log = "0.4"
 parking_lot = "0.12"
 rand = "0.8"
+thiserror = "1.0.38"
```
timorl, line 28: We shouldn't be explicitly specifying the patch version.

Also, we don't use `anyhow`/`thiserror` anywhere else although we likely should. We were also trying to minimize the dependencies of ABFT, but these two are likely fine... So yeah, this second paragraph is only informational I guess, unless someone wants to oppose adding the deps.

timorl (approved):

consensus/src/creation/creator.rs
fixxxedpoint: I've decided to use `anyhow::Result` with a non public error (used only for logging, not leaking anything internal to creator).

kostekIV (approved):

consensus/src/creation/mod.rs
```diff
@@ -104,69 +139,95 @@ async fn create_unit<H: Hasher>(
...
-            Ok((u, ph)) => (u, ph),
-            Err(_) => {
-                terminator.terminate_sync().await;
-                return;
-            }
-        };
+        // NOTE: even we've observed a unit from a higher round, our own unit from previous round
+        // might not be yet added to `creator`. We still might need to wait for its arrival.
+        let should_delay = !(creator.current_round() > round);
+        if should_delay {
```
timorl, line 209: The if could have the condition directly in it, without the additional variable.

I'm also pretty sure you shouldn't add the "NOTE" now, nothing in the code currently suggests we expect to be able to create a unit immediately, just that we will not be observing the artificial delay.

At the same time the comment above might require a slight rework – now we are waiting unless we are behind. Hm, not sure I am expressing this clearly, if it looks good to you leave it as-is.

timorl (approved):

consensus/src/creation/mod.rs
```diff
@@ -104,69 +139,95 @@ async fn create_unit<H: Hasher>(
...
+            CreatorError::ParentsChannelClosed => {
+                debug!(target: "AlephBFT-creator", "Incoming parent channel closed, exiting.")
+            }
+        }
+    }
+}
+
+async fn run_creator<H: Hasher>(
+    conf: Config,
+    io: &mut IO<H>,
```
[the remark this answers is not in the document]
fixxxedpoint: This is for the purpose of the `terminator`. Channels we receive from outside should be kept till `terminate_sync` returns, otherwise other modules might start complaining about closed channels. There are some more details about that `terminator` here which are not clear - I will talk with its author.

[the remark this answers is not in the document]
timorl: Oh, right, damn.

consensus/src/creation/mod.rs
fixxxedpoint: IMHO it's more human-friendly :) to read a bit more verbose version, but I am ok with removing it. Maybe just change that variable to `skip_delay` and leave the comment intact.

timorl: In general I agree with more descriptions, but in this case this "debunked" a misconception the reader was unlikely to develop in the current version of the code, so it would confuse them until they figured out why it might have been an issue in the first place. So, increasing cognitive load instead of decreasing it.

I'm less sure about the variable, I see why it might be nice to have such semi-useless variables for readability sometimes and this is a borderline case I guess.

consensus/src/creation/mod.rs
```diff
@@ -104,69 +126,91 @@ async fn create_unit<H: Hasher>(
...
-                return;
-            }
-        };
+        let skip_delay = creator.current_round() > round;
+        if !skip_delay {
+            let lag = Delay::new(create_lag(round.into()));
+
+            keep_processing_units_until(&mut creator, incoming_parents, lag)
+                .await
+                .map_err(|_| CreatorError::ParentsChannelClosed)?;
```
timorl, line 199: This `map_err` isd no longer necessary. :D

timorl (approved):
You missed one `map_err`, other than that very cool. <3

consensus/src/creation/mod.rs
fixxxedpoint: current version: we have two separate functions now, `keep_processing_units_until` and `create_unit`. The former, retrieves new units from provided channel until some given timeout. The latter, is a loop `try create unit -> if failed download one more unit`. This version is more verbose. What do you think?

consensus/src/creation/mod.rs
fixxxedpoint: See the comment above.

consensus/src/creation/creator.rs
```diff
@@ -2,87 +2,134 @@ use crate::{
...
+    NotEnoughParents,
+    #[error("Missing own parent.")]
+    MissingOwnParent,
+}
+
+#[derive(Error, Debug)]
+enum CreateError {
+    #[error("Unfulfilled constraint: {0}")]
+    MissingContraint(ConstraintError),
+    #[error("Missing units collector for given node_index.")]
```
timorl, line 20: "...for given session." actually.

timorl (approved):
Some minor error problems.

consensus/src/creation/mod.rs
```diff
@@ -104,69 +126,89 @@ async fn create_unit<H: Hasher>(
 /// Section 5.1 for a discussion of this component.
 pub async fn run<H: Hasher>(
     conf: Config,
-    io: IO<H>,
+    mut io: IO<H>,
     mut starting_round: oneshot::Receiver<Option<Round>>,
     mut terminator: Terminator,
 ) {
+    futures::select! {
+        _ = read_starting_round_and_run_creator(conf, &mut io, &mut starting_round).fuse() =>
```
DamianStraszak, line 134: isn't it better to pass just `starting_round` instead of a mut reference?

consensus/src/creation/mod.rs
```diff
@@ -104,69 +126,89 @@ async fn create_unit<H: Hasher>(
...
+    }
+}
+
+async fn run_creator<H: Hasher>(
+    conf: Config,
+    io: &mut IO<H>,
+    starting_round: Round,
+) -> anyhow::Result<(), CreatorError> {
     let Config {
         node_id,
```
DamianStraszak, line 179: Would be good to change this to `node_ix` as well. And all other occurences.

consensus/src/creation/creator.rs
```diff
@@ -2,87 +2,134 @@ use crate::{
...
+        &mut self.round_collectors[round]
     }
 
     /// Returns `None` if a unit cannot be created.
     /// To create a new unit, we need to have at least floor(2*N/3) + 1 parents available in previous round.
     /// Additionally, our unit from previous round must be available.
-    pub fn create_unit(&self, round: Round) -> Option<(PreUnit<H>, Vec<H::Hash>)> {
-        if !self.can_create(round) {
-            return None;
+    pub fn create_unit(&self, round: Round) -> Result<(PreUnit<H>, Vec<H::Hash>)> {
```
DamianStraszak, line 115: I don't like the signature of this function vs how it is used (repeatedly called in a loop). This suggests that we are getting errors all the time, even though these "errors" are fully expected. Also the `MissingCollector` is the only true error -- this happening would mean there is a critical bug in the implementation, which should be distinguishable from the remaining results. 
I propose the following:
- call this function `try_create_preunit`
- make its output type `Result<CreationAttempt>`
- the `CreationAttempt` would be an enum with variants 
`Ready(PreUnit<H>, Vec<H::Hash>)`
`WaitingForEnoughParents`
`WaitingForOwnParent`
- the only error this could return is `MissingCollector`
- in the place where this function is called, log the error on `warn` or `error` level instead of `debug`, but don't log  `WaitingForEnoughParents` or `WaitingForOwnParent`, because they are not errors.

consensus/src/creation/creator.rs
DamianStraszak: "...for given round." actually xD

DamianStraszak (changes requested):
Great PR. There are only some small details that I would love to see adjusted.

consensus/src/creation/mod.rs
timorl: Or the other way around, we use `node_id` almost everywhere else...

consensus/src/creation/creator.rs
timorl: Derp, right. xd

consensus/src/creation/mod.rs
fixxxedpoint: This is due the `Terminator` - `starting_round` shouldn't be dropped until `terminator.terminate_sync().await` returns.

consensus/src/creation/creator.rs
```diff
@@ -2,87 +2,134 @@ use crate::{
...
+    NotEnoughParents,
+    #[error("Missing own parent.")]
+    MissingOwnParent,
+}
+
+#[derive(Error, Debug)]
+enum CreateError {
+    #[error("Unfulfilled constraint: {0}")]
+    MissingConstraint(ConstraintError),
+    #[error("Missing units collector for given round.")]
```
timorl, line 20: Shouldn't this be "for round {0}." now?

timorl (approved):

consensus/src/creation/creator.rs
```diff
@@ -2,87 +2,134 @@ use crate::{
...
-    // initializes the vectors corresponding to the given round (and all between if not there)
-    fn init_round(&mut self, round: Round) {
-        if round > self.current_round() {
-            let new_size = (round + 1).into();
-            self.candidates_by_round
-                .resize(new_size, NodeMap::with_size(self.n_members));
-            self.n_candidates_by_round.resize(new_size, NodeCount(0));
-        }
+    // gets or initializes unit collectors for a given round (and all between if not there)
+    fn get_collector_for_round(&mut self, round: Round) -> &mut UnitsCollector<H> {
```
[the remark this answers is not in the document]
fixxxedpoint: How about naming it `get_or_initialize_collector_for_round`, calling it only in `add_unit` and using explicitly `current_round` in it, instead of `...len() -1` in that condition?

consensus/src/creation/mod.rs
fixxxedpoint: now it's `node_id`

consensus/src/creation/creator.rs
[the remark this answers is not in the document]
DamianStraszak: Yes, that sounds good.

consensus/src/creation/mod.rs
```diff
@@ -54,41 +68,58 @@ pub struct IO<H: Hasher> {
...
-    let mut delay = Delay::new(create_lag(round.into())).fuse();
+) -> Result<(PreUnit<H>, Vec<H::Hash>), CreatorError> {
     loop {
-        if can_create {
-            if let Some(result) = creator.create_unit(round) {
-                return Ok(result);
+        match creator.create_unit(round) {
+            Ok(unit) => return Ok(unit),
+            Err(err) => {
+                debug!(target: "AlephBFT-creator", "Creator unable to create a new unit at round {:?}: {}.", round, err)
```
DamianStraszak, line 77: This should be `trace` as it spits one log message per unit, and it's by no means an error. Is it possible to use `Display` for `round` (i.e. `{}` instead of `{:?}`). If yes, please do, because the log is then shorter.

consensus/src/creation/creator.rs
DamianStraszak: I think the current version is fine. Of course what you propose makes sense, but I would need to give it more thought in order to tell if it's worth it... Anyway, feel free to resolve this comment.

DamianStraszak (approved):

consensus/src/creation/mod.rs
fixxxedpoint: done

PR #303, opened 2023-03-29
A0-1054: Simplify mock Saver and Loader.

woocash2:
* Removed unnecessary cloning a `vec` under a `mutex`
* Refactored the `Saver::new` function to take no arguments.
* Implemented `From` trait for `Saver` which creates a Saver from existing data (same behavior as ::new before the refactor)

### DIFF
--- a/consensus/src/testing/mod.rs
+++ b/consensus/src/testing/mod.rs
@@ -86,7 +86,7 @@ pub fn spawn_honest_member(
     let spawner_inner = spawner;
     let unit_loader = Loader::new(units);
     let saved_state = Arc::new(Mutex::new(vec![]));
-    let unit_saver = Saver::from_data(saved_state.clone());
+    let unit_saver = Saver::from(saved_state.clone());
     let local_io = LocalIO::new(data_provider, finalization_handler, unit_saver, unit_loader);
     let member_task = async move {
         let keychain = Keychain::new(n_members, node_index);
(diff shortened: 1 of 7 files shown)


DamianStraszak (approved):
Good stuff :+1:

lesniak43 (approved):

consensus/src/testing/mod.rs
```diff
@@ -86,7 +86,7 @@ pub fn spawn_honest_member(
     let spawner_inner = spawner;
     let unit_loader = Loader::new(units);
     let saved_state = Arc::new(Mutex::new(vec![]));
-    let unit_saver = Saver::new(saved_state.clone());
+    let unit_saver = Saver::from(saved_state.clone());
```
timorl, line 89: Have you tried `.into()` here?

timorl: Or, even better – if I understand what's happening here correctly then `new` could be returning this `Arc<...>` and you don't have to have the `From` at all. In other places you would just let `(unit_saver, _) = Saver::new()` and here you would have `saved_state` in the second position.

timorl (approved):
Tiny comments, LGTM otherwise.

consensus/src/testing/mod.rs
woocash2: Yeah, `(unit_saver, _) = Saver::new()` would work, but I don't know, `::from` looks more intuitive to me.
`into()` is also possible instead. When we use `::from` its easy to track calls in the IDE but I can see that `.into()` is used more often in the repo.

PR #304, opened 2023-03-31
A0-1078: Extract checking version bump to a separate action

mikogs:


.github/actions/check-version-bumped/action.yml
```diff
@@ -0,0 +1,20 @@
...
+    description: "Path to Cargo.toml"
+    type: string
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Check
+      run: |
+        if ! git diff HEAD origin/main -- ${{ inputs.cargo-toml-path }} | grep -q '^+version ='; then
```
Marcin-Radecki, line 16: It's better to escape paths with double quotes `"` when dealing with paths in scripts, as at some point they might contain spaces, so I suggest here `"${{ inputs.cargo-toml-path }}"`

Marcin-Radecki (approved):

lesniak43 (approved):

PR #310, opened 2023-05-15
A0-2504: More intuitive and flexible tests for unit backup.

woocash2:
Extracted from the function `prepare_test` all the logic related to:
* selecting which units should appear in the backup source,
* simulating ABFT to create units,
* composing unit encodings to form the backup source 
* corrupting the backup source.

So that the function `prepare_test` receives the already prepared backup source as an argument, and its only responsibility is to prepare the loading mechanism, and return appropriate handles and channel endpoints. As a result:
* testing backup is more flexible - with extracted logic it is possible test new different scenarios,
* test cases are more readable - they prepare backup sources explicitly by themselves.

### DIFF
--- a/consensus/src/runway/backup.rs
+++ b/consensus/src/runway/backup.rs
@@ -249,10 +249,9 @@ mod tests {
             }
 
             let mut unchecked_signed_units = Vec::with_capacity(pre_units.len());
-            for (i, (pre_unit, _)) in pre_units.into_iter().enumerate() {
-                unchecked_signed_units.push(
-                    preunit_to_unchecked_signed_unit(pre_unit, session_id, &keychains[i]).await,
-                )
+            for ((pre_unit, _), keychain) in pre_units.into_iter().zip(keychains.iter()) {
+                unchecked_signed_units
+                    .push(preunit_to_unchecked_signed_unit(pre_unit, session_id, keychain).await)
             }
 
             units_per_round.push(unchecked_signed_units);
--- a/consensus/src/runway/backup.rs
+++ b/consensus/src/runway/backup.rs
@@ -275,12 +275,8 @@ mod tests {
         units.iter().map(|u| u.encode()).collect()
     }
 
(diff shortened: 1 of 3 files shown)


consensus/src/runway/backup.rs
```diff
@@ -229,78 +229,70 @@ mod tests {
...
+            let units: Vec<_> = pre_units
+                .iter()
+                .map(|(pre_unit, _)| preunit_to_unit(pre_unit.clone(), session_id))
+                .collect();
+            for creator in creators.iter_mut() {
+                creator.add_units(&units);
+            }
+
+            let mut unchecked_signed_units = Vec::with_capacity(pre_units.len());
+            for (i, (pre_unit, _)) in pre_units.into_iter().enumerate() {
```
timorl, line 252: `pre_units.into_iter().zip(keychains.iter())` should also work and not require explicit indexing.

timorl (approved):
Looks good, one tiny nit that might even just be a personal preference, so feel free to argue if you want to keep the current version.

consensus/src/runway/backup.rs
woocash2: Yep, that looks nicer.

consensus/src/runway/backup.rs
```diff
@@ -229,78 +229,69 @@ mod tests {
...
-                round,
-                ammount: 1,
-                corrupted: false,
-            }
+    fn encode_all(units: Vec<UncheckedSignedUnit>) -> Vec<Vec<u8>> {
+        units.iter().map(|u| u.encode()).collect()
+    }
+
+    fn concatenate_encodings(mut unit_encodings: Vec<Vec<u8>>) -> Vec<u8> {
+        let mut full_encoding = Vec::new();
```
fixxxedpoint, line 279:
```suggestion
        unit_encodings.into_iter().flatten().collect()
```

fixxxedpoint (approved):
lgtm

consensus/src/runway/backup.rs
woocash2: Much nicer, changed.

PR #312, opened 2023-06-09
A0-2583: Prevent reaching max round too fast

woocash2:
Add a function which can be used to check whether reaching the maximum round takes large enough amount of time. The function is not called in AlephBFT as various projects which use aleph-bft may expect different delays. Instead the function is publically exposed to be called in the project which uses AlephBFT.

Tests:
* checked that `ordering` and `blockchain` examples work well
* checked that `aleph-node` (after needed changes) works with this version of AlephBFT (but turned out we also need to make `default_delay_config` public)

### DIFF
--- a/consensus/src/config.rs
+++ b/consensus/src/config.rs
@@ -162,6 +162,21 @@ pub fn default_config(
     )
 }
 
+/// Creates a [`DelayConfig`] with default parameters, suggested by the creators of this package.
+pub fn default_delay_config() -> DelayConfig {
+    DelayConfig {
+        tick_interval: Duration::from_millis(10),
+        unit_rebroadcast_interval_min: Duration::from_millis(15000),
+        unit_rebroadcast_interval_max: Duration::from_millis(20000),
+        unit_creation_delay: default_unit_creation_delay(),
+        coord_request_delay: default_coord_request_delay(),
+        coord_request_recipients: default_coord_request_recipients(),
+        parent_request_delay: Arc::new(|_| Duration::from_millis(3000)),
+        parent_request_recipients: Arc::new(|_| 1),
+        newest_request_delay: Arc::new(|_| Duration::from_millis(3000)),
+    }
+}
+
 /// 5000, 500, 500, 500, ... (till step 3000), 500, 500*1.005, 500*(1.005)^2, 500*(1.005)^3, ..., 10742207 (last step)
 fn default_unit_creation_delay() -> DelaySchedule {
     Arc::new(|t| match t {
@@ -183,20 +198,6 @@ fn default_coord_request_delay() -> DelaySchedule {
 /// 3, 3, 3, 1, 1, 1, 1, ...
 fn default_coord_request_recipients() -> RecipientCountSchedule {
     Arc::new(|t| if t <= 2 { 3 } else { 1 })
-}
-
-fn default_delay_config() -> DelayConfig {
-    DelayConfig {
-        tick_interval: Duration::from_millis(10),
-        unit_rebroadcast_interval_min: Duration::from_millis(15000),
-        unit_rebroadcast_interval_max: Duration::from_millis(20000),
-        unit_creation_delay: default_unit_creation_delay(),
-        coord_request_delay: default_coord_request_delay(),
-        coord_request_recipients: default_coord_request_recipients(),
-        parent_request_delay: Arc::new(|_| Duration::from_millis(3000)),
-        parent_request_recipients: Arc::new(|_| 1),
-        newest_request_delay: Arc::new(|_| Duration::from_millis(3000)),
-    }
 }
 
 fn time_to_reach_round(round: Round, delay_schedule: &DelaySchedule) -> Duration {
--- a/consensus/src/config.rs
+++ b/consensus/src/config.rs
@@ -1,9 +1,13 @@
 use crate::{NodeCount, NodeIndex, Round, SessionId};
+use log::error;
 use std::{
     fmt::{Debug, Formatter},
     sync::Arc,
     time::Duration,
 };
+
+#[derive(Debug)]
+pub struct InvalidConfigError;
 
 /// A function answering the question of how long to delay the n-th retry.
 pub type DelaySchedule = Arc<dyn Fn(usize) -> Duration + Sync + Send + 'static>;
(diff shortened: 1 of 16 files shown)


consensus/src/config.rs
```diff
@@ -113,8 +113,26 @@ pub fn default_config(n_members: NodeCount, node_ix: NodeIndex, session_id: Sess
...
+    let mut total_time = Duration::from_millis(0);
+    for r in 0..round {
+        total_time += delay_schedule(r as usize);
+    }
+    total_time
+}
+
+/// Reaching a high round can be expected to introduce some slowdown.
+/// Use this function to ensure that the slowdown will happen.
+pub fn reaching_round_takes_at_least(
```
timorl, line 126: This is fine, but I would suggest also adding something like `reaching_max_round_takes_at_least(&self, duration: Duration)` to `Config`, to make checking this even more convenient for users.

consensus/src/config.rs
```diff
@@ -113,8 +113,26 @@ pub fn default_config(n_members: NodeCount, node_ix: NodeIndex, session_id: Sess
...
+    round: Round,
+    delay_schedule: DelaySchedule,
+    duration: Duration,
+) -> bool {
+    time_to_reach_round(round, delay_schedule) >= duration
+}
+
 /// 5000, 500, 500, 500, ... (till step 3000), 500, 500*1.005, 500*(1.005)^2, 500*(1.005)^3, ..., 10742207 (last step)
-fn default_unit_creation_delay() -> DelaySchedule {
+pub fn default_unit_creation_delay() -> DelaySchedule {
```
timorl, line 135: Why did you make this `pub`?

timorl (approved):
Cool.

consensus/src/config.rs
```diff
@@ -72,6 +72,14 @@ pub struct Config {
     pub max_round: Round,
 }
 
+impl Config {
+    /// Reaching a high round can be expected to introduce some slowdown.
+    /// Use this function to ensure that the slowdown will happen.
+    pub fn reaching_max_round_takes_at_least(&self, duration: Duration) -> bool {
```
DamianStraszak, line 78: Since this is optional to use, it's very easy to skip or forget this check.

The bug we had in aleph-node was actually that we prepared a correct config, but in a dead code, and then ended up using a default one, and not the prepared one. And I'm pretty sure that it's possible to repeat this bug, by not being careful...

What I would propose to do, is to either:
1) Make sure that this check is always done, for instance in a constructor of `Config`, or
2) Introduce a new type `VerifiedConfig` which you get from `Config` by applying this check (and perhaps in the future also different ones), and make AlephBFT initialization expect `VerifiedConfig`.

DamianStraszak (changes requested):

consensus/src/config.rs
timorl: Out of these I prefer 1. since it doesn't create an additional type, although it will make the config annoyingly opaque compared to others – the fields will probably no longer be `pub`, since we don't want it to be constructable without the check.

On the other hand I hate `pub` fields, so maybe I should take this as a win. ;p

consensus/src/config.rs
```diff
@@ -61,23 +65,15 @@ impl Debug for DelayConfig {
 #[derive(Clone, Debug)]
 pub struct Config {
     /// Identification number of the Member=0,..,(n_members-1).
-    pub node_ix: NodeIndex,
+    pub(crate) node_ix: NodeIndex,
```
timorl, line 68: Augh, I hate `pub(crate)`s with a passion... I think I would prefer even `pub`s, despite the possibility of "evading" the constructor then.

consensus/src/config.rs
```diff
@@ -100,25 +96,47 @@ pub fn exponential_slowdown(
...
+/// Creates a [`Config`], allowing the user to omit specifying `max_round` and `delay_config`
+/// in which case they will be set do default values, suggested by the creators of this package.
+/// `time_to_reach_max_round` is a lower bound on the time needed to reach the maximum round
+/// expected by the user and is only used for verification.
+pub fn create_config(
+    n_members: NodeCount,
+    node_ix: NodeIndex,
+    session_id: SessionId,
+    max_round: Option<Round>,
+    delay_config: Option<DelayConfig>,
```
timorl, line 108: This isn't a great interface, if you want to be able to construct something with default values then write a separate constructor and document what guarantees it gives (in particular what the approximate `time_to_reach_max_round` is – it makes little sense to pass this as an argument if you are using the default). Then this constructor can always require these arguments.

consensus/src/config.rs
```diff
@@ -100,25 +96,47 @@ pub fn exponential_slowdown(
...
+    time_to_reach_max_round: Duration,
+) -> Result<Config, InvalidConfigError> {
+    let max_round = max_round.unwrap_or(5000);
+
+    let delay_config = delay_config.unwrap_or(DelayConfig {
```
timorl, line 113: Is this really the easiest way of creating a `DelayConfig`?

timorl (changes requested):
Some complaints about the interfaces.

consensus/src/config.rs
woocash2: But then we kind of miss the point of forcing the check. We could make them private but it would require changing each field access to a getter call.

consensus/src/config.rs
woocash2: It was initially done like that, and I haven't seen any function like `default_delay_config` or so. But we can consider wrapping this code in such a function.

consensus/src/config.rs
woocash2: Makes sense. I propose 2 functions: `create_config` which has an argument for each field of `Config` and `create_config_with_default_delays` which skips the `delay_config` argument. Each of these would take `time_to_reach_max_round` and perform the check.

DamianStraszak: I think there are 3 types of configs that we will ever need to create:
1. Config for clueless users: this needs to take `n_members`, `node_ix` and `session_id`. The rest should be some sensible defaults. We can also include a check that `time_to_reach_max_round=7days` works (so that we never mess up the defaults inside of AlephBFT).  
2. Config for us, `aleph_node`: we should be able (and actually should be forced) to customize every single field, this should also require us to specify `time_to_reach_max_round` explicitly. We really want to avoid the situation like before that we use the default config, but then change a single field which then gets us into trouble. 
3. Test configs.

I guess the 2 constructors that you propose should cover well all 3 cases.

timorl: I would just suggest `default_config` as a less unwieldy name for the second one.

consensus/src/config.rs
timorl: Yeah, it would be nice for the creation of delay configs to be better designed, especially since we want to do that more manually now. Some sensible constructor(s?) for it, and perhaps even for various pieces of it.

consensus/src/config.rs
timorl: It tells you very little about the interface surface – it's essentially `pub` for the purpose of reasoning about where stuff can be used, but pretends it isn't. We had a lot of instances where someone wanted to meddle in the internals of something in a different part of the code, but didn't want to expose it externally and they would use `pub(crate)` then. In such cases it's much better to either (1) redesign the interface to make more sense for external usage, or (2) restructure the code in such a way that the places that need access to internals get it naturally. Failing that, (3) a more restrictive access grant (e.g. `pub(super)`) is acceptable.

In this case I looked through the code, and unfortunately we use `Config` in quite a couple of places, sometimes quite deep, always assuming the fields are `pub`. Either this should be refactored (solution 2; probably too much work for this PR, but good to keep in mind for future refactors) or they all need to keep the access (solution 3 is not feasible :/). I concede that this is probably the most justified usecase for `pub(crate)` I have seen, so if you really want to ensure nobody creates a config manually you can leave it as is. ._.

timorl (approved):

PR #314, opened 2023-08-04
A0-2846: aleph-bft shouldn't quit after creator's graceful exit

fixxxedpoint:
# Description

Creator of aleph-bft might decide to gracefully exit when it discovers inconsistency of its units backup compared to knowledge collected from network. In that case, it should be allowed to continue instead of panicking.

## Type of change

- New feature (non-breaking change which adds functionality): allow to disable creator
- Fix: channel returned by Terminator was already of type `FusedStream`. When used with `futures::select!` it could cause that select to panic, since its implementation returns `is_terminated() == true` before it return an error (after we drop parents Terminator).


consensus/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aleph-bft"
-version = "0.23.0"
```
kostekIV, line 3: hmmm, prolly 0.24

kostekIV (approved):
nice one. LGTM

lesniak43 (approved):
cool

consensus/Cargo.toml
fixxxedpoint: bumped

PR #315, opened 2023-08-05
A0-2884: Refactor alerter

woocash2:
Refactor alerter to make it possible for synchronous simulation.

* Separate the `alerts` module into synchronous handler and service with a single async method (`Alerter` is now `Handler` and `Service` contains the `run` logic)
* Make possible to create `Service` and `Handler` before running the service

**A delicate change**: changed signature of `rmc` and `io` to not take `keychain` by reference but by value. This was needed because `keychain` is now passed to `Service::new` where it is passed further to `Handler` (old `Alerter`) and `IO`. It can't be passed by reference to them because of the lifetime issues. Instead we pass cloned `keychain` to both of them by value.

If we really want `keychain` to remain passed by reference we can probably implement it by using `Rc`.


consensus/src/alerts/handler.rs
```diff
@@ -0,0 +1,752 @@
...
+
+    pub fn who_is_forking(&self, proof: &ForkProof<H, D, MK::Signature>) -> Option<NodeIndex> {
+        let (u1, u2) = proof;
+        let (u1, u2) = {
+            let u1 = u1.clone().check(&self.keychain);
+            let u2 = u2.clone().check(&self.keychain);
+            match (u1, u2) {
+                (Ok(u1), Ok(u2)) => (u1, u2),
+                _ => {
+                    warn!(target: "AlephBFT-alerter", "{:?} Invalid signatures in a proof.", self.index());
```
timorl, line 108: Similar comment as above, `Result` instead of `Option`. Not sure whether this should be the same error type or two separate tho.

consensus/src/alerts/handler.rs
```diff
@@ -0,0 +1,752 @@
...
+            if rounds.contains(&full_unit.round()) {
+                warn!(target: "AlephBFT-alerter", "{:?} Two or more alerted units have the same round {:?}.", self.index(), full_unit.round());
+                return false;
+            }
+            rounds.insert(full_unit.round());
+        }
+        true
+    }
+
+    pub fn who_is_forking(&self, proof: &ForkProof<H, D, MK::Signature>) -> Option<NodeIndex> {
```
timorl, line 100: Should this even be `pub`? I'm pretty sure not all the methods here should, please un`pub` the ones that are not used externally.

consensus/src/alerts/handler.rs
```diff
@@ -0,0 +1,752 @@
...
+    ) -> NetworkAlert<H, D, MK> {
+        let alert = match alert.check(&self.keychain) {
+            Ok(alert) => alert,
+            Err(e) => {
+                warn!(target: "AlephBFT-alerter","{:?} We have received an incorrectly signed alert: {:?}.", self.index(), e);
+                return None;
+            }
+        };
+        let contents = alert.as_signable();
+        if let Some(forker) = self.who_is_forking(&contents.proof) {
```
timorl, line 177: Errors here as well plx, and this should let you get away with a `?` here. Maybe the `Option` will stay though, as at least one of the cases doesn't have a `warn`.

consensus/src/alerts/handler.rs
```diff
@@ -0,0 +1,752 @@
...
+
+    /// `on_message()` may return an `AlerterResponse` which should be propagated
+    pub fn on_message(
+        &mut self,
+        message: AlertMessage<H, D, MK::Signature, MK::PartialMultisignature>,
+    ) -> Option<AlerterResponse<H, D, MK::Signature, MK::PartialMultisignature>> {
+        use AlertMessage::*;
+        match message {
+            ForkAlert(alert) => {
+                trace!(target: "AlephBFT-alerter", "{:?} Fork alert received {:?}.", self.index(), alert);
```
timorl, line 206: Eh, this can probably also be a more generic thing in `Service`.

consensus/src/alerts/handler.rs
```diff
@@ -0,0 +1,752 @@
...
+                    ))
+                }
+            }
+            AlertRequest(node, hash) => match self.known_alerts.get(&hash) {
+                Some(alert) => Some(AlerterResponse::ForkAlert(
+                    alert.clone().into_unchecked(),
+                    Recipient::Node(node),
+                )),
+                None => {
+                    debug!(target: "AlephBFT-alerter", "{:?} Received request for unknown alert.", self.index());
```
timorl, line 232: Hm, less sure whether this should also be an error, but perhaps? Please look into this wile changing the other things.

consensus/src/alerts/handler.rs
```diff
@@ -0,0 +1,752 @@
...
+
+    /// `alert_confirmed()` may return a `ForkingNotification`, which should be propagated
+    pub fn alert_confirmed(
+        &mut self,
+        multisigned: Multisigned<H::Hash, MK>,
+    ) -> Option<ForkingNotification<H, D, MK::Signature>> {
+        let alert = match self.known_alerts.get(multisigned.as_signable()) {
+            Some(alert) => alert.as_signable(),
+            None => {
+                error!(target: "AlephBFT-alerter", "{:?} Completed an RMC for an unknown alert.", self.index());
```
timorl, line 247: Also errors plx.

consensus/src/runway/mod.rs
```diff
@@ -885,17 +886,18 @@ pub(crate) async fn run<H, D, US, UL, MK, DP, FH, SH>(
     let alerter_keychain = keychain.clone();
     let alert_messages_for_network = network_io.alert_messages_for_network;
     let alert_messages_from_network = network_io.alert_messages_from_network;
+
+    let mut alerter_service = Service::new(
```
timorl, line 890: I'm pretty sure you actually want to be able to separately construct the `Handler` and use it for any "simulations", rather than the `Service`. You don't want to be pushing into IO channels during this preprocessing. This would also let you have `Service` with only `run` (and maybe `new`) public.

rmc/src/lib.rs
```diff
@@ -184,21 +184,21 @@ impl<T: Send + Sync + Clone> TaskScheduler<T> for DoublingDelayScheduler<T> {
...
     network_tx: UnboundedSender<Message<H, MK::Signature, MK::PartialMultisignature>>,
-    keychain: &'a MK,
+    keychain: MK,
     scheduler: Box<dyn TaskScheduler<Task<H, MK>>>,
     multisigned_hashes_tx: UnboundedSender<Multisigned<H, MK>>,
     multisigned_hashes_rx: UnboundedReceiver<Multisigned<H, MK>>,
 }
 
-impl<'a, H: Signable + Hash + Eq + Clone + Debug, MK: MultiKeychain> ReliableMulticast<'a, H, MK> {
+impl<'a, H: Signable + Hash + Eq + Clone + Debug, MK: MultiKeychain> ReliableMulticast<H, MK> {
```
timorl, line 197: Why is the `'a` still here?

And in general thanks for removing it everywhere, it was a bit of a leftover from a botched cleanup anyway. :P

timorl (changes requested):
You should be able to construct the `Handler` and use that for the simulation rather than `Service`. Plus some other comments about the general architecture of this kind of split, mostly that logging should happen in the `Service`.

woocash2:
> you're right, so it's already possible to create `Handler` and call its methods, but I suppose that some component for simulation and message aggregation is needed.

consensus/src/alerts/handler.rs
woocash2: we will probably need a `Result<Option<AlerterResponse>, Error>`. The first `None` I treat as no error so we would return `Ok(None)` and the last `None` would be some kind of error.

consensus/src/alerts/handler.rs
```diff
@@ -0,0 +1,785 @@
...
+
+impl<H: Hasher, D: Data, MK: MultiKeychain> Handler<H, D, MK> {
+    pub fn new(keychain: MK, config: AlertConfig) -> Self {
+        Self {
+            session_id: config.session_id,
+            keychain,
+            known_forkers: HashMap::new(),
+            known_alerts: HashMap::new(),
+            known_rmcs: HashMap::new(),
+            exiting: false,
```
timorl, line 96: Uh, I'm pretty sure this should just be in `Service`.

consensus/src/alerts/handler.rs
```diff
@@ -0,0 +1,785 @@
...
+        }
+    }
+
+    pub fn index(&self) -> NodeIndex {
+        self.keychain.index()
+    }
+
+    pub fn is_forker(&self, forker: NodeIndex) -> bool {
```
timorl, line 104: Another superfluous `pub`?

consensus/src/alerts/handler.rs
```diff
@@ -0,0 +1,785 @@
...
+    }
+
+    // Correctness rules:
+    // 1) All units must be created by forker
+    // 2) All units must come from different rounds
+    // 3) There must be fewer of them than the maximum defined in the configuration.
+    // Note that these units will have to be validated before being used in the consensus.
+    // This is alright, if someone uses their alert to commit to incorrect units it's their own
+    // problem.
+    pub fn verify_commitment(&self, alert: &Alert<H, D, MK::Signature>) -> Result<(), Error> {
```
timorl, line 123: Also not needed externally?

consensus/src/alerts/handler.rs
```diff
@@ -0,0 +1,785 @@
...
+            return Err(Error::WrongCreator(alert.sender));
+        }
+        if full_unit1.round() != full_unit2.round() {
+            return Err(Error::DifferentRounds(alert.sender));
+        }
+        Ok(full_unit1.creator())
+    }
+
+    /// `rmc_alert()` registers the RMC but does not actually send it; the returned hash must be passed to `start_rmc()` separately
+    pub fn rmc_alert(
```
timorl, line 171: Also?

consensus/src/alerts/handler.rs
```diff
@@ -0,0 +1,785 @@
...
+
+    /// `on_message()` may return an `AlerterResponse` which should be propagated
+    pub fn on_message(
+        &mut self,
+        message: AlertMessage<H, D, MK::Signature, MK::PartialMultisignature>,
+    ) -> OnMessageResult<H, D, MK> {
+        use AlertMessage::*;
+        match message {
+            ForkAlert(alert) => {
+                // trace!(target: "AlephBFT-alerter", "{:?} Fork alert received {:?}.", self.index(), alert);
```
timorl, line 236: Probably just remove this? Definitely don't leave it as a comment. :P

consensus/src/alerts/service.rs
```diff
@@ -0,0 +1,148 @@
...
+use log::{debug, error, warn};
+use std::time;
+
+pub struct Service<H: Hasher, D: Data, MK: MultiKeychain> {
+    handler: Handler<H, D, MK>,
+    io: IO<H, D, MK>,
+}
+
+impl<H: Hasher, D: Data, MK: MultiKeychain> Service<H, D, MK> {
+    pub fn new(
```
timorl, line 19: I know you want to do the simulations in a separate PR (rightly so!), but this is still weird. I would suggest more something like:
1. Merge `Service` and `IO` (probably using the former name) to contain all the IO, but no `Handler`.
2. Have a `run` method that takes the `Handler` as an argument.
3. Create both in `member` and (for now) just pass one to the other.

This should make it simpler to do the simulation in the next PR, but if done properly should also remove the need for `alerter_index` (since you'll have access to the `Handler`), as well as remove the `pub`s in `IO`s fields. In the future if we also refactor `Rmc` it would also likely be simpler to slot it in here.

consensus/src/alerts/service.rs
```diff
@@ -0,0 +1,148 @@
...
+            }
+            Ok(Some(AlerterResponse::ForkResponse(maybe_notification, hash))) => {
+                self.io.rmc.start_rmc(hash);
+                if let Some(notification) = maybe_notification {
+                    self.io
+                        .send_notification_for_units(notification, &mut self.handler.exiting);
+                }
+            }
+            Ok(None) => {}
+            Err(error) => debug!(target: "AlephBFT-alerter", "{}", error),
```
timorl, line 83: Since you are touching the code here please also do the `LOG_TARGET` thingy we do in `aleph-node`.

consensus/src/alerts/service.rs
timorl: Oh, and `exiting` shouldn't need to be passed as an argument (although `Handler` might have to sometimes).

timorl (changes requested):
Still some comments.

consensus/src/alerts/service.rs
woocash2: how about passing `Handler` to `Service::new`?

consensus/src/alerts/service.rs
```diff
@@ -0,0 +1,215 @@
...
+            warn!(
+                target: LOG_TARGET,
+                "{:?} Channel with notifications for network should be open",
+                self.keychain.index()
+            );
+            self.exiting = true;
+        }
+    }
+
+    // methods for event handling
```
fixxxedpoint, line 108: Unnecessary comment.

fixxxedpoint (approved):
LGTM. Suggestion: consider implementing a more direct approach instead of `simulation`. I don't know how exactly your interactions with `Handler` will look like but in general a safer approach would be a more direct interaction with it, e.g. add a new constructor to `Handler` that initializes its internal state properly based on persisted state. This way it should be more safe if somebody decides to change `Handler`'s internals. With the simulation approach, you risk ending in some `undefined` state (since it's not a `full` simulation - imagine someone add some more steps to protocol, or it buffers something internally, etc.). Using directly a new constructor assures you (kind of) that the Handler is in correct state after initialization.

consensus/src/alerts/handler.rs
```diff
@@ -0,0 +1,786 @@
...
+            }
+            if rounds.contains(&full_unit.round()) {
+                return Err(Error::SameRound(full_unit.round(), alert.sender));
+            }
+            rounds.insert(full_unit.round());
+        }
+        Ok(())
+    }
+
+    fn verify_fork(&self, alert: &Alert<H, D, MK::Signature>) -> Result<NodeIndex, Error> {
```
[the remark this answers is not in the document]
woocash2: You're probably right. I'd say the same about `verify_commitment`. But these methods can return `handler::Error` and it would be weird to me if `Alert` depended on this error type. We could move `handler::Error` along with these methods to `alerts/mod` but this doesn't seem ideal to me as I see it fit much better inside `handler`.

woocash2:
> LGTM. Suggestion: consider implementing a more direct approach instead of `simulation`. I don't know how exactly your interactions with `Handler` will look like but in general a safer approach would be a more direct interaction with it, e.g. add a new constructor to `Handler` that initializes its internal state properly based on persisted state. This way it should be more safe if somebody decides to change `Handler`'s internals. With the simulation approach, you risk ending in some `undefined` state (since it's not a `full` simulation - imagine someone add some more steps to protocol, or it buffers something internally, etc.). Using directly a new constructor assures you (kind of) that the Handler is in correct state after initialization.

I think it is possible. Our goal is to obtain a handler which witnessed some events in a particular order, and a Service with some of its IO channels partially filled, also depending on the sequence of events.

We can have `Handler::new` which creates the `Handler` in a particular state, and returns the sequence of messages to be passed to `Service::new` which will fill its channels itself. 

That would be nicer indeed.

consensus/src/alerts/handler.rs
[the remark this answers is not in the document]
fixxxedpoint: Maybe we just leave it as is in this PR.

fixxxedpoint (approved):

consensus/src/alerts/service.rs
```diff
@@ -0,0 +1,209 @@
...
+
+pub struct Service<H: Hasher, D: Data, MK: MultiKeychain> {
+    messages_for_network: Sender<(NetworkMessage<H, D, MK>, Recipient)>,
+    messages_from_network: Receiver<NetworkMessage<H, D, MK>>,
+    notifications_for_units: Sender<ForkingNotification<H, D, MK::Signature>>,
+    alerts_from_units: Receiver<Alert<H, D, MK::Signature>>,
+    rmc: ReliableMulticast<H::Hash, MK>,
+    messages_for_rmc: Sender<RmcMessage<H::Hash, MK::Signature, MK::PartialMultisignature>>,
+    messages_from_rmc: Receiver<RmcMessage<H::Hash, MK::Signature, MK::PartialMultisignature>>,
+    keychain: MK,
```
timorl, line 22: You actually only use the index, so maybe just save that?

timorl (approved):
The `exiting` thing feels wrong, I suspect it should rather be an error propagated outwards, but I really don't want to make you rewrite this again so I guess it can stay.

consensus/src/alerts/service.rs
woocash2: fax

PR #330, opened 2023-08-25
return the creator of a unit with the unit data when the unit is finalized

joschisan:
This pr is supposed to simplify the integration of Aleph Bft into Fedimint, a byzantine fault tolerant decentralised  chaumian ecash mint we call a Federation. Fedimint uses Aleph to order incoming transactions submitted by the clients. The inputs and outputs of the transaction are either chaumian ecash notes, on-chain bitcoin or smart contracts that enable integration with the bitcoin lightning network enabling instant settlement. After a transaction has been ordered by Aleph the system has to achieve consensus wether a transaction is valid by validating all inputs and outputs individually. 

An on-chain bitcoin input is the output of a bitcoin transaction that sends money to the Federation and is at least six blocks deep in the blockchain. Therefore we need to achieve consensus on the current block height of the bitcoin blockchain to validate the on-chain bitcoin input.

An on-chain bitcoin output is a bitcoin transaction that the federation threshold signs that sends money from the Federation to the clients requested bitcoin address. To assure that the  bitcoin transaction will be included into the bitcoin blockchain the Federation has to add a high enough fee to the transaction which is also paid by the user. Therefore, to validate an on-chain output of a Fedimint transaction the Federation has to achieve consensus on the current ferrate necessary.

Finally, Lightning outputs of a Fedimint transaction are so called Hash Timelocked Contracts or HTLCs. To check wether a time lock has passed the system has to achieve consensus on the current time.

To achieve consensus on an integer like block height or ferrate to submit additional types of consensus items to the broadcast alongside the transactions from the clients. In our examples a node would submit its local ferrate and block height to the broadcast which we call votes. We remember the last such vote for every node and then recalculate the median to achieve consensus every time a vote is updated. Therefore we need to know which node submitted a vote. The naive solution would be to sign every vote we want to submit with the keychain before submitting it to aleph bft. While this is not an issue with strictly increasing values such as block height in case of the ferrate however there could be attacks by resubmitting an old vote by a different node; this is generally known as replay attacks. Hence we would have to sign an increasing integer for every submitted vote and save the integer corresponding to the last vote to prevent such attacks. This introduces unnecessary complexity as the extra information that needs to be returned is trivially available.

Furthermore, Fedimint is built as an extensible system that can be extended with additional functionality by third parties in the form of new input and output types as well as new consensus items that are needed to validate these inputs and outputs. One extension currently in development is called stability pools which allows clients to enter smart contracts with financial service providers to hedge against falling dollar prices of bitcoin. To enforce those contracts the federation needs to achieve consensus on the current dollar value of bitcoin.

### DIFF
--- a/types/src/dataio.rs
+++ b/types/src/dataio.rs
@@ -1,5 +1,7 @@
 use async_trait::async_trait;
 
+use crate::NodeIndex;
+
 /// The source of data items that consensus should order.
 ///
 /// AlephBFT internally calls [`DataProvider::get_data`] whenever a new unit is created and data needs to be placed inside.
@@ -19,5 +21,5 @@ pub trait DataProvider<Data>: Sync + Send + 'static {
 pub trait FinalizationHandler<Data>: Sync + Send + 'static {
     /// Data, provided by [DataProvider::get_data], has been finalized.
     /// The calls to this function follow the order of finalization.
-    fn data_finalized(&mut self, data: Data);
+    fn data_finalized(&mut self, data: Data, creator: NodeIndex);
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,7 +40,7 @@ dependencies = [
 
 [[package]]
 name = "aleph-bft"
-version = "0.29.0"
+version = "0.30.0"
 dependencies = [
  "aleph-bft-mock",
  "aleph-bft-rmc",
@@ -134,7 +134,7 @@ dependencies = [
 
 [[package]]
 name = "aleph-bft-mock"
-version = "0.10.0"
+version = "0.11.0"
 dependencies = [
  "aleph-bft-types",
  "async-trait",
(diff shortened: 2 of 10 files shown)


types/src/dataio.rs
```diff
@@ -19,5 +21,5 @@ pub trait DataProvider<Data>: Sync + Send + 'static {
 pub trait FinalizationHandler<Data>: Sync + Send + 'static {
     /// Data, provided by [DataProvider::get_data], has been finalized.
     /// The calls to this function follow the order of finalization.
-    fn data_finalized(&mut self, data: Data);
+    fn data_finalized(&mut self, data: Data, creator: NodeIndex);
```
timorleph, line 24: Since you are changing the API you got to bump the (minor because 0.x) version here, and in `consensus` at least.

timorleph (approved):
I think this might encourage people to be using this library wrong and I would prefer to do the things you are describing in another layer (so likely with the double-signatures). However, I don't really see a strong reason not to accept this, so sure. Just fix the versions.

timorleph:
There are some lints left, see unsuccessful checks.

PR #331, opened 2023-08-25
async send method for network trait to handle backpressure

joschisan:


fixxxedpoint (commented):
It looks ok in general. I don't know what exactly is your goal with this change, but please notice that it doesn't actually create much backpressure on other components, since the main component that uses this trait uses `unbounded_channels` anyway. So the only backpressure I see here is delays between `unit` messages and `alerts` being send. This can also create another type of problem, i.e. a single laggy node can slow down significantly communication with other nodes.

joschisan:
The goal with this change is to only drop messages if necessary in one place - the network layer if a peer is offline or laggy. However, I also do not want to drop messages when calling a sync send method on the network layer and dropping it just because the queues just happen to be full then. Furthermore, it could allow the network layer to run async code in its send method, which may be useful depending on the implementation. Its a small change, but allows for a cleaner system imo.

joschisan:
Please let me know how to bump the version numbers for this pr.

PR #335, opened 2023-08-28
A0-544: Alerter backup saving

woocash2:
Backup saving is added into `alerter::Service` as an intermediary step between handling the data by `Handler`, and sending the data out to the network so that the cycle of data of type `Alert` or `Multisigned` which passes through alerter is like this:

* `data` gets handled by the `Handler` which returns some `response` which is stored in the `Service`
* `data` is passed to `backup` via an appropriate channel
* after some time `backup` responds with `data` which triggers a branch in the main `select!` loop
* `response` corresponding to `data` is retrieved from the `Service`'s storage
* parts of `response` is sent through the network (or we do some `rmc` stuff with it) 

Heavy parts:

* refactor `Handler` so that there is no `on_message(message)` but it's split into a number of functions dependent on which variant is the `message`
* refactor `Service` so that `match message` happens inside `handle_network_alert` and not in `on_message` in `Handler` (which no longer exists).

This is done so that backup save can be triggered inside `Service` and not handler (it needs to happen only in a specific variant of message).


consensus/src/alerts/service.rs
```diff
@@ -157,26 +197,70 @@ impl<H: Hasher, D: Data, MK: MultiKeychain> Service<H, D, MK> {
...
+    }
+
     pub async fn run(&mut self, mut handler: Handler<H, D, MK>, mut terminator: Terminator) {
         loop {
             futures::select! {
                 message = self.messages_from_network.next() => match message {
                     Some(message) => self.handle_message_from_network(&mut handler, message),
                     None => {
-                        error!(target: LOG_TARGET, "{:?} Message stream closed.", self.node_index);
+                        error!(target: LOG_TARGET, "{:?} message stream closed.", self.node_index);
```
timorl, line 256: Hm, this reminds me, the fact that we log our own node index here is _extremely_ silly, since you are touching this maybe remove it?

timorl: Ah, but now the logs should start with a capital letter (and generally be structured like sentences.

consensus/src/alerts/service.rs
```diff
@@ -20,6 +20,14 @@ pub struct Service<H: Hasher, D: Data, MK: MultiKeychain> {
     rmc: ReliableMulticast<H::Hash, MK>,
     messages_for_rmc: Sender<RmcMessage<H::Hash, MK::Signature, MK::PartialMultisignature>>,
     messages_from_rmc: Receiver<RmcMessage<H::Hash, MK::Signature, MK::PartialMultisignature>>,
+
+    data_for_backup: Sender<AlertData<H, D, MK>>,
+    responses_from_backup: Receiver<AlertData<H, D, MK>>,
+
+    own_alert_responses: HashMap<H::Hash, OnOwnAlertResponse<H, D, MK>>,
+    network_alert_responses: HashMap<H::Hash, OnNetworkAlertResponse<H, D, MK>>,
+    multisigned_notifications: HashMap<H::Hash, ForkingNotification<H, D, MK::Signature>>,
```
kostekIV, line 29: I would sleep better if we would change hashmaps to some bounded-size version of it.

kostekIV (approved):
One comment, possibly ignorable

consensus/src/alerts/service.rs
timorl: Hm, good point, some `LruCache` or something probably...

consensus/src/alerts/service.rs
woocash2: aight

consensus/src/alerts/service.rs
woocash2: done

kostekIV: Cool, I have no idea about limits here. @timorl pls resolve this conversation if you find 1024 reasonable limit :)

timorl: Whoops, turns out this was a bad idea after all, sorry. ._.''

PR #339, opened 2023-08-30
A0-2961: Improve logs about DAG 

ggawryal:
This PR
* adds short runway DAG status report
* shortens missing coords log by compressing consecutive numbers into a single interval

For the latter, we could alternatively leave missing coords logging as it is (there were cases when we were missing all but just a few coords and the message was 30k chars long) or instead simply truncate it, loosing some information.

Changes log structure.


h4nsu (approved):
On the behalf of all future instances of me-looking-into-logs: THANK YOU <3

consensus/src/runway/mod.rs
```diff
@@ -171,18 +172,65 @@ impl<'a, H: Hasher> RunwayStatus<'a, H> {
...
+fn format_missing_coords(c: &[(usize, Round)]) -> String {
+    c.iter()
+        .group_by(|(creator, _)| *creator)
+        .into_iter()
+        .map(|(creator, rounds)| {
+            // compress consecutive rounds into one interval to shorten logs
+            let mut intervals: Vec<(Round, Round)> = Vec::new();
+            for (_, round) in rounds {
+                if !intervals.is_empty() && intervals.last().unwrap().1 == round - 1 {
+                    intervals.last_mut().unwrap().1 = *round;
```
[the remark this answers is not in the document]
ggawryal: Agreed, we were sorting just before the call of that function, so I'm moving sort to that function instead.

consensus/src/runway/mod.rs
```diff
@@ -1099,3 +1147,26 @@ pub(crate) async fn run<H, D, US, UL, MK, DP, FH, SH>(
...
+        assert_eq!(
+            format_missing_coords(&[(0, 1), (0, 2)]),
+            "{Creator 0: [1-2]}"
+        );
+        assert_eq!(
+            format_missing_coords(&[(0, 1), (0, 3), (0, 4), (0, 5), (0, 6), (0, 9)]),
+            "{Creator 0: 1,[3-6],9}"
+        );
+        assert_eq!(
+            format_missing_coords(&[(1, 3), (0, 1), (1, 1), (3, 0)]),
```
deuszx, line 1168: That's a bit _surprising_ , isn't it? Why is `(1, 3)` before `(0, 1)` ? This broke just now? `(1,3)` should follow `(1,1)` 🤔

consensus/src/runway/mod.rs
```diff
@@ -171,18 +172,65 @@ impl<'a, H: Hasher> RunwayStatus<'a, H> {
...
+    } else if rounds_behind == 0 {
+        format!("syncing - missing {missing_coords} unit(s)")
+    } else {
+        format!("syncing - missing {missing_coords} unit(s) and behind by {rounds_behind} rounds")
+    }
+}
+
+fn format_missing_coords(c: &[(usize, Round)]) -> String {
+    c.iter()
+        .sorted()
```
deuszx, line 189: Maybe use [`sort_by_key`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.sort_by_key) and use just the first element of the tuple?

consensus/src/runway/mod.rs
```diff
@@ -171,18 +172,65 @@ impl<'a, H: Hasher> RunwayStatus<'a, H> {
...
+
+fn format_missing_coords(c: &[(usize, Round)]) -> String {
+    c.iter()
+        .group_by(|(creator, _)| *creator)
+        .into_iter()
+        .map(|(creator, rounds)| {
+            // compress consecutive rounds into one interval to shorten logs
+            let mut intervals: Vec<(Round, Round)> = Vec::new();
+            for (_, round) in rounds {
+                if !intervals.is_empty() && intervals.last().unwrap().1 == round - 1 {
```
kostekIV, line 196:
```suggestion
                if matches!(intervals.last(), Some(interval) if interval.1 == round - 1) {
```

consensus/src/runway/mod.rs
ggawryal: It isn't - I just permuted input coords, cause now they may be not sorted, and the expected output string hasn't changed :stuck_out_tongue:

consensus/src/runway/mod.rs
```diff
@@ -171,18 +172,65 @@ impl<'a, H: Hasher> RunwayStatus<'a, H> {
...
+                    intervals.push((*round, *round));
+                }
+            }
+
+            let intervals_str = intervals
+                .into_iter()
+                .map(|(begin, end)| {
+                    if begin == end {
+                        format!("{}", begin)
+                    } else {
```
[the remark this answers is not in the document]
ggawryal: Done

consensus/src/units/store.rs
```diff
@@ -78,7 +83,7 @@ impl<H: Hasher, D: Data, K: Keychain> UnitStore<H, D, K> {
         }
     }
 
-    pub fn get_status(&self) -> UnitStoreStatus {
+    pub fn get_status(&self, wrt: NodeIndex) -> UnitStoreStatus {
```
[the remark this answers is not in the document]
ggawryal: Meant with respect to, but if this is not clear enough, I prefer `get_status_of(node)` to longer form

consensus/src/runway/mod.rs
```diff
@@ -171,18 +172,65 @@ impl<'a, H: Hasher> RunwayStatus<'a, H> {
...
 impl<'a, H: Hasher> fmt::Display for RunwayStatus<'a, H> {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "Runway status report: ")?;
-        write!(f, "{}", self.status)?;
+        write!(
+            f,
+            "Runway status report: {}",
+            short_report(self.status.rounds_behind(), self.missing_coords.len())
+        )?;
+        write!(f, "; {}", self.status)?;
```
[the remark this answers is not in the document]
ggawryal: `self.status`uses semicolons as separators, but since the first word in `self.status` is `DAG`, we can use period as well :)

consensus/src/runway/mod.rs
```diff
@@ -171,18 +172,65 @@ impl<'a, H: Hasher> RunwayStatus<'a, H> {
...
+            let intervals_str = intervals
+                .into_iter()
+                .map(|(begin, end)| {
+                    if begin == end {
+                        format!("{}", begin)
+                    } else {
+                        format!("[{}-{}]", begin, end)
+                    }
+                })
+                .format(",");
```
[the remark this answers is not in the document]
ggawryal: I don't like spaces there, but from grammar's pov you're correct

consensus/src/runway/mod.rs
```diff
@@ -171,18 +172,67 @@ impl<'a, H: Hasher> RunwayStatus<'a, H> {
     }
 }
 
+fn short_report(rounds_behind: Round, missing_coords: usize) -> String {
+    if rounds_behind <= 2 && missing_coords == 0 {
```
kostekIV, line 176: nit: I think match here would be more readable, feel free to ignore

consensus/src/runway/mod.rs
```diff
@@ -171,18 +172,67 @@ impl<'a, H: Hasher> RunwayStatus<'a, H> {
     }
 }
 
+fn short_report(rounds_behind: Round, missing_coords: usize) -> String {
```
kostekIV, line 175: nit: I somehow dont like that this is not a priv funcion of TunwayStatus struct

consensus/src/runway/mod.rs
```diff
@@ -171,18 +172,67 @@ impl<'a, H: Hasher> RunwayStatus<'a, H> {
...
+    } else if missing_coords == 0 {
+        format!("behind by {rounds_behind} rounds")
+    } else if rounds_behind == 0 {
+        format!("syncing - missing {missing_coords} unit(s)")
+    } else {
+        format!("syncing - missing {missing_coords} unit(s) and behind by {rounds_behind} rounds")
+    }
+}
+
+fn format_missing_coords(c: &[(usize, Round)]) -> String {
```
kostekIV, line 187: Same nit here

kostekIV (approved):
some nits left

consensus/src/runway/mod.rs
ggawryal: Found a tiny bug in 3rd if condition, so maybe you're right

consensus/src/runway/mod.rs
ggawryal: done &

consensus/src/runway/mod.rs
ggawryal: done

PR #340, opened 2023-08-31
A0-3138: Saver unit test

woocash2:
unit test for backup saver which asserts that we get responses from backup in the same order they arrived in backup.

### DIFF
--- a/consensus/src/backup/saver.rs
+++ b/consensus/src/backup/saver.rs
@@ -1,14 +1,13 @@
-use crate::{
-    units::UncheckedSignedUnit, Data, Hasher, MultiKeychain, Receiver, Sender, Terminator,
-};
+use std::io::Write;
 
-use crate::alerts::AlertData;
 use codec::Encode;
 use futures::{FutureExt, StreamExt};
-
-use crate::backup::BackupItem;
 use log::{debug, error};
-use std::io::Write;
+
+use crate::{
+    alerts::AlertData, backup::BackupItem, units::UncheckedSignedUnit, Data, Hasher, MultiKeychain,
+    Receiver, Sender, Terminator,
+};
 
 const LOG_TARGET: &str = "AlephBFT-backup-saver";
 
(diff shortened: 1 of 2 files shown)


consensus/src/backup/saver.rs
```diff
@@ -100,3 +99,133 @@ impl<H: Hasher, D: Data, MK: MultiKeychain, W: Write> BackupSaver<H, D, MK, W> {
...
+            .map(|k| {
+                TestAlertData::OwnAlert(Alert::new(
+                    NodeIndex(0),
+                    (units[k].clone(), units[k].clone()),
+                    vec![],
+                ))
+            })
+            .collect();
+
+        let backup_save_ordering = vec![0, 1, 1, 1, 0, 0, 1, 0, 1, 0];
```
kostekIV, line 201: nit: instead of 0,1 use enum { Alert(alert), Unit(unit) } .
the nyou can iterate this vector and get rid of i, UI and ai.

kostekIV (approved):
lgtm

consensus/src/backup/saver.rs
kostekIV: I was rather thinking of smth like this:
```
let backup_save_ordering = vec![Alert(alerts[0].clone), Unit(units[0.clone()]), ...];
```
but feel free to resolve this

consensus/src/backup/saver.rs
```diff
@@ -100,3 +99,147 @@ impl<H: Hasher, D: Data, MK: MultiKeychain, W: Write> BackupSaver<H, D, MK, W> {
...
+            .map(|k| {
+                TestAlertData::OwnAlert(Alert::new(
+                    NodeIndex(0),
+                    (units[k].clone(), units[k].clone()),
+                    vec![],
+                ))
+            })
+            .collect();
+
+        let backup_save_ordering = vec![
```
ggawryal, line 206: Why are we sending units and alerts in a mixed order, but are only checking their relative orders separately? Is this tests checking anything more than would splitting this into 2 independent unit tests with a common setup do?

woocash2: This is how saver would be used in practice. At any time it can be triggered to save either an `UncheckedSignedUnit` or `AlertData` through separate channels, but we only expect from it that the relative order of unit responses and alert data responses are preserved.

ggawryal: I see, how about renaming `test_proper_response_ordering` to `keeps_relative_order_of_units_and_alerts_separately` or such for clarity?

ggawryal (approved):

consensus/src/backup/saver.rs
woocash2: then we need to index them all manually which I don't like tbh :P

consensus/src/backup/saver.rs
woocash2: I'd still keep `response` in this name as this is what the ordering of we care about. How about `test_proper_relative_response_ordering`?

ggawryal: Sound good

woocash2: done

PR #347, opened 2023-09-14
A0-3185: split rmc

woocash2:
Splits the `rmc` logic into `Service` of 1 async method and synchronous `Handler`.
The split assures that `rmc` can be safely initialized from backup state and provides consistency with the design of our other components.
* Rewritten unit tests for `rmc::Service`
* Ran `blockchain` and `ordering` examples to verify the behavior after the changes

Not tested yet:
* Usage with aleph-node
* Scenarios with forks which should trigger alerts and thus rmcs


timorleph (approved):
Nice.

rmc/src/service.rs
```diff
@@ -0,0 +1,425 @@
...
+    network_tx: UnboundedSender<RmcOutgoingMessage<H, MK>>,
+    scheduler: Box<dyn TaskScheduler<Task<H, MK>>>,
+}
+
+impl<H: Signable + Hash + Eq + Clone + Debug, MK: MultiKeychain> Service<H, MK> {
+    pub fn new(
+        network_rx: UnboundedReceiver<
+            RmcIncomingMessage<H, MK::Signature, MK::PartialMultisignature>,
+        >,
+        network_tx: UnboundedSender<RmcOutgoingMessage<H, MK>>,
```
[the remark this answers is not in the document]
woocash2: hmm, `Network` from `types` won't fit here as it assumes the sent data type is the same as received data type.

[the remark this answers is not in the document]
timorleph: Yeah, I'm suggesting it should return `RMCInterface` (with better name plx) rather than use any traits – we don't really use it as a network anyway, especially with the `start_rmc` thing.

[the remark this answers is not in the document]
woocash2: I'd need to create a separate interface for that. Also `IO` in `alerter` will still be needed.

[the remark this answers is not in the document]
timorleph: Yeah, something like that. The `IO` thing would've been a nice bonus, but it's not crucial.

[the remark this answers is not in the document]
woocash2: Tho, using a trait here seems too generic to me. I think imma just wrap these channels with a `struct RMCIO` or sth like that. This would be implemented with channels but it will obscure them from the user.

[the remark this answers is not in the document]
timorleph: Yeah, that's exactly what I had in mind, sorry for not being clear. ^^"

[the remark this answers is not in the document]
woocash2: I encountered some problems with this approach (dm-ed the details). Considering leaving the channels as they are...

rmc/src/service.rs
```diff
@@ -0,0 +1,420 @@
...
+/// When sufficintly many nodes initiate rmc with the same message `msg` and a node collects enough
+/// signatures to form a complete multisignature under the message, the multisigned message
+/// is sent back through the network by the instance of [`Service<H, MK>`].
+///
+/// We refer to the documentation https://cardinal-cryptography.github.io/AlephBFT/reliable_broadcast.html
+/// for a high-level description of this protocol and how it is used for fork alerts.
+pub struct Service<H: Signable + Hash, MK: MultiKeychain> {
+    network_rx: UnboundedReceiver<RmcIncomingMessage<H, MK::Signature, MK::PartialMultisignature>>,
+    network_tx: UnboundedSender<RmcOutgoingMessage<H, MK>>,
+    scheduler: Box<dyn TaskScheduler<RmcHash<H, MK::Signature, MK::PartialMultisignature>>>,
```
fixxxedpoint, line 36: No need for `dyn` here. You can use a generic parameter (some more juggling with `TestEnvironment`) or just `Double...` since its the only implementation anyway.

rmc/src/service.rs
```diff
@@ -0,0 +1,420 @@
...
+    }
+
+    fn send_hash(&self, hash: RmcHash<H, MK::Signature, MK::PartialMultisignature>) {
+        self.network_tx
+            .unbounded_send(RmcOutgoingMessage::RmcHash(hash))
+            .expect("Sending message should succeed");
+    }
+
+    /// Run the rmc service.
+    pub async fn run(mut self, mut handler: Handler<H, MK>, mut terminator: Terminator) {
```
fixxxedpoint, line 113: Handler can't be just part of the `Service` type?

rmc/src/service.rs
```diff
@@ -0,0 +1,420 @@
...
+                    Ok(Some(multisigned)) => {
+                        self.scheduler.add_task(RmcHash::MultisignedHash(
+                            multisigned.clone().into_unchecked(),
+                        ));
+                        if self
+                            .network_tx
+                            .unbounded_send(RmcOutgoingMessage::NewMultisigned(multisigned))
+                            .is_err()
+                        {
+                            error!(target: "AlephBFT-rmc", "network closed early.");
```
fixxxedpoint, line 94: This just logs, below in `send_hash` it panics. Do you know why?

fixxxedpoint (changes requested):
lgtm

rmc/src/service.rs
woocash2: yeah, I think it can. the same should apply for `alerter` so imma solve it too

rmc/src/handler.rs
```diff
@@ -0,0 +1,118 @@
...
+        let _ = self.on_signed_hash(&unchecked);
+        unchecked
+    }
+
+    /// Update the internal state with the signed hash. If the hash is incorrectly signed then
+    /// [`Error::BadSignature`] is returned. If Adding this signature completes a multisignature
+    /// then `Ok(multisigned)` is returned. Otherwise `Ok(None)` is returned.
+    pub fn on_signed_hash(
+        &mut self,
+        unchecked: &UncheckedSigned<Indexed<H>, MK::Signature>,
```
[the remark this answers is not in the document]
woocash2: `let hash = ...` makes a borrow, so cloning below is still needed. When keeping the reference in the argument, at least we save one clone, cuz we can call `on_signed_hash(&x)` instead `on_signed_hash(x.clone())`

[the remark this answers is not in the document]
woocash2: Also it is consistent with other methods & outer scope functions which take `unchecked` by reference

rmc/src/service.rs
woocash2: yeah, that's much cooler

rmc/src/service.rs
woocash2: yeah... quite inconsistent. I shall replace this code with `send_hash` and keep the panic as it was there previously

rmc/src/service.rs
```diff
@@ -0,0 +1,420 @@
...
+            }
+            RmcIncomingMessage::RmcHash(RmcHash::MultisignedHash(unchecked)) => {
+                match self.handler.on_multisigned_hash(&unchecked) {
+                    Ok(Some(multisigned)) => {
+                        self.scheduler.add_task(RmcHash::MultisignedHash(unchecked));
+                        self.send_message(RmcOutgoingMessage::NewMultisigned(multisigned));
+                    }
+                    Ok(None) => {}
+                    Err(error) => {
+                        warn!(target: LOG_TARGET, "{}", error);
```
fixxxedpoint, line 88: Put some short context in the log-string so it will be easier to find source using logs.

rmc/src/service.rs
```diff
@@ -0,0 +1,420 @@
...
+                match self.handler.on_signed_hash(&unchecked) {
+                    Ok(Some(multisigned)) => {
+                        self.scheduler.add_task(RmcHash::MultisignedHash(
+                            multisigned.clone().into_unchecked(),
+                        ));
+                        self.send_message(RmcOutgoingMessage::NewMultisigned(multisigned));
+                    }
+                    Ok(None) => {}
+                    Err(error) => {
+                        warn!(target: LOG_TARGET, "{}", error);
```
fixxxedpoint, line 102: Put some short context in the log-string so it will be easier to find source using logs.

fixxxedpoint (changes requested):

rmc/src/handler.rs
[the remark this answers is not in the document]
woocash2: Hmm, yeah, I kinda like the idea of separating `handle_signed_hash` out so that we don't check the signature produced by us in `on_start_rmc`. However your implementation doesn't seem to work for me (even after some adjustments). 

`handle_signed_hash` takes `Signed` instead of `UncheckedSigned` which is reasonable as we don't want to check the signature, but `UncheckedSigned` is **needed** here as extracting the `hash` is only possible through `UncheckedSigned` (by calling `as_signable_strip_index`). `Signed` can only be converted to `UncheckedSigned` by consuming the original `Signed` (`into_unchecked`). Thus some cloning is inevitable here (if we want to pass `signed` to `add_signature` later) :smile:. 

Although, I think that the implementation of `Signed` and `Indexed` is unnecessarily restricted. Why do we have to convert to `UncheckedSigned` each time we want to access the `hash` directly? I think that hash residing in `Indexed` should be accessible by reference (e.g. by introducing `as_signable(&self) -> &T` in `Indexed`) and it would allow accessing the `hash` without the need of cloning `Signed` to convert to `UncheckedSigned` and then to access the hash.

Here is how I see `handle_signed_hash` now:
```rust
fn handle_signed_hash(&mut self, signed: Signed<Indexed<H>, MK>) -> Option<Multisigned<H, MK>> {
        let unchecked = signed.clone().into_unchecked(); // needa clone signed to get the hash
        let hash = unchecked.as_signable_strip_index();
        if self.already_completed(hash) {
            return None;
        }
        let new_state = match self.hash_states.remove(hash) {
            None => signed.into_partially_multisigned(&self.keychain),
            Some(partial) => partial.add_signature(signed, &self.keychain),
        };
        match new_state {
            PartiallyMultisigned::Complete { multisigned } => {
                self.hash_states.insert(
                    hash.clone(),
                    PartiallyMultisigned::Complete {
                        multisigned: multisigned.clone(),
                    },
                );
                Some(multisigned)
            }
            incomplete => {
                self.hash_states.insert(hash.clone(), incomplete);
                None
            }
        }
    }
```
and here is how it could look like if we could access &hash from `Indexed`:
```rust
fn handle_signed_hash(&mut self, signed: Signed<Indexed<H>, MK>) -> Option<Multisigned<H, MK>> {
        let hash = signed.as_signable().as_signable().clone(); // access hash without cloning the signed
        if self.already_completed(&hash) {
            return None;
        }
        let new_state = match self.hash_states.remove(&hash) {
            None => signed.into_partially_multisigned(&self.keychain),
            Some(partial) => partial.add_signature(signed, &self.keychain),
        };
        match new_state {
            PartiallyMultisigned::Complete { multisigned } => {
                self.hash_states.insert(
                    hash.clone(),
                    PartiallyMultisigned::Complete {
                        multisigned: multisigned.clone(),
                    },
                );
                Some(multisigned)
            }
            incomplete => {
                self.hash_states.insert(hash.clone(), incomplete);
                None
            }
        }
    }
```

fixxxedpoint (approved):

consensus/src/alerts/mod.rs
```diff
@@ -106,7 +106,7 @@ pub enum AlertMessage<H: Hasher, D: Data, S: Signature, MS: PartialMultisignatur
     /// Alert regarding forks, signed by the person claiming misconduct.
     ForkAlert(UncheckedSigned<Alert<H, D, S>, S>),
     /// An internal RMC message, together with the id of the sender.
-    RmcMessage(NodeIndex, RmcMessage<H::Hash, S, MS>),
+    RmcHash(NodeIndex, RmcMessage<H::Hash, S, MS>),
```
timorleph, line 109: Probably un-change this name, now it's just confusing. :P

consensus/src/alerts/service.rs
```diff
@@ -5,69 +5,73 @@ use crate::{
...
+            responses_from_backup,
+        } = io;
 
-        let rmc = ReliableMulticast::new(
-            messages_from_us,
-            messages_for_us,
-            keychain.clone(),
-            DoublingDelayScheduler::new(time::Duration::from_millis(500)),
+        let node_index = keychain.index();
+        let rmc_handler = aleph_bft_rmc::Handler::new(keychain);
```
timorleph, line 55: Considering what we talked about earlier it's unlikely you will end up creating this handler here, but I guess you can extract it into its proper place in the next PR. Feel free to resolve if this is your plan.

rmc/src/handler.rs
```diff
@@ -0,0 +1,122 @@
...
+    /// Signs hash and updates the internal state with it. Returns the signed
+    /// version of the hash for broadcast. Should be called at most once for a particular hash.
+    pub fn on_start_rmc(&mut self, hash: H) -> OnStartRmcResponse<H, MK> {
+        let signed_hash = Signed::sign_with_index(hash, &self.keychain);
+        let mut maybe_multisigned = None;
+        if !self.already_completed(signed_hash.as_signable().as_signable()) {
+            maybe_multisigned = self.handle_signed_hash(signed_hash.clone());
+        }
+        (signed_hash, maybe_multisigned)
+    }
```
timorleph, line 51: Nit: empty line after this one plx.

rmc/src/lib.rs
```diff
@@ -1,619 +1,34 @@
...
+
+pub use handler::Handler;
+pub use scheduler::DoublingDelayScheduler;
+pub use service::Service;
 
-/// An RMC message consisting of either a signed (indexed) hash, or a multisigned hash.
+/// An RMC hash consisting of either a signed (indexed) hash, or a multisigned hash.
 #[derive(Debug, Encode, Decode, Clone, PartialEq, Eq, Hash)]
-pub enum Message<H: Signable, S: Signature, M: PartialMultisignature> {
+pub enum RmcMessage<H: Signable, S: Signature, M: PartialMultisignature> {
```
timorleph, line 19: Well, this rename looks kinda silly...

rmc/src/handler.rs
```diff
@@ -0,0 +1,122 @@
...
+        Ok(Some(multisigned))
+    }
+
+    fn already_completed(&self, hash: &H) -> bool {
+        matches!(
+            self.hash_states.get(hash),
+            Some(PartiallyMultisigned::Complete { .. })
+        )
+    }
+}
```
timorleph, line 122: It would be nice to have tests here – in particular just deasyncing the scenarios from `service.rs` tests should be fine and not too much effort. Although it would also be great if you checked for the specific situation in which the test were failing, so that `start_rmc` returns a multisig.

timorleph (approved):
Couple minor comments. Looks nice, good interface change after all.

consensus/src/alerts/mod.rs
woocash2: good catch

consensus/src/alerts/service.rs
woocash2: yep, constructing handlers in proper places remains for the next pr

consensus/src/alerts/service.rs
```diff
@@ -289,7 +274,6 @@ impl<H: Hasher, D: Data, MK: MultiKeychain> Service<H, D, MK> {
                 },
             }
             if self.exiting {
-                debug!(target: LOG_TARGET, "Alerter decided to exit.");
```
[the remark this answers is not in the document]
woocash2: good point, removed cuz (wrongly) thought it was redundant and forgot to bring back

rmc/src/handler.rs
```diff
@@ -0,0 +1,122 @@
...
+impl<H: Signable + Hash + Eq + Clone + Debug, MK: MultiKeychain> Handler<H, MK> {
+    pub fn new(keychain: MK) -> Self {
+        Handler {
+            hash_states: HashMap::new(),
+            keychain,
+        }
+    }
+
+    /// Signs hash and updates the internal state with it. Returns the signed
+    /// version of the hash for broadcast. Should be called at most once for a particular hash.
```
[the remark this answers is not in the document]
woocash2: Ok, I'll try returning either one and see if the tests will still pass

rmc/src/lib.rs
woocash2: actually it was `Message --> RmcHash --> RmcMessage` :stuck_out_tongue:
idk, I like `RmcMessage` more because we don't have to make aliases for it in `alerter` or use the full path `aleph_bft_rmc::Message`. Just import `RmcMessage` and use as it is.

rmc/src/handler.rs
woocash2: ok, I'll try to come up with some tests

rmc/src/service.rs
```diff
@@ -0,0 +1,351 @@
...
+            }
+            OnStartRmcResponse::MultisignedHash(multisigned) => {
+                self.scheduler.add_task(RmcMessage::MultisignedHash(
+                    multisigned.clone().into_unchecked(),
+                ));
+                return Some(multisigned);
+            }
+            OnStartRmcResponse::Noop => {}
+        }
+        None
```
timorleph, line 72: Hm, I think this might be marginally more readable if you have two of these at the end of the select branches, but don't have the `return` there. Then it's explicit what every branch returns, rather than having to parse the entire thing to understand that.

rmc/src/lib.rs
```diff
@@ -1,40 +1,20 @@
...
+mod handler;
+mod scheduler;
+mod service;
+
+pub use handler::Handler;
+pub use scheduler::DoublingDelayScheduler;
+pub use service::Service;
 
-/// An RMC message consisting of either a signed (indexed) hash, or a multisigned hash.
+/// An RMC hash consisting of either a signed (indexed) hash, or a multisigned hash.
```
timorleph, line 17: s/hash/message

rmc/src/service.rs
woocash2: uhhh, lemme change the other one as well

rmc/src/handler.rs
```diff
@@ -127,3 +129,223 @@ impl<H: Signable + Hash + Eq + Clone + Debug, MK: MultiKeychain> Handler<H, MK>
...
+        let mut handler = Handler::new(keychain);
+        let expected = Signed::sign_with_index(hash.clone(), &keychain);
+        assert_eq!(
+            handler.on_start_rmc(hash),
+            OnStartRmcResponse::SignedHash(expected)
+        );
+    }
+
+    #[test]
+    fn on_start_rmc_exceeding_quorum_returns_multisigned() {
```
timorleph, line 191: s/exceeding/reaching, same in the comment below and the other test also about that, should be clearer

timorleph (approved):
One naming nit, other than that great! <3

rmc/src/service.rs
```diff
@@ -53,310 +43,250 @@ impl<H, MK, SCH> Service<H, MK, SCH>
...
+                    ));
+                    return Some(multisigned);
+                }
+                Ok(None) => {}
+                Err(error) => {
+                    warn!(target: LOG_TARGET, "failed handling multisigned hash: {}", error);
+                }
+            },
+            Message::MultisignedHash(unchecked) => {
                 match self.handler.on_multisigned_hash(unchecked.clone()) {
```
fixxxedpoint, line 97: You don't need to clone here? Later you can call `multisigned.clone().into_unchecked()` or provide `.as_unchecked` for `Multisigned`...

fixxxedpoint (approved):

PR #362, opened 2023-10-23
A0-1505: Remove alert backup

woocash2:
Remove alert backup since the task got abandoned for now.


timorleph (approved):
I'm approving this as I don't see a problem with the PR itself, but I also raised some concerns as to whether it should happen at all on other channels.

PR #399, opened 2024-01-19
Allow async implementation for BackupReader and BackupWriter

maan2003:
using `io::Write` or `io::Read` forces a sync implementation. 

we were using `block_on` to use async inside the implementation, but this blocking the executor and sometimes leads to deadlocks. 

this is backwards compatible by implementing new traits for any `io::Read` and `io::Write` impls.

see https://github.com/fedimint/fedimint/issues/4054 for more context.

### DIFF
--- a/consensus/src/backup/loader.rs
+++ b/consensus/src/backup/loader.rs
@@ -1,7 +1,12 @@
-use std::{collections::HashSet, fmt, fmt::Debug, io::Read, marker::PhantomData};
+use std::{
+    collections::HashSet,
+    fmt::{self, Debug},
+    marker::PhantomData,
+    pin::Pin,
+};
 
 use codec::{Decode, Error as CodecError};
-use futures::channel::oneshot;
+use futures::{channel::oneshot, AsyncRead, AsyncReadExt};
 use log::{error, info, warn};
 
 use crate::{
@@ -63,26 +68,26 @@ impl From<CodecError> for LoaderError {
     }
 }
 
-pub struct BackupLoader<H: Hasher, D: Data, S: Signature, R: Read> {
-    backup: R,
+pub struct BackupLoader<H: Hasher, D: Data, S: Signature, R: AsyncRead> {
+    backup: Pin<Box<R>>,
     index: NodeIndex,
     session_id: SessionId,
     _phantom: PhantomData<(H, D, S)>,
 }
 
-impl<H: Hasher, D: Data, S: Signature, R: Read> BackupLoader<H, D, S, R> {
+impl<H: Hasher, D: Data, S: Signature, R: AsyncRead> BackupLoader<H, D, S, R> {
     pub fn new(backup: R, index: NodeIndex, session_id: SessionId) -> BackupLoader<H, D, S, R> {
         BackupLoader {
-            backup,
+            backup: Box::pin(backup),
             index,
             session_id,
             _phantom: PhantomData,
         }
     }
 
-    fn load(&mut self) -> Result<Vec<UncheckedSignedUnit<H, D, S>>, LoaderError> {
+    async fn load(&mut self) -> Result<Vec<UncheckedSignedUnit<H, D, S>>, LoaderError> {
         let mut buf = Vec::new();
-        self.backup.read_to_end(&mut buf)?;
+        self.backup.read_to_end(&mut buf).await?;
         let input = &mut &buf[..];
         let mut result = Vec::new();
         while !input.is_empty() {
@@ -163,7 +168,7 @@ impl<H: Hasher, D: Data, S: Signature, R: Read> BackupLoader<H, D, S, R> {
         starting_round: oneshot::Sender<Option<Round>>,
         next_round_collection: oneshot::Receiver<Round>,
     ) {
-        let units = match self.load() {
+        let units = match self.load().await {
             Ok(items) => items,
             Err(e) => {
                 error!(target: LOG_TARGET, "unable to load backup data: {}", e);
--- a/consensus/src/backup/saver.rs
+++ b/consensus/src/backup/saver.rs
@@ -1,8 +1,8 @@
-use std::io::Write;
+use std::pin::Pin;
 
 use crate::{units::UncheckedSignedUnit, Data, Hasher, Receiver, Sender, Signature, Terminator};
 use codec::Encode;
-use futures::{FutureExt, StreamExt};
+use futures::{AsyncWrite, AsyncWriteExt, FutureExt, StreamExt};
 use log::{debug, error};
 
 const LOG_TARGET: &str = "AlephBFT-backup-saver";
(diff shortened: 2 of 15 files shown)


dpc:
> this is backwards compatible by implementing new traits for any `io::Read` and `io::Write` impls.

Isn't the previous way stricly incorrect? The code running it is `async`, so it can't be calling blocking IO methods without some form of offloading from the current executor (like `tokio::task::block_in_place`). The current backward-compat impls preserve the existing incorrect behavior.

Maan2003:
should we just remove `Read` and `Write` impls. we can't use offloading methods because we don't depend on tokio.

dpc:
> should we just remove `Read` and `Write` impls. we can't use offloading methods because we don't depend on tokio.

IMO there's no reason to try to preserve the old (incorrect) behavior.

Maan2003:
we can't remove `Read` and `Write` because `mock` crate impls `Read` for mock reader. 

mock can't impl `BackupReader` because `mock` can't depend main crate (circular deps).
Solution to this would be moving mock to same crate, but that will be big change.

dpc:
> mock can't impl `BackupReader` because `mock` can't depend main crate (circular deps).

So a mock can't depend on the thing it's mocking? Weird, but I don't have time to dig into it.

>  Solution to this would be moving mock to same crate, but that will be big change.

Or move the trait to a separate crate, so both can depend on it. That's how `Read` was working - it was in a 3rd crate (stdlib).

joschisan:
> we can't remove `Read` and `Write` because `mock` crate impls `Read` for mock reader.
> 
> mock can't impl `BackupReader` because `mock` can't depend main crate (circular deps). Solution to this would be moving mock to same crate, but that will be big change.

I think we can move the new traits into the aleph-bft-types crate, that's were the other traits used in the api like Network are too.

consensus/src/backup/loader.rs
```diff
@@ -63,14 +64,29 @@ impl From<CodecError> for LoaderError {
     }
 }
 
-pub struct BackupLoader<H: Hasher, D: Data, S: Signature, R: Read> {
+#[async_trait]
+pub trait BackupReader {
```
timorleph, line 68: If possible it would be nice to not create additional traits here at all. Could you just use `futures::io::AsyncRead`?

consensus/src/backup/loader.rs
```diff
@@ -63,14 +64,29 @@ impl From<CodecError> for LoaderError {
...
+    /// Read the entire backup.
+    async fn read(&mut self) -> std::io::Result<Vec<u8>>;
+}
+
+#[async_trait]
+impl<R: Read + Send> BackupReader for R {
```
timorleph, line 74: Don't worry about backwards compatibility here, it's better to use actually async implementations if we are changing this interface. If you still need an additional trait (see the comment above) then perhaps replace this implementation with one for `AsyncRead`, otherwise just remove it.

Speaking of compatibility – please bump the version in `Cargo.toml` (to `34.0`?) since this changes the interface.

consensus/src/backup/saver.rs
```diff
@@ -1,22 +1,38 @@
...
+use async_trait::async_trait;
 use codec::Encode;
 use futures::{FutureExt, StreamExt};
 use log::{debug, error};
 
 const LOG_TARGET: &str = "AlephBFT-backup-saver";
 
+#[async_trait]
+/// Write backups to peristent storage.
+pub trait BackupWriter {
```
timorleph, line 13: Same comments as in `Reader`, both here and below.

timorleph (changes requested):

consensus/src/backup/loader.rs
timorleph: Ah, I forgot to re-read the discussion before commenting, and it turns out to be very relevant. :P Anyway, if you can just use `AsyncRead` as suggested above then it solves the dependency problem, otherwise add it to `types`, as @joschisan suggested.

woocash2:
Hi, we have looked into your PR and into the issues in your repository, apologies for letting it hang here for a long time. We have introduced traits `Read` and `Write` here mostly with an intention of using `io::File` or a similar kind of a lightweight object with read/write semantics as the implementation. Although we consider it as the preferred way to implement the backup, we understand your need for an asynchronous interface and we're eager to merge the PR in some form. I had my colleague leave some comments so please act upon them, also we'll have to run some checks. Thank you for your contribution.

dpc:
> We have introduced traits `Read` and `Write` here mostly with an intention of using `io::File`

If it was to be done this way then the core alephbft async code calling this interface should be calling it wrapped in some https://docs.rs/tokio/latest/tokio/task/fn.block_in_place.html , otherwise async executor gets blocked on blocking IO operations, potentially stalling async executor.

consensus/src/backup/loader.rs
dpc: https://docs.rs/futures/latest/futures/io/trait.AsyncRead.html is a very inconvenient interface to work with on both sides, because it is meant to be used as general purpose IO trait, not just a single use case like here. It will probably get changed in the future as well when (if?) Rust developers figure out how to square the circle of all the requirements put on it: https://www.ncameron.org/blog/async-read-and-write-traits/

I don't see any benefit of using it. In this case core alephbft logic wants the implementation to handle a single simple operation so expressing it as such makes things easier.

timorleph: This is kinda a good argument for leaving it as the sync version of `Read`, implementing it for a buffer and reading the data into a buffer separately, in whatever way you want. In fact this is what we do in `aleph-node`: https://github.com/Cardinal-Cryptography/aleph-node/blob/9ecbc0d50b52acc90e9c33acc0fed3d64436fcc7/finality-aleph/src/party/backup.rs#L64.

That being said, this only works for `Read`, do you have a proposal for the writing part?

maan2003: using converting from sync `Read` to async Read is trivial using `AllowStdIo` in futures-io.

consensus/src/backup/saver.rs
```diff
@@ -25,14 +25,15 @@ impl<H: Hasher, D: Data, S: Signature, W: Write> BackupSaver<H, D, S, W> {
...
 
-    pub fn save_item(&mut self, item: &UncheckedSignedUnit<H, D, S>) -> Result<(), std::io::Error> {
-        self.backup.write_all(&item.encode())?;
-        self.backup.flush()?;
-        Ok(())
+    pub async fn save_item(
+        &mut self,
+        item: &UncheckedSignedUnit<H, D, S>,
+    ) -> Result<(), std::io::Error> {
+        self.backup.write_all(&item.encode()).await
```
timorleph, line 36: You should `flush` here as well.

timorleph (approved):
Nice, thanks! One little comment left.

timorleph:
Oh, and some failing checks, mostly related to version bumps, one of which should happen in the readme.

consensus/src/backup/saver.rs
maan2003: fixed

timorleph:
`mock` and `rmc` still need version bumps, minor and patch respectively. Although tbh I'm not quite sure why you touched `rmc` at all?

maan2003:
maybe version checks don't understand that removing imports doesn't need bump?

maan2003:
oh those were pub use but not exported at top level

PR #411, opened 2024-02-09
Replace outdated runner

timorleph:
https://github.com/Cardinal-Cryptography/AlephBFT/actions/runs/7815688558

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   deploy:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/gh-pages.yml
```diff
@@ -10,7 +10,7 @@ on:
 
 jobs:
   deploy:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
```
Kosiare, line 13: Please do not use `ubuntu-latest`, instead specify it to `ubuntu-20.04`.

timorleph: Changed as suggested. Do you think it will work now?

Kosiare: According to annotation - yes, it should :)

Kosiare (approved):

PR #423, opened 2024-03-11
A0-4152: Global consensus threshold and other `crypto` improvements

timorleph:
Small, seemingly disjointed changes to the `crypto` API and some resulting changes in the `consensus` unit tests. They were easy to split off of the bigger refactor so I did that to make that one smaller and make it clearer what the version changes entail.


lesniak43 (approved):

consensus/src/creation/creator.rs
```diff
@@ -169,7 +167,7 @@ mod tests {
     fn create_unit_with_minimal_parents(n_members: NodeCount) {
         let n_parents = (n_members.0 * 2) / 3 + 1;
```
kostekIV, line 168: nit: use consensus_threshold

kostekIV (commented):
What about:
* function quorum in `mock/src/crypto/keychain.rs`
* function quorum in `crypto/src/signature.rs`
* in crashed_nodes_recover fn on line 168 in fule `consensus/src/testing/crash_recovery.rs` you can also use consensus_threshold function

kostekIV (approved):

PR #425, opened 2024-03-13
A0-4160: Make packing part of unit creation

timorleph:
Another part of the larger refactor. Currently looks a bit weird, because `consensus` becomes more dependent on non-bare-unit stuff technically, but this divide won't make much sense in the future anyway – specific components will be better separated between themselves, rather than having to separate out whole subthreads.


consensus/src/creation/mod.rs
```diff
@@ -133,68 +108,77 @@ async fn keep_processing_units_until<H: Hasher>(
...
 ) -> anyhow::Result<(), CreatorError> {
-    let Config {
-        node_id,
-        n_members,
-        create_lag,
-        max_round,
-    } = conf;
+    let node_id = conf.node_ix();
+    let n_members = conf.n_members();
+    let create_lag = conf.delay_config().unit_creation_delay.clone();
```
woocash2, line 172:
```suggestion
    let create_delay = conf.delay_config().unit_creation_delay.clone();
```

consensus/src/runway/mod.rs
```diff
@@ -751,10 +738,10 @@ where
                     }
                 },
 
-                signed_unit = self.signed_units_from_packer.next() => match signed_unit {
-                    Some(signed_unit) => self.on_packed(signed_unit),
+                signed_unit = self.signed_units_from_creation.next() => match signed_unit {
+                    Some((signed_unit, _)) => self.on_created(signed_unit),
```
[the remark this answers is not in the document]
timorleph: Not currently, but:
1. It will almost surely be in one of the future parts of the refactor.
1. This was already the case before the refactor. :P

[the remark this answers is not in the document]
timorleph: Welp, it seems you were right after all, the next part of this refactor removes this... I would prefer to keep the removal there though. <_<"

consensus/src/creation/mod.rs
```diff
@@ -133,86 +108,97 @@ async fn keep_processing_units_until<H: Hasher>(
...
-            keep_processing_units_until(&mut creator, incoming_parents, lag).await?;
+            keep_processing_units_until(&mut creator, incoming_parents, delay).await?;
         }
 
-        let (unit, parent_hashes) = create_unit(round, &mut creator, incoming_parents).await?;
-
-        trace!(target: "AlephBFT-creator", "Created a new unit {:?} at round {:?}.", unit, round);
+        let (preunit, parent_hashes) = create_unit(round, &mut creator, incoming_parents).await?;
+        trace!(target: LOG_TARGET, "Created a new preunit {:?} at round {:?}.", preunit, round);
+        let data = data_provider.get_data().await;
```
fixxxedpoint, line 195: Concerns:
1) this could be called parallel to pre-unit creation - these are independent tasks + calls to .get_data slows down units processing throughput of the `creator`
2) would be a good idea to have some common and simple pattern for calling `external` api implementations, which can potentially block indefinitely (this one even uses a Mutex...in a safe way :) )

Both could be solved by wrapping a provided impl of `DataProvider` by something that for example spawns_blocking and buffers calls to `.get_data`.

consensus/src/creation/mod.rs
```diff
@@ -133,86 +108,97 @@ async fn keep_processing_units_until<H: Hasher>(
...
+        trace!(target: LOG_TARGET, "Received data.");
```
fixxxedpoint, line 196: Since this is `trace`, maybe add some details about `received data`, like `was it None or Some`.

consensus/src/runway/mod.rs
```diff
@@ -914,23 +901,38 @@ pub(crate) async fn run<H, D, US, UL, MK, DP, FH, SH>(
     MK: MultiKeychain,
     SH: SpawnHandle,
 {
+    let RunwayIO {
+        data_provider,
+        finalization_handler,
+        backup_write,
+        backup_read,
+        _phantom: _,
```
fixxxedpoint, line 909: You can't use something like `..,` here?

fixxxedpoint (approved):
lgtm

consensus/src/creation/mod.rs
timorleph: Both of these are actually on purpose:
1. We want to have the freshest possible data, while `create_unit` is expected to take quite a while in a normal situation, it has an explicit delay. This had an explicit comment beforehand, should I re-add it?
2. We want this to be able to block, because then you can have a "lazy" consensus mechanism that only does any work when data is available (or not yet ordered, the implementation of such a thing is not completely trivial).

consensus/src/runway/mod.rs
timorleph: I like getting the additional check from the compiler that everything has been unpacked, since here actually everything should be. If there were more such fields I would use `..`, but in this case I would prefer not to.

consensus/src/creation/mod.rs
timorleph: Good point, it turns out `Data` is apparently `Debug`, so I just print the whole thing out.

consensus/src/runway/mod.rs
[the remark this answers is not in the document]
woocash2: ok

PR #429, opened 2024-03-27
A0-4153: Factor unit validation out of runway

timorleph:
This required quite a lot of other changes as well, so this is quite large. It also looks... not ideal and in places _more_ convoluted than beforehand, but completely untangling all this in a single commit would require it to be even larger.


woocash2:
This PR is very large. It would be very helpful if you included a thorough description of what happens and some guides for the reviewers in the PR description e.g.
* Explain how unit validation was done before
* Explain how unit validation is done now
* List all the new / deleted components
* List which responsibilities were moved between which components.
* Explain any changes in logic / simplifications
* Which changes were indented, which were just forced by the intended ones?
* ...anything else which may be helpful

Some of the above may be irrelevant, but just to give a picture of what I would be happy to see.
Since the PR is so big, the review time will also be long, but a thorough description would speed it up `3x` :stuck_out_tongue:

timorleph:
Ah, fair enough.

As the title implies the main goal of this refactor was to get unit validation logic out of `Runway`. The basic validation was already contained in `units::Validator`, but the rest, i.e. fork detection and handling, was now moved into `validation::Validator`, which should be the main focus of this review (plus the changes in `Runway`). This required several more changes to get to work, plus I made a couple possibly strictly speaking unnecessary changes of things that kept getting in my way and it was easier to fix them rather than keep tripping over them.

Perhaps the most important additional change is the complete rewrite of `units::Store`. In fact its purpose is now very different, so much so, that it's probably better to just read the file rather than trying to look at the diff. In particular it no longer keeps explicit track of forks, that responsibility has been moved to `validation::Validator`. Since this store is completely fresh, the whole file is also an important target of review.

To make the new abstraction for the unit store work units had to be abstracted away, rather than transformed. This is perhaps the biggest difference in what is actually happening during unit processing. Before, units were saved after validation in `Runway` and only stripped information about them was sent on to `reconstruction` and later `extension`. Now, they are sent into `reconstruction` directly, without saving (or rather only saved in `validation::Validator` for fork detection purposes), where they end up being encapsulated with explicit parent information. Such encapsulated units are then saved in a `Store` in `Runway`, and still only stripped information is eventually sent to `extender` (this will likely change in future parts of this refactor though). In particular `reconstruction` now depends on a `Unit` trait, which was a struct before. This trait, and the related `WrappedUnit`, should also be reviewed.

This change in abstraction necessitated some further changes again, but most of them pretty trivial – most changes related to this in `creation` and `reconstruction` should at most be skimmed through. One change that was not trivial here though was moving the passing of units to `extension` from `reconstruction` to `runway`. This is also why `extension` moved from `consensus` to `runway` and why one of the `Notification[In|Out]` got deleted – I removed the second one for symmetry, as it was mostly confusing at this point. This is also why there are a couple _more_ channels now, they pass through the, previously internal, variants from the `Notification*` types – this is what I referred to as "more convoluted" in the description above, and I'm already working on removing the complication in the next part of the refactor. In particular `consensus` will almost surely be merged with `runway` in the next PR (possibly under the former name?), since `creation` will also move.

A couple smaller changes, that perhaps could have been avoided:
1. `creation` no longer returns explicit parents, as it became _painfully_ obvious that these are not necessary.
2. parent responses are passed to `reconstruction` with a `HashMap` – I think this was actually needed in some way because of the abstraction changes, but I no longer remember why, so I'm not sure; this will definitely be better when implementing the "more parents" task though.
3. runway status reports had to be somewhat rewritten, but they kept most of their functionality
4. tests (in `testing`) got quite impacted, I even removed one of them (it was kinda duplicated by others fortunately) – it should be sufficient to just skim through them though

Phew, that's quite a lot, if you still have questions I'm available for a call, just send me a DM.

consensus/src/validation.rs
```diff
@@ -0,0 +1,483 @@
...
+    Data, Hasher, MultiKeychain, NodeIndex, NodeSubset, Round,
+};
+
+/// What can go wrong when validating a unit.
+#[derive(Eq, PartialEq)]
+pub enum Error<H: Hasher, D: Data, MK: MultiKeychain> {
+    Invalid(ValidationError<H, D, MK::Signature>),
+    Duplicate(SignedUnit<H, D, MK>),
+    Uncommitted(SignedUnit<H, D, MK>),
+    NewForker(Box<Alert<H, D, MK::Signature>>),
```
woocash2, line 18: Do we need a `Box` here?

kostekIV: Hmm I think its strange to return alert in the error variant. I think more natural way to do it would be including just necessary extra information like other forker unit for example.

consensus/src/validation.rs
```diff
@@ -0,0 +1,483 @@
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> FmtResult {
+        write!(
+            f,
+            "processing units: ({}), forkers: {}",
+            self.processing_units, self.known_forkers
+        )
+    }
+}
+
+type ValidatorResult<H, D, MK> = Result<SignedUnit<H, D, MK>, Error<H, D, MK>>;
```
kostekIV, line 64: wdyt about renaming it to `ValidationResult`?

kostekIV (approved):
Looks quite good :)

consensus/src/validation.rs
timorleph: In somewhat good news this becomes more natural in the next part, where handling of these errors gets factored out of runway. Although it remains an error variant internally, and returns the whole alert (it would be somewhat annoying to do this differently, since we only have all the necessary info inside of `validation`).

consensus/src/units/store.rs
```diff
@@ -1,291 +1,273 @@
...
-    by_hash: HashMap<H::Hash, SignedUnit<H, D, K>>,
-    parents: HashMap<H::Hash, Vec<H::Hash>>,
-    //the number of unique nodes that we hold units for a given round
-    is_forker: NodeSubset,
-    legit_buffer: Vec<SignedUnit<H, D, K>>,
-    max_round: Round,
+/// Stores units and remembers the first instance of any unit with a specific coordinate inserted.
+pub struct UnitStore<U: Unit> {
+    by_hash: HashMap<HashFor<U>, U>,
+    by_coord: NodeMap<HashMap<Round, HashFor<U>>>,
```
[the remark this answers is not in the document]
timorleph: This makes it much easier to get all units for a given creator (needed for alerts), but if it's really slower... hm. I really don't wanna write benches tho. ;_;

consensus/src/validation.rs
timorleph: It's not `pub`, so it seemed that it should refer to the `Validator` more than to the whole module, but sure if you think that's better I'll do that.

consensus/src/runway/mod.rs
```diff
@@ -743,7 +690,16 @@ where
 
                 notification = self.notifications_from_alerter.next() => match notification {
                     Some(notification) => {
-                        self.on_alert_notification(notification)
+                        trace!(target: "AlephBFT-runway", "Received alerter notification: {:?}.", notification);
```
woocash2, line 693: To keep it consistent with the other `select` branches, maybe let's include this logic in some `fn handle_alerter_notification`

woocash2: If you don't want to spam with too many functions, the logic of `handle_validation_error` can be fully included in `handle_alerter_notification`

timorleph: :D You are so very right that I just 20mins ago did that within the next PR. Can we leave it here as-is for now?

woocash2: sure

consensus/src/runway/mod.rs
```diff
@@ -725,16 +665,23 @@ where
...
-                notification = self.rx_consensus.next() => match notification {
-                    Some(notification) => self.on_consensus_notification(notification),
+                unit = self.units_from_reconstruction.next() => match unit {
+                    Some(unit) => self.on_unit_reconstructed(unit),
                     None => {
-                        error!(target: "AlephBFT-runway", "{:?} Consensus notification stream closed.", index);
+                        error!(target: "AlephBFT-runway", "{:?} Reconstructed unit stream closed.", index);
+                        break;
+                    }
+                },
```
woocash2, line 674:
```suggestion
                },
                
```

consensus/src/units/store.rs
woocash2, line 33: For clarity, I would name this field `canonical_units` or `canonical`. Maybe it's just me but two fields in a struct named `by_*` suggest that they store the same data  but with different keys to access them (even if the types dispel the doubt).

consensus/src/units/store.rs
```diff
@@ -1,291 +1,273 @@
...
-
-pub(crate) struct UnitStore<H: Hasher, D: Data, K: Keychain> {
-    by_coord: HashMap<UnitCoord, SignedUnit<H, D, K>>,
-    by_hash: HashMap<H::Hash, SignedUnit<H, D, K>>,
-    parents: HashMap<H::Hash, Vec<H::Hash>>,
-    //the number of unique nodes that we hold units for a given round
-    is_forker: NodeSubset,
-    legit_buffer: Vec<SignedUnit<H, D, K>>,
-    max_round: Round,
+/// Stores units and remembers the first instance of any unit with a specific coordinate inserted.
```
woocash2, line 30: I would also include in this comment info that first instance == canonical, and to the first part, that it stores units by hash. Then everything will be 100% clear

consensus/src/units/store.rs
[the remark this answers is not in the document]
woocash2: I agree that two-level access is convenient to get all canonical units for a given creator.

consensus/src/runway/mod.rs
```diff
@@ -551,52 +493,50 @@ where
         }
     }
 
-    fn on_created(&mut self, signed_unit: SignedUnit<H, D, MK>) {
+    fn on_created(&mut self, unit: SignedUnit<H, D, MK>) {
```
woocash2, line 496:
```suggestion
    fn on_unit_created(&mut self, unit: SignedUnit<H, D, MK>) {
```
for consistency

consensus/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aleph-bft"
-version = "0.36.1"
+version = "0.36.2"
```
woocash2, line 3: I suppose you made sure that no public interface has changed? :stuck_out_tongue:

consensus/src/units/mod.rs
```diff
@@ -200,34 +181,77 @@ pub(crate) type UncheckedSignedUnit<H, D, S> = UncheckedSigned<FullUnit<H, D>, S
 
 pub(crate) type SignedUnit<H, D, K> = Signed<FullUnit<H, D>, K>;
 
-#[derive(Clone, Eq, PartialEq, Debug, Decode, Encode)]
-pub struct Unit<H: Hasher> {
-    pre_unit: PreUnit<H>,
-    hash: H::Hash,
-}
+/// Abstract representation of a unit from the Dag point of view.
+pub trait Unit: 'static + Send + Clone {
```
woocash2, line 185: This and `WrappedUnit` seem cool, but what's the reason why you introduce them? Just asking, not questioning.

consensus/src/units/mod.rs
```diff
@@ -200,34 +181,77 @@ pub(crate) type UncheckedSignedUnit<H, D, S> = UncheckedSigned<FullUnit<H, D>, S
...
-        self.pre_unit.creator()
+
+    fn round(&self) -> Round {
+        self.coord().round()
     }
-    pub(crate) fn round(&self) -> Round {
-        self.pre_unit.round()
+}
+
+pub trait WrappedUnit<H: Hasher>: Unit<Hasher = H> {
```
woocash2, line 203: I think I understand the purpose of `Unit` - you want the store be abstract of the unit type. But I'm not sure I get the purpose of `WrappedUnit`.

consensus/src/units/store.rs
```diff
@@ -1,291 +1,273 @@
...
+        }
+        self.by_hash.insert(unit_hash, unit);
     }
 
-    // Outputs None if this is not a newly-discovered fork or Some(sv) where (su, sv) form a fork
-    pub(crate) fn is_new_fork(&self, fu: &FullUnit<H, D>) -> Option<SignedUnit<H, D, K>> {
-        if self.contains_hash(&fu.hash()) {
-            return None;
+    /// Remove a unit with a given hash. Notably if you remove a unit another might become canonical in its place in the future.
+    pub fn remove(&mut self, hash: &HashFor<U>) {
```
woocash2, line 78: I don't see `remove` being used anywhere in the main code. Will it be used in the future?

timorleph: It's used within `validation` actually – in `finished_processing` iirc.

consensus/Cargo.toml
timorleph: I'm pretty confident, this separation between API and internals is already done relatively well in ABFT. :)

consensus/src/units/mod.rs
timorleph: In `validation` we get a store with units of a different type (with parents in practice, but there is no reason for `validation` to know that) as an argument, but we have to be able to use these as the type we have there, i.e. `SignedUnit`. It would just as well be an `Into<SignedUnit<H, D, MK>> + Unit<H>` I guess?

woocash2 (approved):
Looks aight, a solid refactor. Didn't find anything sus.

consensus/src/runway/mod.rs
timorleph: Done.

consensus/src/units/store.rs
timorleph: Good idea, done.

consensus/src/units/store.rs
timorleph: Yup, added.

woocash2: I meant changing the doc-string so that it contains the definition of `canonical unit`, and that it contains info that all the units are stored by hash.
It's a very minor thing tho, feel free to ignore & resolve.

consensus/src/units/mod.rs
woocash2: Aight, we can keep this simplification in the back of our heads. I won't ask you to change it to `Into<....`, unless you think it's actually more convenient.

consensus/src/units/store.rs
woocash2: Right, I missed that `processing_units` are of type `UnitStore`.

consensus/src/units/store.rs
timorleph: Better now?

PR #432, opened 2024-04-09
A0-4155: Unit saving pipeline

timorleph:
Make the unit saving into a pipeline, to simplify the semantics of units being officially added to the Dag.

### DIFF
--- a/consensus/src/backup/saver.rs
+++ b/consensus/src/backup/saver.rs
@@ -1,6 +1,10 @@
 use std::pin::Pin;
 
-use crate::{units::UncheckedSignedUnit, Data, Hasher, Receiver, Sender, Signature, Terminator};
+use crate::{
+    dag::DagUnit,
+    units::{UncheckedSignedUnit, WrappedUnit},
+    Data, Hasher, MultiKeychain, Receiver, Sender, Terminator,
+};
 use codec::Encode;
 use futures::{AsyncWrite, AsyncWriteExt, FutureExt, StreamExt};
 use log::{debug, error};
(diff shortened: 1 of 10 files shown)


consensus/src/backup/saver.rs
```diff
@@ -10,30 +14,29 @@ const LOG_TARGET: &str = "AlephBFT-backup-saver";
...
 
-    pub async fn save_item(
-        &mut self,
-        item: &UncheckedSignedUnit<H, D, S>,
-    ) -> Result<(), std::io::Error> {
-        self.backup.write_all(&item.encode()).await?;
+    pub async fn save_unit(&mut self, unit: &DagUnit<H, D, MK>) -> Result<(), std::io::Error> {
+        let parents = unit.parents().clone();
+        let unit: UncheckedSignedUnit<_, _, _> = unit.clone().unpack().into();
+        self.backup.write_all(&(unit, parents).encode()).await?;
```
woocash2, line 39: I don't understand why do we save a unit along with its parents.
1. It can potentially incur some slowdown in the pipeline
2. Memory usage of backup increases dramatically
3. We ignore the parents anyway when loading

consensus/src/backup/saver.rs
```diff
@@ -10,30 +14,29 @@ const LOG_TARGET: &str = "AlephBFT-backup-saver";
...
     }
 
-    pub async fn save_item(
-        &mut self,
-        item: &UncheckedSignedUnit<H, D, S>,
-    ) -> Result<(), std::io::Error> {
-        self.backup.write_all(&item.encode()).await?;
+    pub async fn save_unit(&mut self, unit: &DagUnit<H, D, MK>) -> Result<(), std::io::Error> {
+        let parents = unit.parents().clone();
+        let unit: UncheckedSignedUnit<_, _, _> = unit.clone().unpack().into();
```
woocash2, line 38: I think there's too much hustle here to "deconstruct" the `DagUnit`. With all of this cloning, we have 3 copies of `parents` in this scope. Maybe `DagUnit` needs a `fn unit()` which only clones the inner unit?

woocash2 (changes requested):
Including backup save in the pipeline is overall a good idea, although I don't support the idea of saving `(unit, parents)` in the backup. If you want the `DagUnit` to go through the backup, no problem with that, just when doing backup save, only save the inner unit, and pass the whole `DagUnit` further in the pipeline.

If there's some hidden idea behind saving the whole `DagUnit`, please explain.

consensus/src/backup/saver.rs
timorleph: The original idea was that this would be then used for crash resilience in case of forks, where you would need explicit parents to be able to load the saved units. However, given the deprioritization of that work that happened in the meantime I think it's not worth it after all – I'll change this back to just saving parentless units, and revert the version bump (this was the only part of the API that changed, which was part of the motivation of changing it early).

woocash2: Ok, nice. As I thought, there would be some good justification for this. Nevertheless, even if we continued the work on fork-aware crash resilience, in my opinion it would be better to make this change when it would have become relevant (i.e. not in this pr).

consensus/src/backup/saver.rs
timorleph: Ah, I didn't notice this comment still makes sense after the above change (even though now it's only one unnecessary `clone` of parents). Hm, I would prefer to keep the interface simpler, but I guess I can add a `unit` method to `WrappedUnit`, although I'm not convinced the performance worries are that important here – you still need a `clone` of a similarly structured object, no?

timorleph: Ah, I didn't notice this comment still makes sense after the above change (even though now it's only one unnecessary `clone` of parents). Hm, I would prefer to keep the interface simpler, but I guess I can add a `unit` method to `WrappedUnit`, although I'm not convinced the performance worries are that important here – you still need a `clone` of a similarly structured object, no?

woocash2: how about adding it directly to `ReconstructedUnit` without changing `WrappedUnit`?

woocash2: I don't like the unnecessary clone, but if it's more convenient for your abstractions, then it's fine. I'll let you decide and resolve this comment.

woocash2 (approved):

kostekIV (approved):
lgtm, nice change in the add_unit function but a little risky :P

alfellati:
> lgtm, nice change in the add_unit function but a little risky :P

What makes it a little risky?

timorleph:
> What makes it a little risky?

I think @kostekIV means that it will no longer remove the units from the cache immediately after they are being processed by the `dag` component, which was a nice property to have (no need to remember to call `finished_processing` outside of `dag`). It's not really that risky, since the cache cannot create different outputs if it's not cleaned, worst case scenario it would use a tad more memory.

PR #436, opened 2024-04-12
A0-4215: Make extender synchronous

timorleph:
One fewer spawn.

### DIFF
--- a/consensus/src/extension/mod.rs
+++ b/consensus/src/extension/mod.rs
@@ -11,7 +11,7 @@ mod units;
 use extender::Extender;
 
 /// A struct responsible for executing the Consensus protocol on a local copy of the Dag.
-/// It receives units which are guaranteed to be eventually in the Dags
+/// It receives units which are guaranteed to eventually appear in the Dags
 /// of all honest nodes. The static Aleph Consensus algorithm is then run on this Dag in order
 /// to finalize subsequent rounds of the Dag. More specifically whenever a new unit is received
 /// this process checks whether a new round can be finalized and if so, it computes the batch of
--- a/consensus/src/extension/mod.rs
+++ b/consensus/src/extension/mod.rs
@@ -1,8 +1,8 @@
-use futures::{FutureExt, StreamExt};
-
-use log::{debug, warn};
-
-use crate::{Hasher, NodeIndex, NodeMap, Receiver, Round, Sender, Terminator};
+use crate::{
+    dag::DagUnit,
+    units::{Unit, WrappedUnit},
+    Data, FinalizationHandler, Hasher, MultiKeychain,
+};
 
 mod election;
 mod extender;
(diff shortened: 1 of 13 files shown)


lesniak43 (approved):

consensus/src/extension/mod.rs
```diff
@@ -1,189 +1,52 @@
...
-            hash,
-            parents,
-        }
-    }
-}
-
-/// A process responsible for executing the Consensus protocol on a local copy of the Dag.
-/// It receives units via a channel `electors` which are guaranteed to be eventually in the Dags
+/// A struct responsible for executing the Consensus protocol on a local copy of the Dag.
+/// It receives units which are guaranteed to be eventually in the Dags
```
woocash2, line 14:
```suggestion
/// It receives units which are guaranteed to eventually appear in the Dag.
```
(resolvable)

woocash2:
I'm worried that this refactor may have negative impact on the performance. `Extender` contains relatively heavy logic and now sending out a unit through the network is delayed by it.

Have you considered changing order of operations in `on_unit_backup_saved`? I.e. is there a chance that the overall performance will be better if `self.ordering.add_unit(unit.clone())` was executed at the end of the function?

consensus/src/extension/mod.rs
```diff
@@ -1,189 +1,52 @@
...
-                debug!(target: LOG_TARGET, "{:?} Extender decided to exit.", self.node_id);
-                terminator.terminate_sync().await;
-                break;
-            }
+            finalization_handler,
         }
     }
-}
-
-#[cfg(test)]
```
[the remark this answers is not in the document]
timorleph: The only thing this test was related to threading, which is no longer an issue here. The only thing we could additionally be testing here is whether the elected units get passed on to the handler (since all other aspects of the election are tested in submodules), which is kinda already covered by `testing::dag`. I could add a test here as well, but it would be quite annoying for little gain imo.

consensus/src/extension/mod.rs
timorleph: Fixed, although `Dags`, because it refers to all the nodes.

consensus/src/runway/mod.rs
```diff
@@ -471,13 +470,16 @@ where
             warn!(target: "AlephBFT-runway", "Creator channel should be open.");
             self.exiting = true;
         }
-        let unit = unit.unpack();
-        self.send_message_for_network(RunwayNotificationOut::NewAnyUnit(unit.clone().into()));
-
-        if unit.as_signable().creator() == self.index() {
-            trace!(target: "AlephBFT-runway", "{:?} Sending a unit {:?}.", self.index(), unit.as_signable().hash());
-            self.send_message_for_network(RunwayNotificationOut::NewSelfUnit(unit.into()));
+        let unpacked_unit = unit.clone().unpack();
```
woocash2, line 473: :raised_eyebrow:

woocash2 (approved):

PR #439, opened 2024-04-19
A0-1672: units returned by run_session

fixxxedpoint:
A separate method in member `run_session_for_units` that allows one to track finalization of batches of units.

### DIFF
--- a/consensus/src/dag/reconstruction/mod.rs
+++ b/consensus/src/dag/reconstruction/mod.rs
@@ -1,13 +1,14 @@
 use std::collections::HashMap;
 
 use crate::{
-    units::{ControlHash, HashFor, Unit, UnitCoord, UnitWithParents, WrappedUnit},
+    units::{ControlHash, FullUnit, HashFor, Unit, UnitCoord, UnitWithParents, WrappedUnit},
     Hasher, NodeMap, SessionId,
 };
 
 mod dag;
 mod parents;
 
+use aleph_bft_types::{Data, MultiKeychain, OrderedUnit, Signed};
 use dag::Dag;
 use parents::Reconstruction as ParentReconstruction;
 
@@ -76,11 +77,20 @@ impl<U: Unit> UnitWithParents for ReconstructedUnit<U> {
     }
 }
 
-pub type Parents<U> = NodeMap<HashFor<U>>;
-
-impl<U: Unit> From<ReconstructedUnit<U>> for (U, Parents<U>) {
-    fn from(value: ReconstructedUnit<U>) -> Self {
-        (value.unit, value.parents)
+impl<D: Data, H: Hasher, K: MultiKeychain> From<ReconstructedUnit<Signed<FullUnit<H, D>, K>>>
+    for OrderedUnit<D, H>
+{
+    fn from(unit: ReconstructedUnit<Signed<FullUnit<H, D>, K>>) -> Self {
+        let parents = unit.parents;
+        let hash = unit.unit.hash();
+        let (unit, data) = unit.unit.into_signable().into();
+        OrderedUnit {
+            data,
+            parents,
+            hash,
+            creator: unit.creator(),
+            round: unit.round(),
+        }
     }
 }
 
--- a/consensus/src/dag/reconstruction/mod.rs
+++ b/consensus/src/dag/reconstruction/mod.rs
@@ -81,18 +81,22 @@ impl<D: Data, H: Hasher, K: MultiKeychain> From<ReconstructedUnit<Signed<FullUni
     for OrderedUnit<D, H>
 {
     fn from(unit: ReconstructedUnit<Signed<FullUnit<H, D>, K>>) -> Self {
-        let parents = unit.parents().into_iter().map(|(ix, hash)| (ix, *hash)).collect();
+        let parents = unit
+            .parents()
+            .into_iter()
+            .map(|(ix, hash)| (ix, *hash))
+            .collect();
         let unit = unit.unpack();
         let creator = unit.creator();
         let round = unit.round();
         let hash = unit.hash();
-        let (_, data) = unit.into_signable().into();
+        let data = unit.into_signable().data().clone();
         OrderedUnit {
-            data,
             parents,
-            hash,
             creator,
             round,
+            hash,
+            data,
         }
     }
 }
--- a/consensus/src/dag/reconstruction/mod.rs
+++ b/consensus/src/dag/reconstruction/mod.rs
@@ -76,6 +76,12 @@ impl<U: Unit> UnitWithParents for ReconstructedUnit<U> {
     }
 }
 
+impl<U: Unit> From<ReconstructedUnit<U>> for (U, NodeMap<HashFor<U>>) {
+    fn from(value: ReconstructedUnit<U>) -> Self {
+        (value.unit, value.parents)
+    }
+}
+
 /// What we need to request to reconstruct units.
 #[derive(Debug, PartialEq, Eq)]
 pub enum Request<H: Hasher> {
--- a/consensus/src/dag/reconstruction/mod.rs
+++ b/consensus/src/dag/reconstruction/mod.rs
@@ -76,7 +76,9 @@ impl<U: Unit> UnitWithParents for ReconstructedUnit<U> {
     }
 }
 
-impl<U: Unit> From<ReconstructedUnit<U>> for (U, NodeMap<HashFor<U>>) {
+pub type Parents<U> = NodeMap<HashFor<U>>;
+
+impl<U: Unit> From<ReconstructedUnit<U>> for (U, Parents<U>) {
     fn from(value: ReconstructedUnit<U>) -> Self {
         (value.unit, value.parents)
     }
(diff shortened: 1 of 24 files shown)


types/src/dataio.rs
```diff
@@ -1,36 +1,56 @@
...
+
+/// Represents state of the main internal data structure of AlephBFT (i.e. direct acyclic graph) used for
+/// achieving consensus.
+///
+/// Instances of this type are returned indirectly by [`member::run_session_for_units`] method using the
+/// [`UnitFinalizationHandler`]. This way it allows to reconstruct the DAG's structure used by AlephBFT,
+/// which can be then used for example for the purpose of node's performance evaluation.
+pub struct OrderedUnit<D: Data, H: Hasher> {
+    pub data: Option<D>,
+    pub parents: NodeMap<H::Hash>,
```
timorleph, line 40: Probably better to return a `Vec` of hashes here instead – this will always represent parents and the API will be more stable.

consensus/src/extension/mod.rs
```diff
@@ -34,14 +38,8 @@ impl<H: Hasher, D: Data, MK: MultiKeychain, FH: FinalizationHandler<D>> Ordering
...
     fn handle_batch(&mut self, batch: Vec<DagUnit<H, D, MK>>) {
-        for unit in batch {
-            let unit = unit.unpack();
-            self.finalization_handler.unit_finalized(
-                unit.creator(),
-                unit.round(),
-                unit.as_signable().data().clone(),
-            )
-        }
+        let batch = batch.into_iter().map(|unit| unit.into()).collect();
```
timorleph, line 41: An additional (minor) advantage of having a separate trait is that you could take a `Vec` of something that's `Into` instead of doing this here.

consensus/src/dag/reconstruction/mod.rs
```diff
@@ -76,6 +77,23 @@ impl<U: Unit> UnitWithParents for ReconstructedUnit<U> {
     }
 }
 
+impl<D: Data, H: Hasher, K: MultiKeychain> From<ReconstructedUnit<Signed<FullUnit<H, D>, K>>>
+    for OrderedUnit<D, H>
+{
+    fn from(unit: ReconstructedUnit<Signed<FullUnit<H, D>, K>>) -> Self {
+        let parents = unit.parents;
+        let hash = unit.unit.hash();
```
timorleph, line 85: This and above could be accessed through methods (of `UnitWithParents` iirc) instead of fields.

timorleph: You can also get the creator and round this way btw.

consensus/src/dag/reconstruction/mod.rs
```diff
@@ -76,6 +77,23 @@ impl<U: Unit> UnitWithParents for ReconstructedUnit<U> {
...
+        let (unit, data) = unit.unit.into_signable().into();
```
timorleph, line 86: With the changes above this should probably be changed into `unit.unpack().as_signable().data().clone()` and the `From` impl can be removed.

timorleph (approved):
Cool changes, although I have some comments.

Also cc @joschisan – as you can see we are actually implementing the more advanced interface after all, so feel free to comment here.

lesniak43 (approved):

mock/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "aleph-bft-mock"
-version = "0.14.0"
+version = "0.14.1"
```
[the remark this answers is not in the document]
fixxxedpoint: changed

consensus/src/dag/reconstruction/mod.rs
fixxxedpoint: Actually I would prefer to do that using fields directly instead. I would like to treat it as a low-level constructor that knows its own internals instead of using generic traits. This way it fits better to implement `Into` directly (I know internals of `ReconstructedUnit<..>` and want to create `OrderedUnit`...but clippy complains about it :| ). Using generic traits here suggests that I should rather have something like `From<U: Unit>` instead, but this way I can't access `Data`, etc. That `.clone()` also smells a bit. Anyway, I've changed it.

consensus/src/dag/reconstruction/mod.rs
fixxxedpoint: changed

types/src/dataio.rs
fixxxedpoint: changed

consensus/src/dag/reconstruction/mod.rs
fixxxedpoint: changed

consensus/src/member.rs
```diff
@@ -107,35 +106,44 @@ enum TaskDetails<H: Hasher, D: Data, S: Signature> {
...
+
+    pub fn new_using_unit_finalization_handler<
+        H: Hasher,
+        UFH: FinalizationHandler<BatchOfUnits<DP::Output, H>>,
+    >(
+        data_provider: DP,
+        finalization_handler: UFH,
+        unit_saver: US,
+        unit_loader: UL,
+    ) -> LocalIO<DP, UFH, US, UL> {
```
timorleph, line 141: Returning something other than `Self` here is quite unusual. I think it might be better to have this whole `impl` block be for `LocalIO<DP, UFH, US, UL>`, both the constructors return `Self` and the plain `new` is parametrized by `FH: FinalizationHandler<DP::Output>` (but still returns `Self` – that should work because of the autoimpl, right?).

Or is there some problem with `H` being in the `impl`, but not a parameter of `LocalIO`?

consensus/src/dag/reconstruction/mod.rs
```diff
@@ -76,6 +77,26 @@ impl<U: Unit> UnitWithParents for ReconstructedUnit<U> {
     }
 }
 
+impl<D: Data, H: Hasher, K: MultiKeychain> From<ReconstructedUnit<Signed<FullUnit<H, D>, K>>>
+    for OrderedUnit<D, H>
+{
+    fn from(unit: ReconstructedUnit<Signed<FullUnit<H, D>, K>>) -> Self {
+        let parents = unit.parents().into_iter().map(|(_, hash)| *hash).collect();
```
timorleph, line 84: Use `into_values`.

joschisan:
@timorleph the advanced interface works for us just as well.

consensus/src/extension/mod.rs
fixxxedpoint: changed

consensus/src/member.rs
```diff
@@ -575,17 +569,58 @@ where
...
+}
+
+pub type BatchOfUnits<D, H> = Vec<OrderedUnit<D, H>>;
+
+/// Starts the consensus algorithm as an async task. It stops establishing consensus for new data items after
+/// reaching the threshold specified in [`Config::max_round`] or upon receiving a stop signal from `exit`.
+/// Please note that this interface is less stable than [`run_session`] as it exposes intrinsics (i.e. units)
+/// which migh be subject to change.
+#[doc(hidden)]
+pub async fn run_session_for_units<
```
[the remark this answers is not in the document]
fixxxedpoint: changed

types/src/dataio.rs
```diff
@@ -1,36 +1,56 @@
...
+    pub hash: H::Hash,
+    pub creator: NodeIndex,
+    pub round: Round,
+}
+
+impl<D: Data, H: Hasher, FH: FinalizationHandler<D>> FinalizationHandler<Vec<OrderedUnit<D, H>>>
```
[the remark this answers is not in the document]
fixxxedpoint: I've added a new trait - `UnitFinalizationHandler`. Current version doesn't provide an auto-impl for `FInalizationHandler`. Instead there is a wrapper-type - `FInalizationHandlerAdapter`.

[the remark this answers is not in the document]
fixxxedpoint: If we want to `hide` that adapter type, then I need to return something different than `Self` from the `new` constructor of `LocalIO` - `FH` parameter needs to become `impl UnitFInalizationHandler`.

consensus/src/member.rs
```diff
@@ -573,19 +621,27 @@ where
...
-    D: Data,
-    DP: DataProvider<D>,
-    FH: FinalizationHandler<D>,
+    DP: DataProvider,
+    UFH: UnitFinalizationHandler<Data = DP::Output, Hasher = H>,
     US: AsyncWrite + Send + Sync + 'static,
     UL: AsyncRead + Send + Sync + 'static,
-    N: Network<NetworkData<H, D, MK::Signature, MK::PartialMultisignature>> + 'static,
+    N: Network<NetworkData<H, DP::Output, MK::Signature, MK::PartialMultisignature>>
+        + 'static,
```
timorleph, line 639: Since you are already changing so many things anyway, perhaps also add the `'static` requirement straight into `Network`?

timorleph (approved):
A couple small complaints more, just to make the interfaces perfect. ;>

consensus/src/runway/mod.rs
```diff
@@ -100,19 +99,19 @@ type CollectionResponse<H, D, MK> = UncheckedSigned<
     <MK as Keychain>::Signature,
 >;
 
-struct Runway<H, D, FH, MK>
+struct Runway<H, D, UFH, MK>
 where
     H: Hasher,
     D: Data,
-    FH: FinalizationHandler<D>,
+    UFH: UnitFinalizationHandler<Data = D, Hasher = H>,
```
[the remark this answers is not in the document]
timorleph: I don't think there is, but it also doesn't look _so_ verbose to me.

PR #490, opened 2024-11-06
A0-4382: Tiny preparations for dismemberment

timorleph:
This is only moving some code around, as a preparation for the network component refactor, in particular one that will end up splitting `member.rs` into two components with more clearly defined responsibilities.


consensus/src/network/hub.rs
```diff
@@ -0,0 +1,105 @@
...
+    alerts::AlertMessage,
+    member::UnitMessage,
+    network::{NetworkData, NetworkDataInner},
+    Data, Hasher, Network, PartialMultisignature, Receiver, Recipient, Sender, Signature,
+    Terminator,
+};
+use futures::{FutureExt, StreamExt};
+use log::{debug, error, warn};
+
+pub struct Hub<
```
Marcin-Radecki, line 11: Overall, the change looks good, but I just have one question: why rename `NetworkHub` -> `Hub`? I liked the former better; it was the more specific name (and `Hub` still has `network` as a field).

Marcin-Radecki (approved):

consensus/src/network/hub.rs
timorleph: Oh, it was because it's `network::Hub` anyway, I prefer renaming this on import to having `network::NetworkHub`, but this is a small preference, I can rename it back.

Marcin-Radecki: You're right that module gives proper naming anyway, so it's fine to leave it as it is.

PR #495, opened 2024-11-22
A0-4564: Modified example/ordering so that it sends unique elements from data provider

Marcin-Radecki:
* `example/ordering` binary now produces unique `DataProvider` input, so that the successful finish condition is that all nodes need to see at least one integer for union of all node's `DataProvider` items,
* reworked `run.sh`:
  * support for passing unit creation delay (was the `defult_unit_creation_delay()`, which is 5000 ms, that is far greater than production values)
  * added better crash simulation - now `run.sh` `kill`s node process, so it cannot anticipate crash

### DIFF
--- a/examples/ordering/run.sh
+++ b/examples/ordering/run.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -eoxu pipefail
+set -eou pipefail
 
 function usage() {
   cat << EOF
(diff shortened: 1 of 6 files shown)


examples/ordering/run.sh
```diff
@@ -1,97 +1,187 @@
...
+   $0
+     [-n|--nodes NODES]
+       number of all non-crashing nodes; some of them can have stalled data provider
+     [-c|--crashing-nodes CRASHING_NODES]
+       number of nodes that crash while providing data
+     [-s|--stalling-data-providers STALLING_DATA_PROVIDERS]
+       number of nodes that eventually stall while providing data; must be less than --nodes
+     [--crashes-count CRASHES_COUNT]
+       how many times a crashing node should crash
+     [--data-items DATA_ITEMS]
```
timorleph, line 29: Make this consistent with `DATA_SIZE_BATCH` above (no opinion as to which name is better).

Marcin-Radecki: Done

PR #496, opened 2024-11-27
A0-4568: Move response logic out of runway

timorleph:
This almost exclusively moves code from `runway` to the (newly created) `dissemination` module. It also adds tests for this logic, which required some general improvements in the testing setup, thus the changes in other files.


Marcin-Radecki (approved):
Value added!

PR #503, opened 2024-12-05
A0-4563: Added unit test for censorship resistance in Extender

Marcin-Radecki:
Introduced 2 UTs, similar to existing ones, but with prepared DAG that contains non-direct parents. It turns out that current testing code almost support non-direct parents, as `ReconstructedUnit` control hash is created based on the parents injected by the test, except one method `random_unit_with_parents` that assumed each parent round is the same. 

Currently, those tests are disabled since they (expectedly) panic at 
```
units are added in order
thread 'extension::extender::test::given_minimal_dag_with_orphaned_node_when_producing_batches_have_correct_length' panicked at consensus/src/extension/election.rs:56:42:
```

This PR should not contain any changes to the production code.

### DIFF
--- a/consensus/src/extension/election.rs
+++ b/consensus/src/extension/election.rs
@@ -217,9 +217,8 @@ mod test {
             units::Units,
         },
         units::{
-            random_full_parent_reconstrusted_units_up_to, random_reconstructed_unit_with_parents,
-            minimal_reconstructed_dag_units_up_to,
-            TestingDagUnit, Unit,
+            minimal_reconstructed_dag_units_up_to, random_full_parent_reconstrusted_units_up_to,
+            random_reconstructed_unit_with_parents, TestingDagUnit, Unit,
         },
         NodeCount,
     };
@@ -345,7 +344,7 @@ mod test {
                     creator,
                     &parents,
                     &keychains[creator.0],
-                    round
+                    round,
                 ));
             }
         }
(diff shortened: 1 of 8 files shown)


consensus/src/extension/election.rs
```diff
@@ -356,4 +357,31 @@ mod test {
             }
         }
     }
+
+    #[test]
+    #[ignore]
```
timorleph, line 362: Maybe slap a `// TODO(A0-XXXX): this test should pass` above the `#[ignore]`. It's extremely unlikely you'll forget, but just two days ago such a comment reminded me about leftover code, so I'm particularly fond of them.

consensus/src/extension/election.rs
```diff
@@ -356,4 +357,31 @@ mod test {
...
+    fn given_minimal_dag_with_orphaned_node_when_electing_then_orphaned_node_is_not_head() {
```
timorleph, line 363: Btw, doesn't this test pass, even in the current version? I kinda would've suspected it should, although I haven't thought about this too deeply.

consensus/src/extension/extender.rs
```diff
@@ -97,4 +98,29 @@ mod test {
             assert_eq!(batch.len(), n_members.0);
         }
     }
+
+    #[test]
+    #[ignore]
```
timorleph, line 103: Same here.

timorleph (approved):
Strict TDD in ABFT, I've never thought I would see the day... <3 (;_;) <3

consensus/src/extension/election.rs
Marcin-Radecki: It does pass in a situation there are direct parents only in the dag. We'll see what happens after changes to parent voting code.

consensus/src/extension/extender.rs
Marcin-Radecki: Done

mike1729 (approved):

PR #506, opened 2024-12-11
A0-4563: Introduce notion of non-direct parents in Extender

Marcin-Radecki:
This PR extends `UnitWithParents` trait so that it will be possible to distinguish direct and non-direct parents of a unit. Those changes are required for later changes in DAG reconstruction and validation. This change must be a no-op from the algorithm perspective. I am not going to uncomment UT added recently just yet,

### DIFF
--- a/consensus/src/dag/reconstruction/mod.rs
+++ b/consensus/src/dag/reconstruction/mod.rs
@@ -93,7 +93,7 @@ impl<U: Unit> UnitWithParents for ReconstructedUnit<U> {
             .filter_map(|(hash, parent_round)| match self.unit.coord().round() {
                 // round 0 units cannot have non-empty parents
                 0 => None,
-                
+
                 unit_round => {
                     if unit_round - 1 == *parent_round {
                         Some(hash)
--- a/consensus/src/dag/reconstruction/mod.rs
+++ b/consensus/src/dag/reconstruction/mod.rs
@@ -29,7 +29,8 @@ impl<U: Unit> ReconstructedUnit<U> {
                 let unit_round = unit.round();
                 let mut parents_with_rounds = NodeMap::with_size(parents.size());
                 for (parent_index, hash) in parents.into_iter() {
-                    parents_with_rounds.insert(parent_index, (hash, unit_round - 1));
+                    // we cannot have here round 0 units
+                    parents_with_rounds.insert(parent_index, (hash, unit_round.saturating_sub(1)));
                 }
                 Ok(ReconstructedUnit {
                     unit,
@@ -90,7 +91,9 @@ impl<U: Unit> UnitWithParents for ReconstructedUnit<U> {
         self.parents
             .values()
             .filter_map(|(hash, parent_round)| match self.unit.coord().round() {
+                // round 0 units cannot have non-empty parents
                 0 => None,
+                
                 unit_round => {
                     if unit_round - 1 == *parent_round {
                         Some(hash)
@@ -105,7 +108,7 @@ impl<U: Unit> UnitWithParents for ReconstructedUnit<U> {
         self.parents.get(index).map(|(hash, _)| hash)
     }
 
-    fn parents_size(&self) -> NodeCount {
+    fn node_count(&self) -> NodeCount {
         self.parents.size()
     }
 }
--- a/consensus/src/dag/reconstruction/mod.rs
+++ b/consensus/src/dag/reconstruction/mod.rs
@@ -1,5 +1,5 @@
 use std::collections::HashMap;
-
+use aleph_bft_rmc::NodeCount;
 use crate::{
     units::{ControlHash, FullUnit, HashFor, Unit, UnitCoord, UnitWithParents, WrappedUnit},
     Hasher, NodeMap, SessionId,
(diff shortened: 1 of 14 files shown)


consensus/src/dag/reconstruction/mod.rs
```diff
@@ -25,7 +25,17 @@ impl<U: Unit> ReconstructedUnit<U> {
         match unit.control_hash().combined_hash
             == ControlHash::<U::Hasher>::combine_hashes(&parents)
         {
-            true => Ok(ReconstructedUnit { unit, parents }),
+            true => {
+                let unit_round = unit.round();
+                let mut parents_with_rounds = NodeMap::with_size(parents.size());
+                for (parent_index, hash) in parents.into_iter() {
+                    parents_with_rounds.insert(parent_index, (hash, unit_round - 1));
```
Marcin-Radecki, line 32: This can underflow, but in practice `unit_round` is always positive here. Add `assert`?

timorleph: I would go with a `saturating_sub` plus comment that it doesn't matter. Or maybe not even that, since this code is extremely temporary and likely will never even get released, never mind encountering any strange states. :P

consensus/src/dag/reconstruction/mod.rs
```diff
@@ -72,8 +82,31 @@ impl<U: Unit> WrappedUnit<U::Hasher> for ReconstructedUnit<U> {
...
+                    }
+                }
+            })
+    }
+
+    fn parent_for(&self, index: NodeIndex) -> Option<&HashFor<Self>> {
+        self.parents.get(index).map(|(hash, _)| hash)
+    }
+
+    fn parents_size(&self) -> NodeCount {
```
timorleph, line 108: Better call this `node_count` or something, the current name might suggest that it's about the number of parents, which would be confusing.

consensus/src/dag/reconstruction/mod.rs
```diff
@@ -72,8 +82,31 @@ impl<U: Unit> WrappedUnit<U::Hasher> for ReconstructedUnit<U> {
...
-        &self.parents
+    fn parents(&self) -> impl Iterator<Item = &HashFor<U>> {
+        self.parents.values().map(|(hash, _)| hash)
+    }
+
+    fn direct_parents(&self) -> impl Iterator<Item = &HashFor<Self>> {
+        self.parents
+            .values()
+            .filter_map(|(hash, parent_round)| match self.unit.coord().round() {
+                0 => None,
```
timorleph, line 93: Maybe a short comment here for clarity that this never matters, since round 0 units cannot have nonempty parents.

consensus/src/extension/election.rs
```diff
@@ -63,11 +63,12 @@ impl<U: UnitWithParents> CandidateElection<U> {
 
     fn vote_from_parents(
         &mut self,
-        parents: &NodeMap<HashFor<U>>,
+        parents: Vec<HashFor<U>>,
+        parents_size: NodeCount,
```
timorleph, line 67: Perhaps it's clearer now to pass the threshold here, rather than the node count. Or at least rename this to `node_count`, due to the possible confusion I mentioned in another comment.

consensus/src/extension/election.rs
```diff
@@ -102,12 +103,13 @@ impl<U: UnitWithParents> CandidateElection<U> {
             return Ok(());
         }
         let relative_round = voter.round() - self.round;
+        let direct_parents = voter.direct_parents().cloned().collect::<Vec<_>>();
```
timorleph, line 106: Heh, most code we have uses `let direct_parents: Vec<_> = (...).collect()` rather than the turbofish.

timorleph: Wait, is the typing even necessary if these are only used in one call where the type is clear?

consensus/src/testing/dag.rs
```diff
@@ -139,16 +139,16 @@ impl DagFeeder {
...
             .units_map
             .get(&h)
             .expect("we have the unit")
             .parent_hashes()
             .into_iter()
             .collect();
-        assert_eq!(parents.item_count(), expected_hashes.len());
-        for (_, hash) in parents {
+        assert_eq!(parents.len(), expected_hashes.len());
+        for hash in parents {
```
timorleph, line 151: Hm, wouldn't now `collect`ing `parents()` into a `HashSet` actually let you just `assert_eq` on these two hashsets?

timorleph (approved):
Lots of tiny nitpicks, but in general code looks great.

consensus/src/testing/dag.rs
Marcin-Radecki: Yes, changed

timorleph: Uhm, but you can collect them straight into a `HashSet`, no need for the `Vec` first. :P Also preferably by typing the variable, rather than the turbofish.

Marcin-Radecki: Right, done

consensus/src/extension/election.rs
```diff
@@ -102,12 +102,16 @@ impl<U: UnitWithParents> CandidateElection<U> {
             return Ok(());
         }
         let relative_round = voter.round() - self.round;
+        let direct_parents = voter.direct_parents().cloned().collect();
```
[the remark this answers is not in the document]
Marcin-Radecki: Good point, done

PR #514, opened 2025-01-02
A0-4560: Changes in dag reconstruction to include parent round

Marcin-Radecki:
This PR is a follow up after https://github.com/Cardinal-Cryptography/AlephBFT/commit/71f03144faa5856c85f477142c013fe36659b9e3, that introduced notion of parent round in the Extender. 

Here we modify `ControlHash`  accordingly. Previously, it was using `NodeSubset` which was just a `Vec<bool>` so a true/false whether given node index created a parent unit. Since we need now also the parent round, all code that reconstructs unit now have notion of a parent round. 

`NodeSubset` is not removed as it is still used in the codebase.

This change is actually a no-op to the algo - you can see in the PR removed code `round - 1` to just parent round from unit store, however in practice it is still `round - 1` as the changes to Creator are not done yet. This will be done in the next PR.

### DIFF
--- a/consensus/src/dag/reconstruction/mod.rs
+++ b/consensus/src/dag/reconstruction/mod.rs
@@ -96,7 +96,7 @@ impl<U: Unit> UnitWithParents for ReconstructedUnit<U> {
     fn parent_for(&self, index: NodeIndex) -> Option<&HashFor<Self>> {
         self.parents.get(index).map(|(hash, _)| hash)
     }
-    
+
     #[cfg(test)]
     fn parent_round(&self, index: NodeIndex) -> Option<Round> {
         self.parents.get(index).map(|(_, round)| *round)
@@ -243,6 +243,7 @@ impl<U: Unit> Reconstruction<U> {
 
 #[cfg(test)]
 mod test {
+    use crate::units::borked_random_full_parent_units_up_to;
     use crate::{
         dag::reconstruction::{ReconstructedUnit, Reconstruction, ReconstructionResult, Request},
         units::{random_full_parent_units_up_to, Unit, UnitCoord, UnitWithParents},
@@ -306,6 +307,50 @@ mod test {
                     }
                 }
             }
+        }
+    }
+
+    #[test]
+    fn given_dag_units_with_wrong_when_reconstructing_then_unit_is_not_reconstructed() {
+        let mut reconstruction = Reconstruction::new();
+        let dag = borked_random_full_parent_units_up_to(2, NodeCount(4), 43);
+        for initial_unit in dag.get(0).expect("just created") {
+            let ReconstructionResult {
+                mut units,
+                requests,
+            } = reconstruction.add_unit(initial_unit.clone());
+            assert!(requests.is_empty());
+            assert_eq!(units.len(), 1);
+            let reconstructed_unit = units.pop().expect("just checked its there");
+            assert_eq!(
+                reconstructed_unit,
+                ReconstructedUnit::initial(initial_unit.clone())
+            );
+            assert_eq!(reconstructed_unit.parents().count(), 0);
+        }
+
+        for first_round_unit in dag.get(1).expect("just created") {
+            let ReconstructionResult {
+                mut units,
+                requests,
+            } = reconstruction.add_unit(first_round_unit.clone());
+            assert!(requests.is_empty());
+            assert_eq!(units.len(), 1);
+
+            let reconstructed_unit = units.pop().expect("just checked its there");
+            // special case - round-1 unit pretended to be 0th unit, so the reconstruction succeed
+            assert_eq!(reconstructed_unit.parents().count(), 0);
+        }
+
+        for second_round_unit in dag.get(2).expect("just created") {
+            let ReconstructionResult { units, requests } =
+                reconstruction.add_unit(second_round_unit.clone());
+            assert!(units.is_empty());
+
+            assert_eq!(
+                requests.last().expect("just checked"),
+                &Request::ParentsOf(second_round_unit.hash()),
+            );
         }
     }
 
--- a/consensus/src/dag/reconstruction/mod.rs
+++ b/consensus/src/dag/reconstruction/mod.rs
@@ -22,7 +22,7 @@ pub struct ReconstructedUnit<U: Unit> {
 impl<U: Unit> ReconstructedUnit<U> {
     /// Returns a reconstructed unit if the parents agree with the hash, errors out otherwise.
     pub fn with_parents(unit: U, parents: NodeMap<(HashFor<U>, Round)>) -> Result<Self, U> {
-        match unit.control_hash().combined_hash
+        match unit.control_hash().combined_hash()
             == ControlHash::<U::Hasher>::create_control_hash(&parents)
         {
             true => Ok(ReconstructedUnit { unit, parents }),
--- a/consensus/src/dag/reconstruction/mod.rs
+++ b/consensus/src/dag/reconstruction/mod.rs
@@ -93,7 +93,7 @@ impl<U: Unit> UnitWithParents for ReconstructedUnit<U> {
             .filter_map(|(hash, parent_round)| match self.unit.coord().round() {
                 // round 0 units cannot have non-empty parents
                 0 => None,
-                
+
                 unit_round => {
                     if unit_round - 1 == *parent_round {
                         Some(hash)
--- a/consensus/src/dag/reconstruction/mod.rs
+++ b/consensus/src/dag/reconstruction/mod.rs
@@ -25,12 +25,7 @@ impl<U: Unit> ReconstructedUnit<U> {
         match unit.control_hash().combined_hash
             == ControlHash::<U::Hasher>::combine_hashes(&parents)
         {
-            true => {
-                Ok(ReconstructedUnit {
-                    unit,
-                    parents,
-                })
-            }
+            true => Ok(ReconstructedUnit { unit, parents }),
             false => Err(unit),
         }
     }
--- a/consensus/src/dag/reconstruction/mod.rs
+++ b/consensus/src/dag/reconstruction/mod.rs
@@ -29,7 +29,8 @@ impl<U: Unit> ReconstructedUnit<U> {
                 let unit_round = unit.round();
                 let mut parents_with_rounds = NodeMap::with_size(parents.size());
                 for (parent_index, hash) in parents.into_iter() {
-                    parents_with_rounds.insert(parent_index, (hash, unit_round - 1));
+                    // we cannot have here round 0 units
+                    parents_with_rounds.insert(parent_index, (hash, unit_round.saturating_sub(1)));
                 }
                 Ok(ReconstructedUnit {
                     unit,
@@ -90,7 +91,9 @@ impl<U: Unit> UnitWithParents for ReconstructedUnit<U> {
         self.parents
             .values()
             .filter_map(|(hash, parent_round)| match self.unit.coord().round() {
+                // round 0 units cannot have non-empty parents
                 0 => None,
+                
                 unit_round => {
                     if unit_round - 1 == *parent_round {
                         Some(hash)
@@ -105,7 +108,7 @@ impl<U: Unit> UnitWithParents for ReconstructedUnit<U> {
         self.parents.get(index).map(|(hash, _)| hash)
     }
 
-    fn parents_size(&self) -> NodeCount {
+    fn node_count(&self) -> NodeCount {
         self.parents.size()
     }
 }
--- a/consensus/src/dag/reconstruction/mod.rs
+++ b/consensus/src/dag/reconstruction/mod.rs
@@ -26,15 +26,10 @@ impl<U: Unit> ReconstructedUnit<U> {
             == ControlHash::<U::Hasher>::combine_hashes(&parents)
         {
             true => {
-                let unit_round = unit.round();
-                let mut parents_with_rounds = NodeMap::with_size(parents.size());
-                for (parent_index, hash) in parents.into_iter() {
-                    // we cannot have here round 0 units
-                    parents_with_rounds.insert(parent_index, (hash, unit_round.saturating_sub(1)));
-                }
+                let parents = unit.control_hash().parents_round_lookup.clone();
                 Ok(ReconstructedUnit {
                     unit,
-                    parents: parents_with_rounds,
+                    parents,
                 })
             }
             false => Err(unit),
--- a/consensus/src/dag/reconstruction/mod.rs
+++ b/consensus/src/dag/reconstruction/mod.rs
@@ -23,7 +23,7 @@ impl<U: Unit> ReconstructedUnit<U> {
     /// Returns a reconstructed unit if the parents agree with the hash, errors out otherwise.
     pub fn with_parents(unit: U, parents: NodeMap<(HashFor<U>, Round)>) -> Result<Self, U> {
         match unit.control_hash().combined_hash
-            == ControlHash::<U::Hasher>::combine_hashes(&parents)
+            == ControlHash::<U::Hasher>::create_control_hash(&parents)
         {
             true => Ok(ReconstructedUnit { unit, parents }),
             false => Err(unit),
--- a/consensus/src/dag/reconstruction/mod.rs
+++ b/consensus/src/dag/reconstruction/mod.rs
@@ -97,11 +97,6 @@ impl<U: Unit> UnitWithParents for ReconstructedUnit<U> {
         self.parents.get(index).map(|(hash, _)| hash)
     }
 
-    #[cfg(test)]
-    fn parent_round(&self, index: NodeIndex) -> Option<Round> {
-        self.parents.get(index).map(|(_, round)| *round)
-    }
-
     fn node_count(&self) -> NodeCount {
         self.parents.size()
     }
@@ -297,12 +292,6 @@ mod test {
                             parents.iter().zip(reconstructed_unit.parents())
                         {
                             assert_eq!(&parent.hash(), reconstructed_parent);
-                            assert_eq!(
-                                reconstructed_unit
-                                    .parent_round(parent.creator())
-                                    .expect("we should have round"),
-                                round - 1
-                            );
                         }
                     }
                 }
--- a/consensus/src/dag/reconstruction/mod.rs
+++ b/consensus/src/dag/reconstruction/mod.rs
@@ -97,6 +97,10 @@ impl<U: Unit> UnitWithParents for ReconstructedUnit<U> {
         self.parents.get(index).map(|(hash, _)| hash)
     }
 
+    fn parent_round(&self, index: NodeIndex) -> Option<Round> {
+        self.parents.get(index).map(|(_, round)| *round)
+    }
+
     fn node_count(&self) -> NodeCount {
         self.parents.size()
     }
@@ -238,13 +242,14 @@ impl<U: Unit> Reconstruction<U> {
 
 #[cfg(test)]
 mod test {
-    use std::collections::HashMap;
-
     use crate::{
         dag::reconstruction::{ReconstructedUnit, Reconstruction, ReconstructionResult, Request},
         units::{random_full_parent_units_up_to, Unit, UnitCoord, UnitWithParents},
         NodeCount, NodeIndex,
     };
+    use aleph_bft_types::{NodeMap, Round};
+    use rand::Rng;
+    use std::collections::HashMap;
 
     #[test]
     fn reconstructs_initial_units() {
@@ -290,6 +295,12 @@ mod test {
                             parents.iter().zip(reconstructed_unit.parents())
                         {
                             assert_eq!(&parent.hash(), reconstructed_parent);
+                            assert_eq!(
+                                reconstructed_unit
+                                    .parent_round(parent.creator())
+                                    .expect("we should have round"),
+                                round - 1
+                            );
                         }
                     }
                 }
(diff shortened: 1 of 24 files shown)


consensus/src/units/mod.rs
```diff
@@ -56,32 +56,35 @@ impl Display for UnitCoord {
...
+    /// Calculate parent control hash, which includes all parent hashes into account.
+    pub(crate) fn combine_hashes(parent_map: &NodeMap<(H::Hash, Round)>) -> H::Hash {
+        // we include parent rounds with calculating hash but this is okay - we cannot
+        // have two units with the same hash but different rounds
         parent_map.using_encoded(H::hash)
     }
 
-    pub(crate) fn parents(&self) -> impl Iterator<Item = NodeIndex> + '_ {
-        self.parents_mask.elements()
+    pub(crate) fn parents(&self) -> impl Iterator<Item = (NodeIndex, &(H::Hash, Round))> + '_ {
```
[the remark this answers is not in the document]
timorleph: AAaa,a wait, this is much more wrong, why is there `H::Hash` at all? It makes everything much _much_ bigger and makes the control hash pointless – the entire idea with it was to _not_ send hashes for every unit, just a single combined one. And I _almost_ missed this, I was just about to submit a review with just the other comments. :woozy_face: 

Anyway, remove the hashes from the control hash.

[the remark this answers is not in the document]
timorleph: This actually makes the first comment in this thread a much more reasonable suggestion – cloning hashes is annoying, but `UnitCoord` is tiny (and possibly even `Copy`?).

timorleph (changes requested):
Oof, see my last comments for what's really wrong with this PR. Except for that one major thing, the rest is quite reasonable, even if in places I'm a bit sad about the juggling between `NodeMap`s and `UnitCoords` (I don't think this can be made better, so no inline comments about it, unless I forgot to delete one :P).

DamianStraszak (changes requested):
Initial pass, I will take another look after the hashes are removed from control hash.

consensus/src/units/mod.rs
```diff
@@ -56,32 +56,35 @@ impl Display for UnitCoord {
 /// parents
 #[derive(Clone, Eq, PartialEq, Hash, Debug, Decode, Encode)]
 pub struct ControlHash<H: Hasher> {
-    pub(crate) parents_mask: NodeSubset,
+    pub(crate) parents_round_lookup: NodeMap<(H::Hash, Round)>,
```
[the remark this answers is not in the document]
timorleph: > It should be something like `Vec<Option<NodeIndex>>`

Good news, `NodeMap<Round>` is exactly that! (Or rather with `Round` in place of `NodeIndex`, but I'm pretty sure that's what you meant). I'm not sure we can make this significantly smaller with a special `Codec` implementation though, since now we actually need to send the rounds rather than just bits. :/

consensus/src/units/mod.rs
```diff
@@ -56,32 +56,35 @@ impl Display for UnitCoord {
...
+            parents_round_lookup: parents_round_lookup.clone(),
+            combined_hash: Self::combine_hashes(parents_round_lookup),
         }
     }
 
-    pub(crate) fn combine_hashes(parent_map: &NodeMap<H::Hash>) -> H::Hash {
+    /// Calculate parent control hash, which includes all parent hashes into account.
+    pub(crate) fn combine_hashes(parent_map: &NodeMap<(H::Hash, Round)>) -> H::Hash {
+        // we include parent rounds with calculating hash but this is okay - we cannot
+        // have two units with the same hash but different rounds
```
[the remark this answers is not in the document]
timorleph: Doesn't seem like you did, did you miss a commit?

timorleph (approved):
Looks good imo, caught you on one lie. ;p

consensus/src/dag/reconstruction/mod.rs
```diff
@@ -21,22 +21,11 @@ pub struct ReconstructedUnit<U: Unit> {
 
 impl<U: Unit> ReconstructedUnit<U> {
     /// Returns a reconstructed unit if the parents agree with the hash, errors out otherwise.
-    pub fn with_parents(unit: U, parents: NodeMap<HashFor<U>>) -> Result<Self, U> {
+    pub fn with_parents(unit: U, parents: NodeMap<(HashFor<U>, Round)>) -> Result<Self, U> {
```
timorleph, line 24: Maybe add some tests for this function with actual nontrivial `Round` values, e.g. wrong rounds with matching hashes, everything correct, but there are some ancient units etc.

consensus/src/dag/reconstruction/parents.rs
```diff
@@ -3,13 +3,14 @@ use crate::{
...
 };
+use aleph_bft_types::Round;
 use std::collections::{hash_map::Entry, HashMap};
 
 /// A unit in the process of reconstructing its parents.
 #[derive(Debug, PartialEq, Eq, Clone)]
 enum ReconstructingUnit<U: Unit> {
     /// We are trying to optimistically reconstruct the unit from potential parents we get.
-    Reconstructing(U, NodeMap<HashFor<U>>),
+    Reconstructing(U, NodeMap<(HashFor<U>, Round)>),
```
timorleph, line 13: Some tests that check that this addition works would be nice as well.

timorleph (approved):
I hate that GitHub confuses reviews with comments. ;_;

consensus/src/units/testing.rs
```diff
@@ -135,15 +135,18 @@ fn random_initial_reconstructed_units(
         .collect()
 }
 
-fn parent_map<U: Unit<Hasher = Hasher64>>(parents: &Vec<U>) -> NodeMap<Hash64> {
+fn parent_map<U: Unit<Hasher = Hasher64>>(
+    parents: &Vec<U>,
+    parent_round: Round,
```
timorleph, line 140: Wait, this makes very little sense, you can just use `parent.round()` and make this function much more generic, no?

consensus/src/testing/crash_recovery.rs
```diff
@@ -132,10 +132,10 @@ fn verify_backup(buf: &mut &[u8]) -> HashSet<UnitCoord> {
         let unit = <UncheckedSignedUnit<Hasher64, Data, Signature>>::decode(buf).unwrap();
         let full_unit = unit.as_signable();
         let coord = full_unit.coord();
-        let parent_ids = &full_unit.as_pre_unit().control_hash().parents_mask;
+        let parent_ids = &full_unit.as_pre_unit().control_hash().parents;
 
-        for parent_id in parent_ids.elements() {
-            let parent = UnitCoord::new(coord.round() - 1, parent_id);
+        for (parent_id, &round) in parent_ids.iter() {
+            let parent = UnitCoord::new(round, parent_id);
```
timorleph, line 138: You don't need this line if you iterate over `control_hash().parents()` rather than touching internals (probably a leftover from other changes after previous reviews).

consensus/src/units/validator.rs
```diff
@@ -138,7 +138,7 @@ impl<K: Keychain> Validator<K> {
                     return Err(ValidationError::NotEnoughParents(pre_unit.clone()));
                 }
                 let control_hash = &pre_unit.control_hash();
-                if !control_hash.parents_mask[pre_unit.creator()] {
+                if control_hash.parents.get(pre_unit.creator()).is_none() {
```
timorleph, line 141: Oof, as @DamianStraszak pointed out through other channels this is wrong, together with a significant part of this method. So, the problems:
1. The parent of this creator should not only exist, but also be of round exactly `n-1`.
2. We should check that we have at least `threshold` parents of round exactly `n-1`.
3. We should check that there are no parents of rounds _higher_ than `n-1`.

All this due to `ControlHash` suddenly becoming much less trivial. In fact, I would recommend moving this `ControlHash` validation logic into a method of `ControlHash`, like `pub fn validate(&self, unit_round: Round) -> Result<(), Error>` or something. Perhaps even create a submodule for `ControlHash`, since it became so nontrivial, and then add some tests there.

timorleph (changes requested):
Damian found some serious problems, but these seem to actually be the last one this time. <_<"

consensus/src/testing/crash_recovery.rs
Marcin-Radecki: Correct, changed

consensus/src/units/testing.rs
Marcin-Radecki: You're right, I don't know why I did this in the first place

Marcin-Radecki (commented):
I pushed changes for all remarks except control hash validation

consensus/src/units/mod.rs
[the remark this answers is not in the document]
Marcin-Radecki: This should be now cleared, as this comment pointed to different commit in the PR, but for some reasons GH still shows it there

consensus/src/dag/reconstruction/parents.rs
timorleph: Oh, no, that's not what I meant, sorry for not being clear, adding methods just to test them doesn't seem great. <_<"

I wanted there to be a test in a similar vein to the one in `dag::reconstruction`, i.e. in this case passing a unit with wrong round in `add_unit` I guess?

Marcin-Radecki: Okay, let me think about it, as when the unit is created I cannot easily change its round, can I. But you're right I should revert this `parent_round` method.

consensus/src/dag/reconstruction/parents.rs
```diff
@@ -178,6 +180,7 @@ impl<U: Unit> Reconstruction<U> {
                             unit_hash,
                             parent_coord.creator(),
                             *parent_hash,
+                            parent_coord.round(),
```
Marcin-Radecki, line 183: While writing some unit test to cover scenario in which a unit pretend to have different coord - there are two distinc sources of information passed now down to `with_parents:
1. `parent_hash` from the line above comes from `units_by_coord` map, which is filled what a unit claims to have its coord
2. `parent_coord.round` is really taken from its control hash
and since we don't have any validation between unit coord and its control hash, at least in the tests I can pass here mismatched data, ie see commit https://github.com/Cardinal-Cryptography/AlephBFT/pull/514/commits/aef31c8892f6aa4345c75729986ca367ce91afc6 . Now my question is, is this line okay, or should I change `units_by_coord` to keep both hash and round, so sources of both value is the same?

consensus/src/dag/reconstruction/parents.rs
Marcin-Radecki: This should be now addressed.

consensus/src/units/control_hash.rs
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    /// Returns number of all members in abft consensus
+    pub(crate) fn n_members(&self) -> NodeCount {
+        self.parents.size()
+    }
+
+    /// Validate
+    pub fn validate(&self, unit_coord: UnitCoord) -> Result<(), Error> {
+        assert!(unit_coord.round > 0, "Round must be greater than 0");
```
Marcin-Radecki, line 57: We should validate also initial units here

consensus/src/units/control_hash.rs
```diff
@@ -0,0 +1,257 @@
...
+        assert_eq!(
+            ch.validate(UnitCoord::new(3, NodeIndex(1)))
+                .expect_err("validate() should return error, returned Ok(()) instead"),
+            Error::NotEnoughParentsForRound(2)
+        );
+    }
+
+    #[test]
+    fn given_control_hash_when_there_are_parents_from_greater_rounds_then_err_is_returned_from_validate(
+    ) {
```
Marcin-Radecki, line 242: Add a test discussed offline

consensus/src/dag/reconstruction/mod.rs
```diff
@@ -308,6 +299,50 @@ mod test {
         }
     }
 
+    #[test]
+    fn given_dag_units_with_wrong_when_reconstructing_then_unit_is_not_reconstructed() {
```
Marcin-Radecki, line 303: This test does not make sense since it's done after validation

consensus/src/units/control_hash.rs
Marcin-Radecki: Done

consensus/src/units/control_hash.rs
Marcin-Radecki: Done, but please bear with me I added them in ControlHash module only.

consensus/src/dag/reconstruction/mod.rs
Marcin-Radecki: Removed

consensus/src/dag/reconstruction/parents.rs
Marcin-Radecki: This comment is not valid anymore, since I did not take into account validation, which filters this scenario

consensus/src/units/validator.rs
Marcin-Radecki: This should be now addressed.

consensus/src/units/control_hash.rs
```diff
@@ -0,0 +1,412 @@
...
+            Error::NotDescendantOfPreviousUnit(node_index) => {
+                write!(
+                    f,
+                    "non-zero round unit is not descendant of its creator's previous unit, creator index: {:?}",
+                    node_index)
+            }
+            Error::DescendantOfPreviousUnitButWrongRound(round) => {
+                write!(
+                f,
+                "non-zero round unit is descendant of its creator's previous unit, but has wrong round {:?}",
```
timorleph, line 42: Nit: technically "previous" means "exactly preceeding", so wrong round in this case means it's _not_ a descendant of the previous unit.

consensus/src/units/control_hash.rs
```diff
@@ -0,0 +1,412 @@
...
+        }
+    }
+}
+
+/// Combined hashes of the parents of a unit together with the set of indices of creators of the
+/// parents. By parent here we mean a parent hash and its round.
+#[derive(Clone, Eq, PartialEq, Hash, Debug, Decode, Encode)]
+pub struct ControlHash<H: Hasher> {
+    pub(crate) parents: NodeMap<Round>,
+    pub(crate) combined_hash: H::Hash,
```
timorleph, line 69: Do these really have to be `pub`? You seem to have proper access and creation methods for everything, no?

consensus/src/units/control_hash.rs
```diff
@@ -0,0 +1,412 @@
...
+}
+
+impl<H: Hasher> ControlHash<H> {
+    /// Creates new control hash from parents hashes and rounds
+    pub(crate) fn new(parents_with_rounds_and_hashes: &NodeMap<(H::Hash, Round)>) -> Self {
```
timorleph, line 74: Speaking of `pub` – I'm pretty sure you can just remove the `(crate)` everywhere here, it's ugly and serves no purpose.

consensus/src/units/control_hash.rs
```diff
@@ -0,0 +1,412 @@
...
+        }
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+pub mod tests {
+    use crate::units::control_hash::Error;
+    use crate::units::UnitCoord;
+    use crate::{units::ControlHash, NodeCount, NodeIndex};
```
timorleph, line 178: Combine these (I thought clippy or fmt complained about this, but apparently not?).

consensus/src/units/control_hash.rs
```diff
@@ -0,0 +1,412 @@
...
+    #[test]
+    fn given_control_hash_then_basic_properties_are_correct() {
+        let parent_map = vec![
+            Some(([0; 8], 2)),
+            None,
+            Some(([2; 8], 2)),
+            Some(([3; 8], 2)),
+            Some(([4; 8], 2)),
+            Some(([5; 8], 2)),
+            Some(([5; 8], 1)),
```
timorleph, line 202: Uhm, please don't pass multiple different units with the same hash – this is not a situation we want to test, at least not if we are not testing some weird corner case on purpose. Some tests below also have this problem.

(It currently works, but we might at some point assume that this is not the case, which could break these tests for no good reason.)

consensus/src/units/control_hash.rs
```diff
@@ -0,0 +1,412 @@
...
+            ch.validate(UnitCoord::new(0, NodeIndex(1)))
+                .expect_err("validate() should return error, returned Ok(()) instead"),
+            Error::RoundZeroWithSomeParents(NodeCount(parent_map.item_count()))
+        );
+    }
+
+    #[test]
+    fn given_initial_round_when_validate_with_different_node_count_then_validate_returns_err() {
+        // Correct control hash for initial unit with count = 4 is
+        // [16, 0, 0, 0, 0, 129, 99, 217, 65, 183, 158, 24, 201];
```
timorleph, line 249: Hmm, perhaps it would be better to construct the control hash, encode it, and then flip a bit in the latter part of it, rather than manually putting in the encoding? It will still require a comment, to make clear that you are flipping a bit in the control hash, but it will at least be less fragile to potential future changes to `ControlHash`.

consensus/src/units/control_hash.rs
```diff
@@ -0,0 +1,412 @@
...
+    fn given_initial_round_when_validate_called_for_initial_unit_then_function_panics() {
+        let parent_map = vec![
+            Some(([0; 8], 2)),
+            None,
+            Some(([2; 8], 2)),
+            Some(([3; 8], 2)),
+        ]
+        .into();
+        let ch = ControlHash::<Hasher64>::new(&parent_map);
+        let _ = ch.validate_non_initial_round(UnitCoord::new(0, NodeIndex(1)));
```
timorleph, line 298: No need to test an internal function like that imo.

consensus/src/units/control_hash.rs
timorleph: Oh, not for `combined_hash`, but it would probably be nice if there was one.

consensus/src/units/control_hash.rs
```diff
@@ -0,0 +1,412 @@
...
+        let borked_hash_of_fourth_round_unit =
+            ControlHash::<Hasher64>::new(&parents_from_round_three_but_one_hash_replaced);
+        assert_ne!(
+            borked_hash_of_fourth_round_unit,
+            control_hash_of_fourth_round_unit
+        );
+
+        let mut parents_from_round_three_but_one_unit_replaced = parents_from_round_three.clone();
+        parents_from_round_three_but_one_unit_replaced
+            .insert(NodeIndex(2), all_parents_from_round_three[3].unwrap());
```
timorleph, line 400: I think this was supposed to change the `round`, not replace the whole thing.

consensus/src/units/validator.rs
```diff
@@ -34,20 +31,10 @@ impl<H: Hasher, D: Data, S: Signature> Display for ValidationError<H, D, S> {
...
-                "nonzero round unit with only {:?} parents: {:?}",
-                pu.n_parents(),
-                pu
-            ),
-            NotDescendantOfPreviousUnit(pu) => write!(
+            ParentValidationFailed(pu, control_hash_error) => write!(
                 f,
-                "nonzero round unit is not descendant of its creator's previous unit: {:?}",
-                pu
+                "parent validation failed for nit: {:?}. Internal error: {:?}",
```
timorleph, line 36: Nit: s/nit/unit :>

Also just "...for unit {:?}: {}" will probably look nicer. (Note the `{}` for the error.)

consensus/src/units/validator.rs
```diff
@@ -118,40 +101,18 @@ impl<K: Keychain> Validator<K> {
...
-                if !control_hash.parents_mask[pre_unit.creator()] {
-                    return Err(ValidationError::NotDescendantOfPreviousUnit(
-                        pre_unit.clone(),
-                    ));
-                }
-            }
-        }
+        let unit_coord = UnitCoord::new(pre_unit.round(), pre_unit.creator());
+        pre_unit.control_hash.validate(unit_coord).map_err(|e| {
+            ValidationError::<H, D, K::Signature>::ParentValidationFailed(pre_unit.clone(), e)
```
timorleph, line 106: I don't think you need the turbofish here.

consensus/src/units/validator.rs
```diff
@@ -118,40 +101,18 @@ impl<K: Keychain> Validator<K> {
...
+        })?;
         Ok(su)
     }
 }
 
 #[cfg(test)]
 mod tests {
     use super::{ValidationError::*, Validator as GenericValidator};
+    use crate::units::ControlHashError;
```
timorleph, line 115: Join the import with the one below.

consensus/src/units/control_hash.rs
```diff
@@ -0,0 +1,412 @@
...
+        self.previous_round_have_enough_parents(unit_coord.round)?;
+        self.check_if_parents_greater_than_previous_round(unit_coord.round)?;
+
+        Ok(())
+    }
+
+    fn check_if_parents_greater_than_previous_round(&self, round: Round) -> Result<(), Error<H>> {
+        let parents_greater_than_previous_round = self
+            .parents()
+            .filter(|&parent| parent.round > round - 1)
```
timorleph, line 139: You can just use `any` here.

consensus/src/units/control_hash.rs
```diff
@@ -0,0 +1,412 @@
+use crate::{units::UnitCoord, Hasher, NodeCount, NodeIndex, NodeMap, Round};
+use codec::{Decode, Encode};
+use std::fmt::{Display, Formatter, Result as FmtResult};
+use std::hash::Hash;
```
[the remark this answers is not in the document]
Marcin-Radecki: Done

consensus/src/units/control_hash.rs
Marcin-Radecki: Changed

consensus/src/units/control_hash.rs
```diff
@@ -0,0 +1,412 @@
...
+                write!(
+                    f,
+                    "non-zero round unit has not enough parents from the round {:?}",
+                    round
+                )
+            }
+            Error::ParentsHigherThanRound(round) => {
+                write!(
+                    f,
+                    "non-zero round unit has parents higher than round {:?}",
```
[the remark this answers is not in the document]
Marcin-Radecki: Yes, removed those prefixes

consensus/src/units/control_hash.rs
Marcin-Radecki: Removed `pub(create)`, added getter `control_hash()` and fixed some UT which directly used `control_hash` field

consensus/src/units/control_hash.rs
Marcin-Radecki: Good point, removed

consensus/src/units/control_hash.rs
Marcin-Radecki: Good point, done

consensus/src/units/control_hash.rs
Marcin-Radecki: Done

consensus/src/units/control_hash.rs
Marcin-Radecki: This was a mistake propagated by a copy-paste - now fixed.

consensus/src/units/control_hash.rs
Marcin-Radecki: `I'm now starting from encoded control hash, though at some point I need to assumed some internals of encoding, hopefully though this version is better.

consensus/src/units/control_hash.rs
Marcin-Radecki: Good point, test removed

consensus/src/units/control_hash.rs
Marcin-Radecki: Right, changed

consensus/src/units/validator.rs
Marcin-Radecki: Done

consensus/src/units/validator.rs
Marcin-Radecki: Right, removed

consensus/src/units/validator.rs
Marcin-Radecki: Done

DamianStraszak (approved):

PR #548, opened 2025-03-21
A0-4382: Better types and minor reshuffles

timorleph:
This does almost nothing by itself, but it should make the following changes (separating out initial unit collection, and consolidating request logic) much simpler.


consensus/src/dissemination/mod.rs
```diff
@@ -1,25 +1,156 @@
...
+pub enum DisseminationResponse<H: Hasher, D: Data, S: Signature> {
+    /// Response to a coord request, just a single unit.
     Coord(UncheckedSignedUnit<H, D, S>),
+    /// All the parents of the specified unit.
     Parents(H::Hash, Vec<UncheckedSignedUnit<H, D, S>>),
+    /// The newest unit response for initial unit collection.
     NewestUnit(UncheckedSigned<NewestUnitResponse<H, D, S>, S>),
 }
+
+/// A message that has to be passed between committe members for consensus to work.
```
Marcin-Radecki, line 86: Nit: typo `committe` -> `committee`

consensus/src/dissemination/mod.rs
```diff
@@ -1,25 +1,156 @@
...
+            Unit(u) => UnitMessage::Unit(u),
+            Request(node_id, DisseminationRequest::Unit(ReconstructionRequest::Coord(coord))) => {
+                UnitMessage::CoordRequest(node_id, coord)
+            }
+            Request(
+                node_id,
+                DisseminationRequest::Unit(ReconstructionRequest::ParentsOf(hash)),
+            ) => UnitMessage::ParentsRequest(node_id, hash),
+            ParentsResponse(h, units) => UnitMessage::ParentsResponse(h, units),
+            Request(node_id, DisseminationRequest::NewestUnit(_, salt)) => {
```
Marcin-Radecki, line 136: This is am only  one-off case here that we pass `_`, so maybe some comment if it stays permanent or will be refactored in the next PRs?

consensus/src/dissemination/task.rs
```diff
@@ -0,0 +1,13 @@
+use crate::{dag::Request, runway::Salt, units::UncheckedSignedUnit, Data, Hasher, Signature};
+
+/// Repeatable task, either requesting or broadcasting a unit.
```
Marcin-Radecki, line 3: Nit: technically this is not yet `RepeatableTask`, but "just" some task

consensus/src/network/unit.rs
```diff
@@ -0,0 +1,40 @@
...
+    /// Response to RequestNewest: (our index, maybe unit, salt) signed by us
+    NewestResponse(UncheckedSigned<NewestUnitResponse<H, D, S>, S>),
+}
+
+impl<H: Hasher, D: Data, S: Signature> UnitMessage<H, D, S> {
+    pub fn included_data(&self) -> Vec<D> {
+        use UnitMessage::*;
+        match self {
+            Unit(uu) => uu.as_signable().included_data(),
+            CoordRequest(_, _) => Vec::new(),
```
Marcin-Radecki, line 30: Nit: we can use `|` operator to group all match arms which returns `Vec::new()`

Marcin-Radecki (approved):

consensus/src/network/mod.rs
```diff
@@ -74,101 +76,75 @@ mod tests {
...
-        if let Units(RequestCoord(dni, duc)) = decoded.0 {
+        if let Units(CoordRequest(dni, duc)) = decoded.0 {
             assert!(ni == dni && uc == duc, "decoded should equal encoded");
         } else {
-            panic!("Decoded RequestCoord as something else");
-        }
-    }
-
-    #[test]
-    fn decoding_network_data_units_response_coord() {
```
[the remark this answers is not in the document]
timorleph: The related variant was redundant, and now only a raw unit response exists, so there is nothing here to test.

consensus/src/dissemination/mod.rs
timorleph: This will be gone in the next PR, and perhaps even could've been gone in this one, but I don't want to backport the changes I already have prepared here. ^^"

consensus/src/network/unit.rs
timorleph: Hah, good point, improved.

consensus/src/dissemination/task.rs
timorleph: Yup, somewhat improved the comment in general.

consensus/src/dissemination/mod.rs
timorleph: Darn, and I was so proud of always spelling this correctly. :skull:

PR #551, opened 2025-03-31
A0-4567: Move request logic into its own module

timorleph:
This leaves member almost empty, ripe for removal. The last part of the refactor should come shortly.


consensus/src/dissemination/task.rs
```diff
@@ -1,11 +1,684 @@
...
+pub enum DisseminationTask<H: Hasher> {
     /// Perform a request.
     Request(Request<H>),
     /// Broadcast a unit.
-    /// TODO(A0-4567): This should soon only contain the hash.
-    Broadcast(UncheckedSignedUnit<H, D, S>),
+    Broadcast(H::Hash),
+}
+
+enum TaskDetails<H: Hasher, D: Data, S: Signature> {
```
Marcin-Radecki, line 27: There are two delay concepts in this enum, and it's not clear to reader for the first look what is the distinction between them. The former is the time a task will wait in the queue before execution, ie it is a delay time common for all tasks. The latter delay is...wait, are they meaning the same things? After all, both are enqueued using the same method `trigger_tasks` for `task_queue`. Maybe we can have one variant?

Marcin-Radecki (approved):

consensus/src/member.rs
```diff
@@ -243,111 +111,24 @@ where
...
-        ));
-    }
-
-    fn on_unit_discovered(&mut self, new_unit: UncheckedSignedUnit<H, D, S>) {
-        let unit_creator = new_unit.as_signable().creator();
-        let unit_round = new_unit.as_signable().round();
-        if self
-            .top_units
-            .get(unit_creator)
-            .map(|round| round < &unit_round)
```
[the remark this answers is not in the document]
timorleph: Yup, it's the other way around in all ways, the comparison is in the opposite direction, the `unwrap_or` defaults to `false` and the result is that `None` gets returned, so it's all consistent. :>

consensus/src/dissemination/task.rs
timorleph: They are the same thing, the only difference between the variants is whether a message should be sent, but this requires the additional info in the other variant. Oh, and in `Delay` the task execution counter does not get increased.

PR #557, opened 2025-04-09
A0-4383: Consolidate into a single service and update docs

timorleph:
Consolidates `member` and `runway` into a single component called simply `consensus`. Does not change the API, this is purely a refactor.

Also updated the docs. It was supposed to only remove references to `runway` and `member`, but I found a couple other useful updates on the way.


consensus/src/consensus/mod.rs
```diff
@@ -0,0 +1,259 @@
...
+        responses_from_service,
+        config.delay_config().newest_request_delay.clone(),
+    ) {
+        Ok(handle) => handle.fuse(),
+        Err(_) => return,
+    };
+    pin_mut!(starting_round_handle);
+    debug!(target: LOG_TARGET, "Initial unit collection spawned.");
+
+    //TODO: just put the Service here
```
Marcin-Radecki, line 188: Is this a remark for a future PR, or not removed comment?

Marcin-Radecki (approved):
Great work!

PR #559, opened 2025-04-21
fix dependency issues and rustc warnings

reject-i:
###Description
- Remove unused dependencies:
  - aleph-bft-crypto: async-trait, tokio
  - aleph-bft-examples-blockchain: sha3, unsigned-varint
- Remove ineffective #[must_use] attributes from trait methods
- Set workspace resolver to "2" to match edition 2021

All tests continue to pass. Verified with cargo +nightly udeps.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -53,12 +53,10 @@ dependencies = [
 name = "aleph-bft-crypto"
 version = "0.11.0"
 dependencies = [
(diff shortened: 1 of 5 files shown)


reject-i:
[Marcin-Radecki](https://github.com/Marcin-Radecki) Hi ! check pls , thanks

